PR #4, opened 2023-06-19
make schema and query planning more general

jpdus:
Don't really know whether it makes sense to merge because it alters the example quite a lot (and still doesn't work really well with 3.5), feel free to edit or just ignore/close.

However I found this generalizes on you example well and works for queries beyond simple questions and is more reliable and uses less tokens than your example.

Imho it makes more sense to let the model only figure out direct dependencies and then parse the dependency graph programatically?

### DIFF
--- a/question_answer_subquery.py
+++ b/question_answer_subquery.py
@@ -8,45 +8,29 @@ from openai_function_call import OpenAISchema
 from tenacity import retry, stop_after_attempt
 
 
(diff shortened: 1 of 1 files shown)


jxnl:
if you rewrite this one as "task_planning.py" and as a seperate script and it'd make sense to merge and consider building this out to do something like what autogpt is.

PR #7, opened 2023-06-20
Task planning with central order

jpdus:
(apparently I am too dumb for Git, new try... ;) )

Hey Jason I see you were busy as well. I tried a slightly different approach:

- I think for practical/maintenance reasons it might be better to have a central execution engine and figuring out execution order centrally using topological sort.
- I also implemented a PoC for an async version
- I am still not sure whether the simple structure is better but probably it's easier to avoid cluttering the function signatures this way?
- Anyway as expierence is still limited it might not hurt to try different approaches and see what scales better.

(but please don't feel obligued to merge my experiments in your project, you just encouraged me in https://github.com/jxnl/openai_function_call/pull/4#issuecomment-1598282134 and I really like your code and PoCs...; thisis just an early experiment feel free to edit or close, would be interested in you opinion)

I think it´s easily possible to build something like AutoGPT (but more reliable and steerable) out of it....


jxnl:
oh i like the toposort! much better than my method,

might be best to just keep the aexecute (more practical and less cluttered example)
I'd also love to see a different / more creative example to differentiate it

and you could rename to "task_planner_topological_sort.py"
Maybe instead of thinking about querys think about tasks to get closer to autogpt.

task_planning.py
```diff
@@ -0,0 +1,180 @@
...
+    query: str = Field(
+        ...,
+        description="Contains the query in text form. If there are multiple queries, this query can only be answered when all dependant subqueries have been answered.",
+    )
+    subqueries: List[int] = Field(
+        default_factory=list,
+        description="List of the IDs of subqueries that need to be answered before we can answer the main question. Use a subquery when anything may be unknown and we need to ask multiple questions to get the answer. Dependencies must only be other queries.",
+    )
+
+    def execute(self, subquery_answers: List[QueryAnswer] | None) -> QueryAnswer:
```
jxnl, line 34: done need. keep it async

task_planning.py
```diff
@@ -0,0 +1,180 @@
+# Proof of Concept for a task planning and execution system using
+# OpenAIs Functions and topological sort, based on the idea in 
+# question_answer_subquery.py. 
+# This version is simplified and centralizes the execution logic
+# in contrast tu the recursive approach in question_answer_subquery.py
```
jxnl, line 5: feel free to add "added by" line :)

task_planning.py
```diff
@@ -0,0 +1,180 @@
...
+from typing import List, Generator
+from openai_function_call import OpenAISchema
+
+
+class QueryAnswer(OpenAISchema):
+    query_id: int
+    answer: str
+
+
+class Query(OpenAISchema):
```
jxnl, line 19: would be cool to rewrite this as a Task and SubTask planner

task_planning.py
```diff
@@ -0,0 +1,180 @@
...
+    for query_id in execution_order:
+        subquery_answers[query_id] = query_dict[query_id].execute(
+            subquery_answers=[
+                subquery_answers[i] for i in query_dict[query_id].subqueries
+            ]
+        )
+    return subquery_answers
+
+
+async def aexecute_queries_in_order(query_plan: QueryPlan) -> dict[int, QueryAnswer]:
```
jxnl, line 132: what are your thoughts on implementing this as a method in queryplan?

jxnl (commented):
actually i take it back would be cool to have this update the main example i have. 

the use of the toposort and answers dict is much better than my implementation

jpdus:
> oh i like the toposort! much better than my method,
> 
> might be best to just keep the aexecute (more practical and less cluttered example) I'd also love to see a different / more creative example to differentiate it
> 
> and you could rename to "task_planner_topological_sort.py" Maybe instead of thinking about querys think about tasks to get closer to autogpt.

Yep, that was my next idea, appreciate the input. 
I like how Jerry implemented this in [Llama Index](https://github.com/jerryjliu/llama_index/blob/main/llama_index/tools/query_plan.py) following your example but while being more elegant, this might still not be very flexible; e.g. we still need to experiment whether it makes sense to include the tool/function call (vs e.g. just a tool ID/suggestion?) at the planning step or in a separate step afterwards?
Depending on the tool, the requirements regarding necessary context and complex function call signatures can change quite a lot and for token efficiency it might be better to just use a very slim setup in the models and planning step and do some central computation/augmentation afterwards without LLM calls?

Reg. example: I thought about some coding related example (e.g. build some python program that requires documentation/code lookup, shared context); what are your thoughts about external  libraries or example data (can be text files) for the repo - I guess this should remain a very barebone collection of examples? Or can include some example file with (optional) deps just for this file?

task_planning.py
jpdus: I changed the naming throughout to Tasks instead of queries

task_planning.py
jpdus: sure, makes way more sense, was just to tired to see that...

jpdus:
Many thanks for the suggestions, tried to adress your points.
If you really want to add this to your main example it's probably better if you do it b/c you are way more experienced and have a clear idea for the code structure.  (Additionally: There are also cases where the "pure" recursive approach has advantages; I think if subtasks for different parent tasks that start in parallel have different runtimes, we will wait unnecessarily with my current implementation?).

I am still working on a better "real world example" and applications; i added a new and slightly more complex example that works fine, but the planner already outputs the task_ids in the correct order ...

jxnl:
i think the goal is more examples than full examples. i'll leave that to the framework companies.

task_planner_topological_sort.py
```diff
@@ -0,0 +1,193 @@
...
+import asyncio
+from pydantic import Field
+from typing import List, Generator
+from openai_function_call import OpenAISchema
+import random
+
+
+class TaskAnswer(OpenAISchema):
+    task_id: int
+    answer: str
```
jxnl, line 18:
```suggestion
    result: str
```

task_planner_topological_sort.py
```diff
@@ -0,0 +1,193 @@
...
+
+class Task(OpenAISchema):
+    """
+    Class representing a single task in a task plan.
+    """
+
+    id: int = Field(..., description="Unique id of the task")
+    task: str = Field(
+        ...,
+        description="Contains the task in text form. If there are multiple tasks, this task can only be answered when all dependant subtasks have been answered.",
```
jxnl, line 29:
```suggestion
        description="Contains the task in text form. If there are multiple tasks, this task can only be executed when all dependant subtasks have been answered.",
```

jxnl (approved):

jxnl:
looks good i think im happy to merge this!

jxnl (approved):

jxnl:
Merge whenever. Unless you have some other nits.

PR #12, opened 2023-06-21
Reorganize examples into directories

### LINKED ISSUE #11: Help: Reorganize module strucutre
would be nice to have a structure where theres a directory per example so we can have a readme.md for each example and a list of evals to run.

DaveOkpare:
Fixes #11 

Added `examples` directory which contains all examples as subdirectories.


jxnl:
are imports still working as is?

DaveOkpare:
> are imports still working as is?

Yes, the imports are working as-is.

jxnl:
Ok let's wait for the task planner PR to merge first and land this after wards

Could be nice to get toposort in the query planner example.

PR #19, opened 2023-06-22
Added example directory and fixed broken imports

DaveOkpare:
I added this block of code to all the examples. I couldn't come up with a fancier solution.

```python3
root_dir = dirname(dirname(dirname(abspath(__file__))))
sys.path.append(root_dir)

from openai_function_call import OpenAISchema
```

Also, there were some unused imports from `query_planner_execution.py` which I removed to make it cleaner


DaveOkpare:
Also, there were some unused imports from `query_planner_execution.py` which I removed to make it cleaner

Issue #10, opened 2023-06-21
Help: Add links to the advanced usage


jxnl:
adding the links from files to the readme would be helpful, also adding more code snippets.

Zeno1408:
@jxnl   which links?

jxnl:
In the main read me. Linking to files in examples.

ivanleomk:
Is this still open? If not happy to take a stab at it.

PR #25, opened 2023-06-26
Experimental Prompting DSL

jxnl:
Title: Advanced OpenAI API Interaction Abstraction

---

This pull request introduces new classes and methods to improve and simplify interactions with the OpenAI API. 

Key changes include:

1. New message classes for more flexible chat interactions.
2. `ChatCompletion` class to seamlessly build entire chat interactions.
3. `MultiTask` helper function to handle tasks involving multiple subtasks.

The changes provide robustness, flexibility, and a cleaner way to interact with the OpenAI API.

Screen Shot 2023-06-26 at 5 25 06 PM


experimental/promptbuilder/prompt_builder.py
```diff
@@ -0,0 +1,297 @@
...
+from openai_function_call import OpenAISchema
+
+
+class MessageRole(Enum):
+    USER = auto()
+    SYSTEM = auto()
+    ASSISTANT = auto()
+
+
+@dataclass
```
jpdus, line 16: is there a reason for using dataclasses here instead of subclassing BaseModel? (Not asking a a review, just for learning)

I would assume b/c of post_init, but is that really necessary if the `Message` class itself is only used as base class?

jpdus:
This looks very cool, like it a lot! 

3 minor comments: 
* I was really confused by the "or" notation until looking at the source; there might be a solution/notation that feels more natural?
* This would imho need some explanation/documentation of possible Prompt components right from the start (can by in comments)
* it might make sense to disentangle prompt components and function/output components for more clarity? It is cool that mixing up is possible but imho, this adds to the confusion - in the examples the output function and prompt components are not really related and i naturally asked myself why `MultiTask(subtask_class=Search)` came after ExpertSystem and before CoT....

jxnl:
> This looks very cool, like it a lot! 
> 
> 
> 
> 3 minor comments: 
> 
> * I was really confused by the "or" notation until looking at the source; there might be a solution/notation that feels more natural?
> 
> * This would imho benefit from some explanation/documentation of possible Prompt components right from the start (can be in comments)
> 
> * it might make sense to disentangle prompt components and function/output components for more clarity? It is cool that mixing up is possible but imho, this adds to the confusion - in the examples the output function and prompt components are not really related and i naturally asked myself why `MultiTask(subtask_class=Search)` came after ExpertSystem and before CoT....



1) data class just gets me post init as you said. Only way my tests pass was to add it to every class. In pydantic 2 there's a better way 

2) funny enought chaining like this is more natural to me. It's less cluttered than a.pipe(b).pipe(c) 

3) order does not matter rn. It's almost commutative composition. (Message history is trickier) 

But systems and cot could come last and it wouldn't matter.

fpingham:
Hey, this looks really awesome! One question I have though is what is the best way of combining several features in the same message since with the current abstractions each new object is a new message. Say we want to include tips into our system message, we could create a `SystemTips` class and build our system message like this:

```
SystemPersonality(name='helpful general assistant'')
SystemGuidelines(guidelines=['You never swear', 'You are polite and say please and thank you often', 'You only answer from the provided data']
SystemTask(task='Segment emails into search queries')
SystemTips(tips=[
                "When unsure about the correct segmentation, try to think about the task as a whole",
                "If acronyms are used expand them to their full form",
                "Use multiple phrases to describe the same thing",
            ]
```

And we would have `ChatCompletion` concatenate each of the object's contents in order to build a final system message. In this way the user can add unlimited abstractions that buildup to a single message and easily choose the order of how these components are combined to build the final message.

I have not thought of what's the best way to implement this but was wondering if it sounds useful before exploring further.

jxnl:
Would you want to try to PR or that?

Basically a light compiler that says: if there are multiple system  messages. Merge them into one rather than picking one or throwing and error?

Disiok:
Finally got around to look at this properly.

I see this approach as the opposite of Scott's guidance DSL: this being **code orchestrating prompt** components, guidance being **prompt orchestrating code** components. 

Some notes: 
* offering common prompt components as reusable functions is very nice
* Love the high-level declarative syntax, but not sure if developers are okay with being this abstracted away the final prompt. * It feels like the user still needs to understand how the primitives in this DSL maps to the underlying system message/function call to use this effectively. So unclear how leaky the abstraction is.
* The DSL makes a strong assumption that primitives in the DSL fall neatly into individual message, but not sure if assumptions hold well. I can see using a more powerful "compiler" to get around this, but not sure what that compiler should look like.

jxnl:
> 

could always do

`ChatComplete() | UserMessage("Long ass prompt") | Multi(task=TaskClass)`

you can also review the prompt by doing 

`completion.kwargs`

--

At a higher level im pretty tired of managing some blank tempkate strings being passed around, templated into templates. where as here, the template is first class

template, completiong arguments, tool use are all on the same level.

rather than having template arguments so 'far away' from templates and their strings.

jxnl:
> 

could always do

`ChatComplete() | UserMessage("Long ass prompt") | Multi(task=TaskClass)`

you can also review the prompt by doing 

`completion.kwargs`

--

At a higher level im pretty tired of managing some blank tempkate strings being passed around, templated into templates. where as here, the template is first class

template, completiong arguments, tool use are all on the same level.

rather than having template arguments so 'far away' from templates and their strings.

jxnl:
> 

could always do

`ChatComplete() | UserMessage("Long ass prompt") | Multi(task=TaskClass)`

you can also review the prompt by doing 

`completion.kwargs`

--

At a higher level im pretty tired of managing some blank tempkate strings being passed around, templated into templates. where as here, the template is first class

template, completiong arguments, tool use are all on the same level.

rather than having template arguments so 'far away' from templates and their strings.

roblox34:
Duplicate of #10

experimental/promptbuilder/prompt_builder.py
inayet: @jxnl Hi Jason, first, thank you for sharing your experience and wisdom, second, thank you again. 

Comment only from a novice, newbie, beginner's perspective - For examples for "functions" provide dynamic examples that iterates/finds the right function out of possibly hundreds of functions to send to ChatGPT, otherwise stuffing all those functions will reach content limit. I tried samantic search using pinecone but was a miss and a hit

PR #33, opened 2023-07-03
fixed completions endpoint and readme details for dsl example

fpingham:


jxnl (approved):
This is what I get for running out of openai credits and not being able to tests. 

Do you mind adding a test as well? Otherwise happy to merge.

PR #42, opened 2023-07-07
added better guidelines and trying a less repetitive instantiation me…

fpingham:


fpingham:
This is another way we can go about instatiating the different System classes: extra 'define' method but we can retain Pydantic compatibility.

No problem if you want to roll-back, just showing it as an option.

jxnl:
decided to make it simple and make functions to return  Messages

PR #61, opened 2023-07-13
Add refactoring example to gpt-engineer

cristobalcl:
Now there are two scripts in `gpt-engineer` folder.

`generate.py` creates a new program that is saved in JSON file.
`refactor.py` takes the previously generated JSON file and creates a diff file.


jxnl:
looks great @cristobalcl do you want to add markdown docs in the /docs page? i can take a stab some time otherwise

cristobalcl:
Sure! Let me add something to the “Creating Multiple Files Program” example, is that OK?

jxnl:
makes sense! anything else i can also patch later

jxnl (approved):

PR #82, opened 2023-08-20
Example: Integration of SQLAlchemy for Storing Chat Messages and Completions with Support for Any SQL Backend

jxnl:
# Introduction
This PR introduces the integration of SQLAlchemy to handle the persistence of chat messages and completions in any SQL backend. By leveraging SQLAlchemy, the chat interactions can be stored in a database system of choice, providing a flexible and robust way to manage and query the data. For testing purposes, an SQLite database will be used.

## Changes

1. **Database Models**: Defined SQLAlchemy models for `Message` and `ChatCompletion` to represent the chat messages and completions.
2. **Database Connection**: Added functionality to create a database engine compatible with any SQL backend, with the ability to define the connection string. SQLite is used as the default for testing.
3. **Insertion Functions**: Implemented functions to insert chat completions, messages, and responses into the database.
4. **Patching OpenAI Functions**: Added wrappers to patch OpenAI functions for synchronous and asynchronous chat completions, ensuring that the data is inserted into the database when these functions are called.
5. **Utility Functions**: Implemented helper functions to create `Message` objects from the chat data.
6. **Async Support**: Integrated asyncio to support asynchronous operations, allowing for non-blocking database inserts.

## TODO

- Refactor the `functions` and `function_call` fields in the `ChatCompletion` class to use foreign keys instead of storing them as strings.

## How to Test
To test the changes, you can run the script with the SQLite database engine, which will create the database file `chat.db` and the necessary tables. You can also modify the connection string to test with other SQL backends.

## Dependencies
- SQLAlchemy


rudolfolah:
This is very useful for production systems!

[In chaincrafter I have a branch that logs much of the same data from the OpenAI response](https://github.com/rudolfolah/chaincrafter/blob/logging-sqlite/python/chaincrafter/loggers.py#L23-L37) as your code (chat completion, messages).

It also logs the model parameters. Not sure how useful that would be for the use cases that `instructor` is designed for.

PR #101, opened 2023-09-10
Minor comment change

Phodaie:


Phodaie:
Just a minor change in commented print output

PR #110, opened 2023-09-22
Chain of density in examples 

jxnl:


examples/chain-of-density/run.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+
+# This multitask helper will be used to generate a chain of summaries.
+# Allows us to extract data via streaming to see resuls faster
+ChainOfDenseSummaries = instructor.MultiTask(
+    Summary,
+    name="chain-of-dense-summaries",
+    description="""
+        Repeat the following 2 steps 5 times.
```
andrewnc, line 43: How good are models at following this sort of "counting instruction?"

examples/chain-of-density/run.py
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    # Generate a chain of summaries, however we can also stream the results
+    # to see the results faster
+    for summary in summarize_article(example["text"]):
+        pprint(summary.model_dump())
+
+    """
```
andrewnc, line 90: Seems good!

examples/chain-of-density/run.py
jxnl: By adding that index it does ok lol

Issue #109, opened 2023-09-22
Decoupling the llm backend


veeragoni:
**Is your feature request related to a problem? Please describe.**
I see the library is tightly coupled with openai function calling. but it would be good to decouple the model from pydantic way of doing things and use any model (llms from langchain) that way we can experiment with smaller/self-hosted/other cloud models

**Describe the solution you'd like**
ability to pass pydantic structures to any llm and get results back. for eample, something like using langchain tools where function calling is isolated from llm.

**Describe alternatives you've considered**
custom tools in langchain implementation for function calling

**Additional context**
not sure its already possible. I haven't experimented yet, but it looks like its coupled based on the repo subtitles /examples

jxnl:
take a look at askmarvin.ai !

jxnl:
what is the issue with using the pydantic output parser in langchain or llamaindex?

PR #116, opened 2023-10-11
Add support for Llama2, Claude, Palm, Cohere, Replicate (100+ LLMs)

krrishdholakia:
Hi @jxnl @jlondonobo,

Noticed some users asking for other models - https://github.com/jxnl/instructor/issues/109. 

My PR shows users how to call their LLM providers (bedrock, togetherai, huggingface tgi, replicate, ai21, cohere, ai21 etc.) by pointing your openai endpoint to a [local proxy](https://docs.litellm.ai/docs/proxy_server) they can use for experimentation with self-hosted Sweep.

This makes **no code changes** to Instructor.

Happy to add additional tests/documentation if the initial PR looks good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,60 @@ model = openai.ChatCompletion.create(
 assert model.name == "JASON"
 ```
 
+## Use Anthropic,Huggingface,Palm,Ollama, etc.[Full List](https://docs.litellm.ai/docs/providers)
+
+### Create OpenAI-proxy
+We'll use [LiteLLM](https://docs.litellm.ai/docs/) to create an OpenAI-compatible endpoint, that translates OpenAI calls to any of the [supported providers](https://docs.litellm.ai/docs/providers).
+
+
+Example to use a local CodeLLama model from Ollama.ai with Instructor: 
+
+Let's spin up a proxy server to route any OpenAI call from Instructor to Ollama/CodeLlama
+
+```python
+pip install litellm
+```
+```python
+$ litellm --model ollama/codellama
+
+#INFO: Ollama running on http://0.0.0.0:8000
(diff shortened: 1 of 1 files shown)


jxnl:
wait all thes support function calls now?

krrishdholakia:
for the ones that aren't supported, you can set `--add_function_prompt` and litellm will catch that and map it to the system prompt instead

README.md
```diff
@@ -140,6 +140,60 @@ model = openai.ChatCompletion.create(
...
+    def validate_name(cls, v):
+        if v.upper() != v:
+            raise ValueError("Name must be in uppercase.")
+        return v
+
+model = openai.ChatCompletion.create(
+    model="gpt-3.5-turbo",
+    response_model=UserDetails,
+    max_retries=2,
+    api_base="http://0.0.0.0:8000",
```
jxnl, line 188: Can this be done via the `openai.api_base` as well?

README.md
```diff
@@ -140,6 +140,60 @@ model = openai.ChatCompletion.create(
...
+
+#INFO: Ollama running on http://0.0.0.0:8000
+```
+
+[Docs](https://docs.litellm.ai/docs/proxy_server)
+
+### Update Instructor
+
+```python
+import instructor
```
jxnl, line 167: are we missing an import?

jxnl:
1) can you add some examples like `example/lite-llm`
2) add it to the docs/index.md

jxnl:
> @krrishdholakia was this discarded/discontinued? Seemed like something I and others would want.

i believe you can use litellm and just patch the client as you would normally docs were not added but i believe they have support now

PR #121, opened 2023-10-23
Blog post: Validation is Validation

jxnl:
### Summary by CodeRabbit

- New Feature: Introduced the concept of "probabilistic validation" as an alternative to Constitutional AI, enhancing the software's intuitiveness and reliability.
- Documentation: Added blog posts and examples demonstrating the use of Pydantic and Instructor for validation in software development, improving user understanding.
- Refactor: Simplified and improved code for validation and error handling, boosting software performance and reliability.
- Style: Made minor changes to string formatting and import statements for better code readability and consistency.
- Test: Updated test cases to reflect changes in the code, ensuring software quality and reliability.


docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+tags:
+    - pydantic
+    - validation
+    - guardrails 
+    - constitutional ai
+---
+
+# Constitutional AI is just Probabilistic Validation
+
+Earlier this year, there was a lot of discussion about Constitutional AI and its role in guiding models. There was also talk about the need for guard rails and special systems to ensure safe outputs. However, I found myself unimpressed. Why create a new term when we already have one that describes the concept? That term is "validation," and now we can do it probabilistically.
```
eugeneyan, line 13: [NIT]

> That term is "validation," and now we can do it probabilistically.

Typically, we think of validation as being deterministic with a clear pass or fail. Is the date in the right format, or is the response within a certain length? In this case, I agree that validation and guardrails are synonymous, at least in the [original vision of the latter](https://github.com/guardrails-ai/guardrails#-rail-spec).

On the other hand, I think Constitutional AI, and your example in line 84, is subjective and thus probabilistic, as you rightly pointed out. Thus, I wonder if calling this validation would be overloading the term?

I like how you named it `llm_validator` as it mirrors `field_validator`. Nonetheless, I wonder if the following suggestions would help users better distinguish between validation as deterministic vs. moderation as subjective/probabilistic:
- `probabilistic_validator`: To make clear that it's not deterministic. Also, allows for non-LLM validations.
- `llm_moderator`: To make clear that it's not validation

Either way, not a one-way door. Even if `Instructor` sticks to `llm_validator`, it's a simple name change to update in future.

docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+```python
+from pydantic import BaseModel, field_validator, ValidationError
+from instruct import llm_validator
+
+class UserModel(BaseModel):
+    id: int
+    name: str
+    beliefs: str
+
+    @field_validator('beliefs')
```
eugeneyan, line 81: Running validation on beliefs is pretty cool. Might want to include one or two more examples of other types of validations.

docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+    - validation
+    - guardrails 
+    - constitutional ai
+---
+
+# Constitutional AI is just Probabilistic Validation
+
+Earlier this year, there was a lot of discussion about Constitutional AI and its role in guiding models. There was also talk about the need for guard rails and special systems to ensure safe outputs. However, I found myself unimpressed. Why create a new term when we already have one that describes the concept? That term is "validation," and now we can do it probabilistically.
+
+Tools like Constitutional AI address shortcomings by using AI feedback to evaluate outputs. In this approach, an AI system provides a set of principles for making judgments about generated text. These principles guide the model's behavior. Essentially, it's a form of validation. Instead of manually checking against a set of rules, we now have a model that can perform the validation. For example, a content moderation system could use a model to automatically ban certain keywords. Previously, an error might have been thrown upon detection, but now we have a self-correcting system.
```
eugeneyan, line 15: Reading this again, I wonder if starting with this paragraph (and discarding the paragraph before) might be more punchy.

docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+Validation is a fundamental concept in Software 1.0. However, when we want an ai system to raise an error or self-correct, we introduce new vocabulary and terms. Instead, we can apply the same idea of having types, such as an integer or a string, but with additional constraints. For example, a string type that is not "an apology" or "a threat." The underlying principles remain the same.
+
+```python
+def validation_function(cls, value):
+    if condition(value):
+        raise ValueError("Value is not valid")
+    return mutation(value)
+```
+
+We should be able to define new types that are powered by probabilsitic models and use them in the same way we use define validators in Pydantic.
```
rlouf, line 66:
```suggestion
We should be able to define new types that are powered by probabilistic models and use them in the same way we use define validators in Pydantic.
```

docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+    id: int
+    name: str
+    beliefs: str
+
+    @field_validator('beliefs')
+    @classmethod
+    def moderate_beliefs(cls, v: str) -> str:
+        validator = llm_validator("don't say objectionable things")
+        return validator(v)
+```
```
jphme, line 115: This seems to be the same code as above. .

docs/blog/posts/constitutional-ai-or-validation.md
jxnl: Yeah even fuzzy validator might make more sense.

docs/blog/posts/constitutional-ai-or-validation.md
jxnl: Good call on using different examples. 

I should include one that checks the chain of thought makes sense.

docs/blog/posts/constitutional-ai-or-validation.md
jxnl: Ah thanks. I meant to add more validators to show how it's done.

docs/blog/posts/constitutional-ai-or-validation.md
jacoblee93, line 15: I tend to think of validation in as being quick, synchronous, and light vs. verification as heavierweight, at least in a DB context, but that might just be me

docs/blog/posts/constitutional-ai-or-validation.md
jacoblee93, line 81: This is sweet!

docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+class UserModel(BaseModel):
+    id: int
+    name: str
+
+    @field_validator('name')
+    @classmethod
+    def name_must_contain_space(cls, v: str) -> str:
+        if ' ' not in v:
+            raise ValueError('must contain a space')
+        return v.title()
```
adriangb, line 41: I know this is an example but it'd be interesting if you used more PEP 593 validated types:

```python
from typing import Annotated

from pydantic import AfterValidator, BaseModel

def name_must_contain_space(cls, v: str) -> str:
        if ' ' not in v:
            raise ValueError('must contain a space')
        return v.title()

Name = Annotated[str, AfterValidator(name_must_contain_space)]
"""A name. It must not contain a space!"""

class UserModel(BaseModel):
    id: int
    name: Name
```

docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+class UserModel(BaseModel):
+    id: int
+    name: str
+    beliefs: str
+
+    @field_validator('beliefs')
+    @classmethod
+    def moderate_beliefs(cls, v: str) -> str:
+        validator = llm_validator("don't say objectionable things")
+        return validator(v)
```
adriangb, line 85: Is there a world where you want an async call (as in `asyncio` / `await thing()`)? If the model is running locally or in many other cases I imagine no. But if this is a call to web service then it could make a lot of sense. It's not something Pydantic supports but just saying...

docs/blog/posts/constitutional-ai-or-validation.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+### Step 2: Using the Client with Retries
+
+In the following code snippet, the `UserModel` is specified as the `response_model`, and `max_retries` is set to 2.
+
+```python
+model = openai.ChatCompletion.create(
+    model="gpt-3.5-turbo",
+    response_model=UserModel,
```
adriangb, line 207: Does this mean that OpenAI knows how to integrate with Pydantic?

docs/blog/posts/constitutional-ai-or-validation.md
jxnl: yeah my docs end u using that, 

```
AfterValidator(llm_validator("..."))
```

Do you think its confusing to do both?

docs/blog/posts/constitutional-ai-or-validation.md
jxnl: `instructor.patch()` added this should leave a comment

docs/blog/posts/constitutional-ai-or-validation.md
adriangb: No it's not, just that I find now I almost always want to use the `Annotated` version

docs/blog/posts/constitutional-ai-or-validation.md
jxnl: i think in prod settings i'd want async fanouts + timeouts 
but i actually think the correct path is distilation into a finetune model where the validators is just about getting the training data.

PR #125, opened 2023-10-26
Validation part 1 proposed edits

ivanleomk:
### Summary by CodeRabbit

- Documentation: Enhanced the blog post on dynamic validation, providing clearer explanations and more comprehensive examples.
- New Feature: Introduced a custom `llm_validator` for more flexible validation using language models.
- Improvement: Added a patching mechanism to the `ChatCompletion` class for easier integration with OpenAI models and Pydantic objects.
- Improvement: Implemented a `max_retries` parameter to handle invalid responses more effectively.
- New Feature: Demonstrated the use of field and model validators for robust data validation.
- Improvement: Showcased how to handle errors and re-asking in the validation process.


docs/blog/posts/validation-part1.md
```diff
@@ -47,12 +47,21 @@ user: UserDetail = openai.ChatCompletion.create(
...
+assert user.name == "Jason" # (3)!
 assert user.age == 25
 ```
 
+1.  To simplify your work with OpenAI models and streamline the extraction of Pydantic objects from prompts, we
+    offer a patching mechanism for the `ChatCompletion` class.
+
+2.  Invalid responses that fail to be validated succesfully will trigger up to as many reattempts as you define.
+
+3.  As long as you pass in a `response_model` parameter to the `ChatCompletion` api call, your response object will
```
jxnl, line 62:
```suggestion
3.  As long as you pass in a `response_model` parameter to the `ChatCompletion` api call, the returned object will always
```

docs/blog/posts/validation-part1.md
```diff
@@ -264,9 +278,11 @@ def validate_chain_of_thought(values):
 
 We can then take advantage of the `model_validator` decorator to perform a validation on a subset of the model's data.
 
-> We're defining a model validator here which runs before pydantic parses the input into its respective fields. That's why we have a **before** keyword used in the `model_validator` class.
+> We're defining a model validator here which runs before `Pydantic` parses the input into its respective fields. That's why we have a **before** keyword used in the `model_validator` class.
 
 ```python
+from pydantic import BaseModel,model_validator
```
jxnl, line 284:
```suggestion
from pydantic import BaseModel, model_validator
```

docs/blog/posts/validation-part1.md
```diff
@@ -298,24 +314,22 @@ If we create a `AIResponse` instance with an answer that does not follow the cha
...
 ```
 
-Beyond validating multiple attributes of a model, we can also introduce context to our validation functions, in order to give our models more information to work with.
-
 ### Validating Citations From Original Text
 
-Let's see a more concrete example. Let's say we use RAG to answer a question given a text chunk, in earlier systems we'd just simply tell you the chunk id and you'd have to go and find the text chunk yourself. But what if we could validate that the answer is actually supported by the text chunk? We can do this by passing the text chunk as a context to the validator.
-
-Pydantic allows us to do so easily by utilising a context object. This is an arbitrary dictionary which you can access inside the `info` argument in a decorated validator function.
+Let's see a more concrete example. Let's say that we've asked our model a question about some text source and we want to validate that the generated answer is supported by the source. While we could verify this by looking up the original source manually, a more scalable approach is to use a validator to do this automatically.
```
jxnl, line 319:
```suggestion
Let's see a more concrete example. Let's say that we've asked our model a question about some text source and we want to validate that the generated answer is supported by the source. This would allow us to minimize hallucinations and prevent statements that are not backed by the original text. While we could verify this by looking up the original source manually, a more scalable approach is to use a validator to do this automatically.
```

jxnl (approved):

PR #135, opened 2023-11-04
Chain of density

ivanleomk:
Added a sample implementation of the Chain of Density implementation. Modified pipeline slightly to use a mix of GPT4 for critiquing and GPT3.5 for generation of output

## Summary by CodeRabbit

- **New Features**
  - Introduced a new "Chain Of Density" summarization technique using GPT-3.5.
  - Added a new Python script for article summarization using Pydantic and OpenAI's GPT-4 model.

- **Documentation**
  - Added a comprehensive guide on implementing the "Chain Of Density" summarization technique.
  - Provided instructions for installing dependencies, downloading datasets, generating examples for fine-tuning, and running a fine-tuning job.

- **Bug Fixes**
  - Fixed a typo in the model parameter in the `openai.ChatCompletion.create()` function.

- **Chores**
  - Updated Python version used in the workflow from 3.x to 3.10.
  - Added new dependencies: openai, pydantic, instructor, nltk, and rich.


examples/chain-of-density/chain_of_density.py
```diff
@@ -0,0 +1,223 @@
...
+            omitted_entities = [
+                entity.entity_name for entity in evaluation.omitted_entities
+            ]
+            raise ValueError(
+                f"The following entities were omitted from the summary - {','.join(omitted_entities)}. Please generate a new summary which does not omit them."
+            )
+
+        return self
+
+
```
jxnl, line 121: Consider using tenacity. Or rate limit. (The libraries)

examples/chain-of-density/chain_of_density.py
```diff
@@ -0,0 +1,223 @@
...
+    article: str,
+    existing_summary: str,
+    entity_ctx: str,
+    error_msgs: List[str] = [],
+    remaining_retries=3,
+):
+    # # We then perform a new summary and validate that the entity density has increased ( We have not lost any entities )
+    completion = openai.ChatCompletion.create(
+        model="gpt-3.5-turbo-16k",
+        functions=[RewrittenSummary.openai_schema],
```
jxnl, line 132: Any reason response_model does not work here?

examples/chain-of-density/chain_of_density.py
```diff
@@ -0,0 +1,223 @@
...
+            max_retries=2,
+            max_tokens=1000,
+        )
+        missing_entities: List[MissingEntities] = [
+            i for i in MissingEntities.from_response(missing_entities_response)
+        ][0][1]
+
+        entity_ctx = ""
+        for entity in missing_entities:
+            assert isinstance(entity, MissingEntity)
```
jxnl, line 218: Error message

examples/chain-of-density/chain_of_density.py
```diff
@@ -0,0 +1,223 @@
...
+            {"role": "user", "content": f"Here is the Article: {article}"},
+        ],
+        max_retries=2,
+    )
+    summary_chain.append(summary.summary)
+
+    # # Then we perform a denser summarization
+    for _ in range(summary_steps):
+        # At the end of each step, we compare the most recent summarization and see what additional missing entities we need to perform
+        missing_entities_response = openai.ChatCompletion.create(
```
jxnl, line 197: Goal is to use Spacy here right?

examples/chain-of-density/chain_of_density.py
```diff
@@ -0,0 +1,223 @@
...
+        return rewrite_summary(
+            article,
+            existing_summary,
+            entity_ctx,
+            error_msgs,
+            remaining_retries=remaining_retries - 1,
+        )
+
+
+def summarize_article(article: str, summary_steps: int = 3):
```
jxnl, line 177: Think this and rewrite summary could be one function.

examples/chain-of-density/chain_of_density.py
```diff
@@ -0,0 +1,223 @@
...
+    An entity is a real-world object that's assigned a name - for example, a person, country a product or a book title.
+    """
+
+    entity_name: str = Field(
+        ...,
+        description="This is an entity which was present in the previous summary and not in the newly generated summary",
+    )
+
+
+class MaybeOmittedEntities(BaseModel):
```
jxnl, line 45: Do you really need a maybe there. The fact that it's a list means if it's missing it just won't be in the list.

examples/chain-of-density/chain_of_density.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    This represents whether the new summary has omitted any entities that were present in the previous summary provided.
+    """
+
+    omitted_entities: Optional[List[OmittedEntity]] = Field(default=[])
```
jxnl, line 50: Use default_factory=list 

Never use defaults as [] (you can google to see why)

jxnl, line 50: Why not list[str]

jxnl (changes requested):
How close are the prompts here to the original prompts ?

ivanleomk:
> How close are the prompts here to the original prompts ?

Prompts here are based off the original prompts from the paper and are modified slightly. There are a bunch of them which I came up on my own but here are the rough annotations from the original paper


CleanShot 2023-11-04 at 23 22 49

> Figure 2 displays the prompt along with an example output. Rather than be prescriptive about the types of entities, we simply define a Missing Entity as:
> • Relevant: to the main story.
> • Specific: descriptive yet concise (5 words or fewer).
> • Novel: not in the previous summary.
> • Faithful: present in the Article.
> • Anywhere: located anywhere in the Article.

Initial Summary comes from the line about

> The first summary should be long ( 4 - 5 sentences, ~80 words ) yet highly non-specific, containing little information beyond the entities marked as missing. Use overly verbose language and fillers (e.g., "this article discusses") to reach ~80 words.

Rewritten Prompts come from the few paragraph points about

> Guidelines:
> - Make every word count: re-write the previous summary to improve flow and make space for additional entities.
> - Make space with fusion, compression, and removal of uninformative phrases like "the article discusses"
> - The summaries should become highly dense and concise yet self -contained, e.g., easily understood without the Article.
> - Missing entities can appear anywhere in the new summary.
> - Never drop entities from the previous summary. If space cannot be made, add fewer new entities

I rephrased the third point about the highly dense and concise summaries and used it as the description for the Pydantic class

Entity Definition comes from [spaCY](https://spacy.io/usage/linguistic-features#named-entities)

> An entity is a real-world object that's assigned a name - for example, a person, country a product or a book title.

Missing Entities Task comes from the step 1 of the prompt

> Identify 1-3 informative entities (";" delimited) from the article which are missing from the previously generated summary

examples/chain-of-density/chain_of_density.py
ivanleomk: Hmm I thought that using an OmittedEntity class would provide more contextual information originally since it gives 
a clear description of what the model is supposed to represent. Will try experimenting with a list[str] to see if I can get similar results

ivanleomk: Interesting, didn't realise that using default as [] would result in a single mutable value used for all future instances rather than a new fresh instance. Will make the necessary modifications

Source: https://www.perplexity.ai/search/0bc6fe17-f23f-46e8-b2cf-12bba4163d9d?s=u

examples/chain-of-density/chain_of_density.py
ivanleomk: I think moving forward will revert back to the original model which used omitted_entities and included_entities with default factories of list. This will have the same impact and we can do the `.missing` validation that we talked about,

examples/chain-of-density/chain_of_density.py
ivanleomk: Hmm was originally using the model_validate() function which required the use of a context. This didn't have a native retry from what I could see since I had to parse it using `new_summary = RewrittenSummary.from_response(..code)` and so I wrote up a new function for it.

Will simplify the logic and move away from using context validation here.

examples/chain-of-density/chain_of_density.py
ivanleomk: Found spacy a bit unreliable in terms of entity detection even with the 500mb large model and so I figured I would switch back to a llm validator

examples/chain-of-density/chain_of_density.py
ivanleomk: Okie, will add it in

examples/chain-of-density/chain_of_density.py
ivanleomk: Wanted to use the context validation here to validate the schema and the docs showed this method so I used it.

Reference: https://jxnl.github.io/instructor/examples/exact_citations/?h=fieldvalidation#validation-method-validate_sources

examples/chain-of-density/chain_of_density.py
ivanleomk: Okie

examples/chain-of-density/chain_of_density.py
jxnl: ah i should clena up the docs to use `response_model` as much as possible

examples/chain-of-density/chain_of_density.py
jxnl: yeah keep it simple, ust List[str] and ues the field descriptor instead.

examples/chain-of-density/chain_of_density.py
ivanleomk: Sure thing man.

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,354 @@
...
+  - validation
+  - chain of density
+authors:
+  - jxnl
+  - ivanleomk
+---
+
+# Implementing Chain Of Density
+
+By the end of this article, you'll have a fine-tuned GPT 3.5 model which can take in source text and output a summary which mimics a chain-of-density summarization using `Instructor`'s new jobs cli feature. As usual, all of the code is avaliable for reference under our `examples/chain-of-density` folder.
```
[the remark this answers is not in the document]
jxnl: haha

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,354 @@
+---
+draft: False
+date: 2023-11-05
+tags:
+  - pydantic
+  - validation
+  - chain of density
+authors:
+  - jxnl
```
jxnl, line 9: i think order matters, you should be first.

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,354 @@
+---
+draft: False
+date: 2023-11-05
+tags:
+  - pydantic
+  - validation
+  - chain of density
```
jxnl, line 7:
```suggestion
  - chain of density
  - finetuneing
  - gpt-3.5-turbo
  - distilation
```

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,354 @@
...
+  <figcaption>Improved process with Instructor</figcaption>
+</figure>
+
+### Data Modelling
+
+#### Initial Summary
+
+Let's start by walking through some of the data models that we'll be using as the `response_model` for our open ai function calls
+
+Firstly, we'll need a data model for the initial summary that we will be generating. We'll take the description of this class straight from the original prompt.
```
jxnl, line 89:
```suggestion
Firstly, we'll need a data model for the initial summary that we will be generating. We'll take the description of this class straight from the original prompt. Its important to note that these docstrings serve a purpose, they are directly used by the LLM when generating the outputs.
```

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,354 @@
...
+        default_factory=list,
+        description="this is a list of Entities found absent from the new summary that were present in the previous summary",
+    )
+    missing: List[str] = Field(
+        default_factory=list,
+        description="This is a list of 1-3 informative Entities from the Article that are missing from the new summary which should be included in the next generated summary.",
+    )
+```
+
+!!! Using Validators in Instructor
```
jxnl, line 138:
```suggestion
!!! notes "Using Pydantic Validators with Instructor"
```

examples/chain-of-density/chain_of_density.txt
```diff
@@ -0,0 +1,9 @@
+openai
```
jxnl, line 1: lets call this `requirements.txt` so its more clear

ivanleomk:
Migrated and updated with benchmarks that I calculated, also created a Hugging Face data card here - https://huggingface.co/datasets/ivanleomk/gpt4-chain-of-density.

I will tidy up a colab notebook so that people can reproduce my benchmarks using the data ( just the visualisations ) so you can double check my figures too just in case if you want.

Am still pending some new lib with the upgrade to Open AI library and changes to the ReadMe to add in the hugging face dataset links

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,408 @@
...
+| ------------------- | ------------- | -------------- | ----------- |
+| 3.5 Finetuned (20)  | 0.664         | 0.15           | 43,612      | 
+| 3.5 Finetuned (50)  | 1.112         | 0.153          | 45,128      |
+| 3.5 Finetuned (100) | 2.328         | 0.153          | 44,925      |
+| GPT-4 Turbo         | -             | 1.41           | 265,397     |
+| GPT-4               | -             | 7.63           | 238,290     |
+
+!!! note
+
+    Using a fine-tuned model was able to reduce the inference time by almost 20-40x while keeping entity density relatively constant. At the same time, our costs dropped by almost 10x when compared against `GPT-4 Turbo` and by almost 150x when compared against GPT-4. This is a strong nod to the power of fine-tuning and generating small models to perform niche tasks.
```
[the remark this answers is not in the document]
jxnl: +1

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,408 @@
+---
+draft: False
+date: 2023-11-05
+tags:
+  - pydantic
+  - validation
+  - chain of density
+  - finetuneing
+  - gpt-3.5-turbo
+  - distilation
```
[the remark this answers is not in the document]
jxnl: +1 whoops

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,467 @@
...
+  <figcaption>Improved process with Instructor</figcaption>
+</figure>
+
+### Data Modelling
+
+#### Initial Summary
+
+Let's start by walking through some of the data models that we'll be using as the `response_model` for our open ai function calls
+
+Firstly, we'll need a data model for the initial summary that we will be generating. We'll take the description of this class straight from the original prompt. Its important to note that these docstrings serve a purpose, they are directly used by the LLM when generating the outputs.
```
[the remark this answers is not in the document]
eugeneyan: > Its important to note that these docstrings serve a purpose, they are directly used by the LLM when generating the outputs.

Make it clearer why the docstrings are important. I _think_ it's used to [validate](https://github.com/jxnl/instructor/blob/main/instructor/function_calls.py#L91) that initial and rewritten summaries contain the expected Pydantic fields (summary, absent, missing). If so, then do we need such extensive docstrings? Also, the doctrings seem duplicative of the system content in the examples below.

Please correct my understanding if it's wrong. Either way, I think it's an opportunity to educate users on how docstrings come into play in `Instructor`.

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,467 @@
...
+
+```py hl_lines="2 9 11-18 37 40"
+from typing import List
+from chain_of_density import summarize_article #(1)!
+import csv
+import logging
+import instructor
+from pydantic import BaseModel
+
+logging.basicConfig(level=logging.INFO) #(2)!
```
eugeneyan, line 335: This feels like an unnecessary detail that the user has to be careful not to trip over—perhaps `instructor.Instructions` can have it's own logging handler so the user doesn't have to even be aware of it?

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,467 @@
...
+- Entity Density : This is entities per token, the higher the better for density.
+- Latency : Time to last token generated in seconds
+- Costs : How much does the entire experiment cost
+
+We used a total of 20 articles as a validation set which our fine tuned models had not seen before. This was the overall performance that we observed.
+
+| Model               | Mean Latency (s) | Mean Entity Count | Mean Entity Density | Tokens |
+| ------------------- | ---------------- | ----------------- | ------------------- | ------ |
+| GPT-4 (COD)         | 49.5             | 11.3              | 0.138               | 81.65  |
+| GPT-3 (COD)         | 145.94           | 11.05             | 0.105               | 105.7  |
```
eugeneyan, line 437: Is this gpt-3 or gpt-3.5? If the former, I wonder if we can add a benchmark for gpt-3.5? If the latter, make it clearer.

Also, we can make the benefits of COD + distillation more convincing by adding benchmarks for gpt-4 and gpt-3.5 _without_ COD—how much lift does this additional effort buy us? This will help users prioritize between different approaches to improve their summaries.

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,467 @@
...
+| 3.5 Finetuned (20)  | 2.25             | 14.7              | 0.154               | 95.45  |
```
eugeneyan, line 438: Comparing finetuning on 20 summaries to the rows below, it seems that finetuning only on 20 summaries had the highest absolute entity count and density? Hmm, why is that? Might want to add a few hypotheses even if you don't have the answers.

docs/blog/posts/chain-of-density.md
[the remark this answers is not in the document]
ivanleomk: To my understanding, the docstrings are used in the function call parameters (Eg below) so a more descriptive docstring helps guide the eventual output by cleanly specifying what you want 

```
{
  "functions": [
    {
      "name": "GeneratedSummary",
      "description": "This represents a highly concise summary that includes as many entities as possible from the original source article.\n\nAn Entity is a real-world object that's assigned a name - for example, a person, country a product or a book title.\n\nGuidelines\n- Make every word count\n- The new summary should be highly dense and concise yet self-contained, eg., easily understood without the Article.\n- Make space with fusion, compression, and removal of uninformative phrases like \"the article discusses\"",
      "parameters": {
        "properties": {
          "summary": {
            "description": "This represents the final summary generated that captures the meaning of the original article which is as concise as possible. ",
            "title": "Summary",
            "type": "string"
          }
        },
        "required": [
          "summary"
        ],
        "type": "object"
      }
    }
  ]
}
}
```

docs/blog/posts/chain-of-density.md
ivanleomk: @jxnl thoughts?

docs/blog/posts/chain-of-density.md
ivanleomk: Ah, this is GPT-3.5, my bad on this. I took this out and replaced it with the benchmarks I calculated for a Vanilla summary using GPT 3.5 that just asked for a concise summary.

docs/blog/posts/chain-of-density.md
ivanleomk: Added it in a new branch! But my hypotheses are that

1. Model might not be benefitting from the higher number of examples due to the epochs ( 20, 50 and 76 are all trained with 4 epochs which is the default number provided by OpenAI)
2. Larger variety of examples might cause the model to optimize for different objectives - not just for token density. The COD summarization method tends to produce more abstract summaries with each rewrite, so it might be learning other metrics under the hood to optimize for

docs/blog/posts/chain-of-density.md
```diff
@@ -0,0 +1,467 @@
...
+
+By the end you'll end up with a GPT 3.5 model, (fine-tuned using Instructor's great tooling), capable of producing summaries that rival the effectiveness of Chain of Density. As always, all code is readily available in our `examples/chain-of-density` folder in our repo for your reference.
+
+??? abstract "Datasets and Colab Notebook"
+
+    We've also uploaded all our generated data to Hugging Face [here](https://huggingface.co/datasets/ivanleomk/gpt4-chain-of-density) for you to use if you'd like to try reproducing these experiments. We've also added a [Colab Instance](https://colab.research.google.com/drive/1iBkrEh2G5U8yh8RmI8EkWxjLq6zIIuVm?usp=sharing) for you to check our generated values.
+
+## Part 1) Chain of Density
+
+Summarizing extensive texts with AI can be challenging, often relying on inconsistent techniques. Salesforce AI Research's novel method, chain of density, enhances AI-based text summarization, outperforming human-generated summaries.
```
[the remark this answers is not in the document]
ivanleomk: Fixed up

PR #136, opened 2023-11-04
Azure support multitask.py

zboyles:
Azure support for Multiple Extractions

Without validation, Azure use breaks with some chunks not containing any `choices` and the first `function_call` chunk not containing `arguments`.

As of API version "2023-07-01-preview", most chunks related to a function call seem to spit out all at once. So be aware the experience won't match OpenAI until it's addressed in a future API update.

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling in the new task extraction process to prevent potential issues and improve overall app stability.


jxnl:
it's now in 0.2.10

PR #145, opened 2023-11-05
fix: match the monkey-patched method signature with upstream openai

teedrupp:
When instructor monkey-patches the create and acreate methods, it changes the type of the function from 'classmethod' to 'function'. This works fine in most cases, but when other tools attempt to, themselves, wrap the OpenAI class methods, this produces an error because those tools are expecting classmethods; not functions.

The example I encountered is using the python "wrapt" library/module. When wrapping the monkey-patched code, it produces the following error

  TypeError: missing 1 required positional argument

This patch changes the method signature to match the one that OpenAI themselves use.

I've included tests to ensure that the existing instructor behavior remains unchanged. Additionally, I tested with the tests that are marked with pytest unimplemented marks to ensure that the continued class instances are correct.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the chat completion function to support class methods.
- **Tests**
	- Added and updated tests to verify the chat completion function's behavior with and without patching.

### DIFF
--- a/instructor/patch.py
+++ b/instructor/patch.py
@@ -141,7 +141,7 @@ def wrap_chatcompletion(func: Callable) -> Callable:
 
     @wraps(func)
     async def new_chatcompletion_async(
-        response_model=None, validation_context=None, *args, max_retries=1, **kwargs
+        cls, response_model=None, validation_context=None, *args, max_retries=1, **kwargs
     ):
         response_model, new_kwargs = handle_response_model(response_model, kwargs)  # type: ignore
         response, error = await retry_async(
@@ -158,7 +158,7 @@ def wrap_chatcompletion(func: Callable) -> Callable:
 
     @wraps(func)
     def new_chatcompletion_sync(
-        response_model=None, validation_context=None, *args, max_retries=1, **kwargs
+        cls, response_model=None, validation_context=None, *args, max_retries=1, **kwargs
     ):
         response_model, new_kwargs = handle_response_model(response_model, kwargs)  # type: ignore
         response, error = retry_sync(
@@ -175,7 +175,7 @@ def wrap_chatcompletion(func: Callable) -> Callable:
 
     wrapper_function = new_chatcompletion_async if is_async else new_chatcompletion_sync
     wrapper_function.__doc__ = OVERRIDE_DOCS
-    return wrapper_function
+    return classmethod(wrapper_function)
 
 
 original_chatcompletion = openai.ChatCompletion.create
--- a/tests/test_patch.py
+++ b/tests/test_patch.py
@@ -1,7 +1,7 @@
 from pydantic import BaseModel
 import pytest
 import openai
-from instructor import patch
+from instructor import patch, unpatch
 
 
 @pytest.mark.skip("Not implemented")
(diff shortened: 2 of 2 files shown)


teedrupp:
to add some context to the changes, here are links to the wrapt docs that seem to point out this being a potential problem for them.

* https://github.com/GrahamDumpleton/wrapt/blob/develop/src/wrapt/wrappers.py#L718
* https://github.com/GrahamDumpleton/wrapt/blob/develop/src/wrapt/wrappers.py#L723
* https://github.com/GrahamDumpleton/wrapt/blob/develop/src/wrapt/wrappers.py#L741
* https://github.com/GrahamDumpleton/wrapt/blob/develop/src/wrapt/wrappers.py#L749

I tried some number of other ways to work around this quirk (such as patching after binding) but it was only by changing to return a `classmethod` that I was able to repeatedly get success without apparently changing the existing behavior of instructor (afaik, given the available tests).

Hope this is something you'd consider merging.

jxnl:
I'll pull this in an check it myself. 

Do you know what other tools that also wrap it look like? I might consider changing the pattern all together

jxnl:
would love your thoughts on #146

teedrupp:
@jxnl i'm only familiar with the `wrapt` class at this point. here's a short snippet you can toy with that exhibits the behavior (albeit without using instructor specifically). I've included the cases that I was trying out so you can uncomment them and hopefully see similar.

My reasons for wrapping the OpenAI calls was because I needed to track their usage to record prompt and completion tokens, insofar as it relates to spend. I seem to be able to do this after the function signature returned by instructor is changed to `classmethod`.

I still need to call `wrapt` after the monkey-patching, but that's acceptable in my case.

```
import wrapt
from types import MethodType


def tracker(wrapped, instance, args, kwargs):
    print("bar")
    print(wrapped, instance, args, kwargs)
    resp = wrapped(*args, **kwargs)
    return resp


def baz(*args, **kwargs):
    print("baz")


class Foo:
    @classmethod
    def bar(cls):
        print("foo")


# case 1, wrapt with normal call to Foo.bar
# should observe the following (expected)
#
#    bar
#    <bound method Foo.bar of <class '__main__.Foo'>> <class '__main__.Foo'> () {}
#    foo
#
#wrapt.wrap_function_wrapper('__main__', 'Foo.bar', tracker)
#Foo.bar()

# case 2, monkey patch "deletes" the wrapt decorator
# should observe the following (expected because of how Python works, but unexpected because
# I didnt know that instructor monkey-patched until I read the code).
#
#    baz
#
#wrapt.wrap_function_wrapper('__main__', 'Foo.bar', tracker)
#Foo.bar = baz
#Foo.bar()

# case 3, wrapt after the monkey patching. function patched into classmethod
# should observe the following
#
# TypeError: missing 1 required positional argument
#
#Foo.bar = baz
#wrapt.wrap_function_wrapper('__main__', 'Foo.bar', tracker)
#Foo.bar()

# case 4, changing function to classmethod
# should observe the following
#
#    bar
#    <bound method baz of <class '__main__.Foo'>> <class '__main__.Foo'> () {}
#    baz
#
# This is the behavior I'm trying to get. By wrapping the calls to OpenAI, I'm seeking to "track" their usage
#
#Foo.bar = classmethod(baz)
#wrapt.wrap_function_wrapper('__main__', 'Foo.bar', tracker)
#Foo.bar()

# case 5, changing function to classmethod, but wrapt before monkey-patching
# should observe the following
#
#    baz
#
# this is a similar to case 2, but the monkey-patch is a classmethod instead of a function
#
#wrapt.wrap_function_wrapper('__main__', 'Foo.bar', tracker)
#Foo.bar = classmethod(baz)
#Foo.bar()
```

teedrupp:
> would love your thoughts on #146

i'll take a look

jxnl:
i updated to openai  1.0.0 and added the cls stuff

teedrupp:
cool. i'll take a look. if you want to close this, that's cool. otherwise i'll close it after i have a look at 0.3.0

PR #175, opened 2023-11-13
fixes retry_async message unpacking

### LINKED ISSUE #174: Validation error retry hotfix needs to be applied to `retry_async`
**Describe the bug**
Retrying validation errors with async fails. I believe this is due to the OpenAI sdk update to 1.0. When `retry_async` tries to handle a `ValidationError` or `JSONDecodeError`, it attempts to append the previous message to the message chain. But with the update to OpenAI `dict(**response.choices[0].message)` now fails with the error `dict() argument after ** must be a mapping, not ChatCompletionMessage`.

Presumably this same issue applied to `retry_sync`, but it was fixed with the [hotfix commited here](https://github.com/jxnl/instructor/commit/877f57a2baea1766774d028b7fc976979ef83129#diff-1ccc47ac75961e3a4f37760e35da89919e2497efffdb67f3e5b6675a419c44b5). A similar fix to `retry_async` should be all that is necessary to resolve the issue.

bradenkinard:
This fixes #174 by applying the same message appending logic as in [the hotfix made here](https://github.com/jxnl/instructor/commit/877f57a2baea1766774d028b7fc976979ef83129#diff-1ccc47ac75961e3a4f37760e35da89919e2497efffdb67f3e5b6675a419c44b5) for `retry_sync`.

## Summary by CodeRabbit

- **Refactor**
	- Improved error handling and message storage in the retry mechanism for asynchronous operations.

### DIFF
--- a/instructor/patch.py
+++ b/instructor/patch.py
@@ -89,7 +89,7 @@ async def retry_async(
                 None,
             )
         except (ValidationError, JSONDecodeError) as e:
-            kwargs["messages"].append(dict(**response.choices[0].message))  # type: ignore
+            kwargs["messages"].append(response.choices[0].message)  # type: ignore
             kwargs["messages"].append(
                 {
                     "role": "user",


jxnl:
thanks will add some more tests so this does not happen next time

jxnl:
https://github.com/jxnl/instructor/releases/tag/0.3.4

PR #197, opened 2023-11-19
expositional review

MeDott29:
@coderabbitai provide some suggestions for improving the transition into '## Introduction to Function Calling'
should there be a side note about what it means to call a class with another class?

## Summary by CodeRabbit

- **Documentation**
  - Updated the tutorial to reflect the use of an earlier version of the Python package, ensuring compatibility with the current environment.

### DIFF
--- a/tutorials/1.introduction.ipynb
+++ b/tutorials/1.introduction.ipynb
@@ -617,7 +617,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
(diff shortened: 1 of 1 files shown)


MeDott29:
@coderabbitai let's look at explaining the transition between two secitons

MeDott29:
@coderabbitai help describe the transition between the last code block in '## Asking for JSON from OpenAI' and the beginning of '## Introduction to Function Calling' in 'introduction.ipynb'

PR #221, opened 2023-11-25
Support Streaming MultiTask with response_model

### LINKED ISSUE #213: Support Streaming MultiTask with `response_model`
Right now, the streaming feature doesn't work with the `response_model` in Multitask. Instead of `model.from_streaming_response() -> List[obj]`, we can consider a more elegant approach like the following:

```python
class User(BaseModel):
    name: str
    age: int

users: Iterable[User] = openai.chat.completion.create(
    ...
    stream=True,
    response_model=Iterable[User]
)

for user in users:
    ...
```

or

```python
users: Iterable[User] = openai.chat.completion.create(
    ...
    stream=True,
    response_model=MultiTask(User)
)

for user in users:
    ...
```

To handle this, we can introduce a check like:

```python
if stream=True and response_model is Iterable, List, MultiTask, or has_attr("from_streaming_response"):
```

I believe there is room for improvement in this developer experience.

Anmol6:
Resolve #213 

## Summary by CodeRabbit

- **New Features**
  - Added a search functionality to the app.


jxnl (commented):
Well, this is amazing. Do you mind adding a little bit to the documentation?

jxnl (commented):
Also does this work for json mode or tools call too?

Anmol6:
> Also does this work for json mode or tools call too?

just merged from main, updating it right now!

examples/streaming_multitask/streaming_multitask.py
```diff
@@ -16,14 +16,14 @@ class User(BaseModel):
     age: int
 
 
-def stream_extract(input: str, cls) -> Iterable[User]:
-    MultiUser = instructor.MultiTask(cls)
-    completion = client.chat.completions.create(
+Users = Iterable[User]
+
+def stream_extract(input: str) -> Iterable[User]:
```
jxnl, line 21:
```suggestion
def stream_extract(input: str) -> Users
```

this would work too right? or even `response_model=Iterable[User]`

love it

jxnl:
at least do something that just says only function allowed or etc.

examples/streaming_multitask/streaming_multitask.py
Anmol6: yup that works

jxnl (approved):
Looks awesome! let me know whenever you'd down to merge.

jxnl:
Looks like coverage fell but thats because the openai tests are not run, no need to mock that

PR #249, opened 2023-12-04
Fixed the system prompt for JSON mode, enabling the use of Pydantic nested models

aastroza:
I realized that in JSON mode, when using nested models, the system prompt was only utilizing the schema of the top-level model and not incorporating the information from the referenced models, leading to hallucinations. This issue can be replicated by running tests [test_classification_enums.py](tests/openai/evals/test_classification_enums.py) and [test_entities.py](tests/openai/evals/test_entities.py) some test cases fail because the JSON response cannot be validated.

With these changes, the tests pass without any issues.

Additionally, I moved a couple of comment lines that were incorrectly positioned.

## Summary by CodeRabbit

- **New Features**
  - Enhanced messaging system to include references to nested models, providing more context in conversations.
  - Introduced a new stopping mechanism for message processing under specific conditions.

- **Tests**
  - Added tests for new nested structures functionality, ensuring reliable behavior across different operational modes.

- **Documentation**
  - Updated to reflect the addition of `Item` and `Order` classes available for use.


jxnl:
This is great thank you. I'll take a look tonight.

jxnl:
If you haven't already, can you add a test Nested structures?

instructor/patch.py
```diff
@@ -92,11 +92,16 @@ def handle_response_model(
         elif mode == Mode.JSON or mode == Mode.MD_JSON:
             if mode == Mode.JSON:
                 new_kwargs["response_format"] = {"type": "json_object"}
-                # check that the first message is a system message
-                # if it is not, add a system message to the beginning
-                message = f"""Make sure that your response to any message matches the json_schema below,
-                        do not deviate at all: \n{response_model.model_json_schema()['properties']}
-                        """
+                # Check for nested models
+                if response_model.model_json_schema()['$defs']:
```
jxnl, line 96: can we do something like 

mesges = f" .. response_model.model_json_schema()"

if "defs" in response_model.model_json_schema():
    messages += "\nhere are some more definitions to adhere too:\n{...}"
    
   I wonder how clean it can be if we just did `response_model.model_json_schema()`

aastroza: Done!

jxnl:
@aastroza i made a change, 

1) it moves the test to evals
2) it uses pytest parameterize

take a look! hopefully you learn something too, let me know if theres anything else you want to do before we merge.

aastroza:
It's much cleaner and understandable now. Thanks for taking the time.

For now, I don't have anything else to add. I'm planning to contribute with new evals soon, which was my initial motivation for solving this issue. I'll do it over the course of the week.

Issue #308, opened 2023-12-30
Move CLI into extras so default dependency chain doesn't include commonmark (which doesn't support Python 3.12)


charles-dyfis-net:
**Is your feature request related to a problem? Please describe.**
Right now, instructor can only be used with Python 3.11 and older releases, because it depends on `rich`, which depends on `commonmark`, which cannot be installed on Python 3.12 (as it uses the `future` library that leverages the deprecated and removed `imp` standard-library module) and is deprecated in favor of `markdown-it-py`.

**Describe the solution you'd like**
If the indirect dependency on commonmark can't be outright removed, it should at least be made optional so users who don't care about CLI functionality can still leverage the library.

**Describe alternatives you've considered**
It's a reasonable alternative -- but with unscoped effort -- to consider updating `rich` to replace the `commonmark` dependency with `markdown-it-py`.

It might be easier to build an updated `future` that no longer attempts to use `imp`, though with commonmark no longer updated, any users intending to work around this bug would need to specify the updated `future` version explicitly.

jxnl:
would be happy to take a PR that updates the toml!

charles-dyfis-net:
One thing I've validated for my own use case, _without_ creating an extra, is adding a new dependency on a patched version of `future` (to make commonmark installable):

```toml
[tool.poetry.group.transitive.dependencies]
future = { git = "https://github.com/odie5533/python-future", rev = "1901c1c347bcad603e8404b64656994eb2cc0439" } # patch to unbreak 3.12 support; dependency of commonmark
```

...but while that works for the parts of instructor I use, I can't say with certainty that it doesn't break something else.

Your call on whether you prefer that or the original extra proposal.

jxnl:
i think extra makes sense 

```
pip install instructor
pip install instructor[cli]
pip install instructor[all]
```

charles-dyfis-net:
Works for me; PR incoming shortly.

PR #309, opened 2023-12-30
Move rich to a new "cli" extras group

charles-dyfis-net:
Per #308.

## Summary by CodeRabbit

- **New Features**
  - Introduced CLI support option in installation instructions.
  - Enhanced CLI with improved error handling and warning for missing dependencies.

- **Documentation**
  - Updated README with additional dependencies information and CLI installation steps.

- **Refactor**
  - Improved CLI initialization process for better user experience.


charles-dyfis-net:
Hmm. It's interesting that the dependency chain here as reflected in `poetry.lock` _doesn't_ include commonmark.

I encountered the dependency when using [`poetry2nix`](https://github.com/nix-community/poetry2nix); inspecting its codebase, it appears to be the culprit, injecting a commonmark dependency into `rich` at https://github.com/nix-community/poetry2nix/blob/528d500ea826383cc126a9be1e633fc92b19ce5d/overrides/default.nix#L2820-L2824

Unless minimizing the mandatory dependency set is seen as valuable on its own, it might make sense to close this PR and the associated ticket and treat this purely as a poetry2nix bug.

Issue #310, opened 2023-12-30
Outdated documentation for llm_validator for openai_client parameter


asimkhan73301:
**Describe the bug**
In #152 an openai_client parameter is introduced in `llm_validator` function and it's default to None where as it's required. If no value is provided it throws an `NoneType` object error at the moment. Most of the llm_validator used in the docs doesn't provide `openai_client` parameter causing errors in the example code.

**To Reproduce**
Issue can be reproduced by following the self critique example code.
```
from openai import OpenAI
import instructor

# Apply the patch to the OpenAI client
# enables response_model keyword
client = instructor.patch(OpenAI())

question = "What is the meaning of life?"
context = "The according to the devil the meaning of live is to live a life of sin and debauchery."

class QuestionAnswerNoEvil(BaseModel):
    question: str
    answer: Annotated[
        str,
        BeforeValidator(
            llm_validator("don't say objectionable things", allow_override=True)
        ),
    ]

try:
    qa: QuestionAnswerNoEvil = client.chat.completions.create(
        model="gpt-3.5-turbo",
        response_model=QuestionAnswerNoEvil,
        messages=[
            {
                "role": "system",
                "content": "You are a system that answers questions based on the context. answer exactly what the question asks using the context.",
            },
            {
                "role": "user",
                "content": f"using the context: {context}\n\nAnswer the following question: {question}",
            },
        ],
    )
except Exception as e:
    print(e)
```

**Expected behavior**
The code is expected to work fine as per the example output in the cookbook. 

Expected output:
```
1 validation error for QuestionAnswerNoEvil
answer
  Assertion failed, The statement promotes sin and debauchery, which is objectionable.
```

The documentation should be updated to incorporate the `openai_client` parameter or we can add a default value for openai_client in the llm_validator function.
 

**Screenshots**
```
     70 def llm(v):
---> 71     resp = openai_client.chat.completions.create(
     72         response_model=Validator,
     73         messages=[
     74             {
     75                 "role": "system",
     76                 "content": "You are a world class validation model. Capable to determine if the following value is valid for the statement, if it is not, explain why and suggest a new value.",
     77             },
     78             {
     79                 "role": "user",
     80                 "content": f"Does `{v}` follow the rules: {statement}",
     81             },
     82         ],
     83         model=model,
     84         temperature=temperature,
     85     )  # type: ignore
     87     # If the response is  not valid, return the reason, this could be used in
     88     # the future to generate a better response, via reasking mechanism.
[...]

PR #311, opened 2023-12-30
Add default openai_client in LLM Validator

asimkhan73301:
As per #310, set default openai_client in case none is provided.

## Summary by CodeRabbit

- **Refactor**
  - Improved the initialization process of user-related functionalities to ensure a seamless experience even with default settings.


jxnl:
adde the patch real quick, but didn't add a test.

asimkhan73301:
My first ever open source contribution lol. Let me go through that one.

asimkhan73301:
@jxnl fixed patch import as it's considered as module import instead of the function because of the same patch.py file. 

Added a test case for default openai_client llm_validator call.

jxnl (approved):

PR #334, opened 2024-01-08
Field level streaming

shanktt:
This is a WIP for adding partial field-level streaming similar to what exists in the [instructor-js](https://github.com/instructor-ai/instructor-js/blob/main/examples/extract_user_stream/index.ts) library. 

Currently `Partial()` fails when encountering fields with generic types like `List[User]`. This means you can do 

```python
class User(BaseModel):
    name: str
    age: str

return client.chat.completions.create(..., response_model=Iterable[Partial(User)], stream=True)
```

But not this at the moment
```python
class Users(BaseModel):
    users: List[User]
    comments: str
    
return client.chat.completions.create(..., response_model=Partial(Users), stream=True)
```

Adding for direct reference in the comment chain
partial_streaming

I could use some help figuring out how to deal with generics when constructing the `Partial` model or if there's an easier way to accomplish this. I've also temporarily hijacked `multitask.tasks_from_chunks` to show how the partial json parsing will look like.

- [x] Support generics in `Partial` / find a better way to take existing model and make all its fields Optional
- [x] Refactor partial json parsing 
- [x] Dedupe partial models to only yield when there are differences


jxnl:
can you add the gif you had before?

shanktt:
Adding for direct reference in the comment chain
partial_streaming

instructor/dsl/partialjson.py
samuelcolvin, line 1: This is going to be extremely slow.

it really wouldn't be that hard to add something to [jiter's python parser](https://github.com/pydantic/jiter/blob/main/src/python.rs) to return the existing python object if you hit an EOF error, and thereby support that functionality.

We could then expose it through pydantic-core.

shanktt:
Streaming in action:
Screen Recording 2024-01-09 at 9 05 43 PM

jxnl:
can you try to fix the test while i change the example

shanktt:
How does the doc look @jxnl? Also still can't make sense of how the tests pass locally but fail here

jxnl:
looks good @shanktt do you think you're able to benchmark the raw stream vs the this partial json thing?

I'm curious what the additional overhead is or if the tokens per second >> json parse.

thisiskeithkwan:
@shanktt  Thanks for the awesome work!
I am trying to implement in into my project of fastapi endpoint + nextjs 14 frontend

Is there an official doc for this change for reference?

shanktt:
@thisiskeithkwan These should work until the pr gets merged (still in-progress so could change): 
[partial_streaming.py](https://github.com/jxnl/instructor/pull/334/files#diff-80b532071fe5fbbe4e5306d217952fb079b2231a664855c83b5b26d610457c97)
[field_streaming.md](https://github.com/jxnl/instructor/pull/334/files#diff-e0f8bd8a50c8677a81235482f996d51632b1fcd5e77a3f887112d86bfae78a02)

shanktt:
@jxnl Let me know if the benchmark looks good to you.

examples/partial_streaming/benchmark.py
```diff
@@ -0,0 +1,151 @@
...
+Raw Streaming Benchmark End
+"""
+
+benchmark_partial_streaming()
+"""
+Partial Streaming Benchmark Start
+users=[PartialUser(name='John Doe', email='johndoe@email.com', twitter='@TechGuru44'), PartialUser(name='Jane Smith', email='janesmith@email.com', twitter='@DigitalDiva88'), PartialUser(name='Alex Johnson', email='alexj@email.com', twitter='@CodeMaster2023')] date='2024-03-15' location='Grand Tech Arena, 4521 Innovation Drive' budget=50000 deadline='2024-02-20'
+Final response received 5.40 seconds after request
+156 total output tokens
+28.89 tok/s
```
jxnl, line 149: oh yeaht his is not great..

instructor/dsl/partialjson.py
jxnl: yeah it looks like its a 10% reduction in tokens per second...
Going to merge that and probs warn in documentation so we can get some testers

shanktt:
Got the benchmark and graph streaming fixed up. Do you still want to mock the streamed response and just benchmark the partialjson function like you suggested on discord?

jxnl:
> Got the benchmark and graph streaming fixed up. Do you still want to mock the streamed response and just benchmark the partialjson function like you suggested on discord?

prosb easier to figure out the actual latency costs! Then we can merge once we add a note thats like "this adds about 5ms per token" (i dont think its that high

alew3:
This looks very cool, any examples on how to hook up a JS to read the stream?

thisiskeithkwan:
> This looks very cool, any examples on how to hook up a JS to read the stream?

Yo, I need this too

jxnl:
@alew3 @thisiskeithkwan any reason you won't want to use the js library in general?

thisiskeithkwan:
Because harder to pair with other reranker API in js ?

Issue #341, opened 2024-01-12
Track Cumulative Token Usage During Any Previous Failed Attempts


lazyhope:
**Is your feature request related to a problem? Please describe.**
The current implementation of `_raw_response.usage` only accounts for the token usage of the latest successful response. It overlooks the tokens consumed during any previous failed attempts. This omission hampers effective cost management, particularly when the `max_retries` limit is reached, leaving us in the dark about the number of tokens expended in unsuccessful attempts.

**Describe the solution you'd like**
To capture comprehensive token usage, including that from previous failed attempts, a cumulative `tokens_used` variable can be incorporated. This variable would aggregate the total token count by accumulating the token usage from each `response`. The modification could be done in the below code snippet as an example:

https://github.com/jxnl/instructor/blob/3f901bc8d3b4af97707341f151b85bfc41f005a5/instructor/patch.py#L284-L298

**Describe alternatives you've considered**
We could also consider modifying the error handling to somehow report token usage when the GPT process fails after hitting the `max_retries` threshold. This approach might change the current error-handling code structure found at:

https://github.com/jxnl/instructor/blob/3f901bc8d3b4af97707341f151b85bfc41f005a5/instructor/patch.py#L322-L324

PR #343, opened 2024-01-13
Introduce `total_usage` variable to track cumulative token usage

lazyhope:
This pull request aims to solve Issue #341, by introducing a new variable named `total_usage` of type `openai.types.completion_usage.CompletionUsage`, which accumulates the token usage from all previous responses, including those that failed validation or JSON decoding, allowing better cost management. 

Currently it does not return the total usage if `max_retries` is reached, since instructor just raise an exception when it happens. It would be worthwhile to consider enhancing the functionality to ensure that `total_usage` is reported even when `max_retries` is exceeded, to provide a complete picture of token usage for monitoring and optimization purposes.

### DIFF
--- a/docs/concepts/raw_response.md
+++ b/docs/concepts/raw_response.md
@@ -25,7 +25,7 @@ print(user._raw_response)
 
 !!! tip "Accessing tokens usage"
 
-    This is the recommended way to access the tokens usage, since it is a pydantic model you can use any of the pydantic model methods on it. For example, you can access the `total_tokens` by doing `user._raw_response.usage.total_tokens`.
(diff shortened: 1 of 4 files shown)


jxnl:
this looks great can you

1) add documentation for usage
2) make sure that serialization works correctly? 

I'll test some stuff tonight!

lazyhope:
> this looks great can you
> 
> 1. add documentation for usage
> 2. make sure that serialization works correctly?
> 
> I'll test some stuff tonight!

Thanks for the feedback! 

I have added a simple documentation and use `ChatCompletion(**model._raw_response.model_dump())` to make sure the serialization works correctly (and I also checked it manually). Please feel free to let me know if there is any other issue!

PR #355, opened 2024-01-22
Update docs to remove references to deprecated apatch.

thedayisntgray:
The method `apatch` is deprecated. Users should instead use `patch`.


jxnl:
nothing here?

Issue #358, opened 2024-01-24
completion_tokens can be None?


indigoviolet:
**Describe the bug**

Frequently seeing errors like this, while using the OpenAI vision model

```python
client_image = instructor.patch(
    OpenAI(api_key=os.getenv("OPENAI_API_KEY")), mode=instructor.Mode.MD_JSON
)

    return client_image.chat.completions.create(
        model="gpt-4-vision-preview",
        response_model=response_model,
        max_tokens=2048,
        temperature=0,
        messages=[
            {
                "role": "user",
                "content": [
                    {
                        "type": "text",
                        "text": prompt,
                    },
                    {
                        "type": "image_url",
                        "image_url": {
                            "url": f"data:image/jpeg;base64,{base64_encode(img)}"
                        },
                    },
                ],
            }
        ],
    )
```

```
File [~/dev/plus/python/.venv/lib/python3.11/site-packages/instructor/patch.py:303), in retry_sync(func, response_model, validation_context, args, kwargs, max_retries, strict, mode)
    [301](~/dev/plus/python/.venv/lib/python3.11/site-packages/instructor/patch.py:301) stream = kwargs.get("stream", False)
    [302](~/dev/plus/python/.venv/lib/python3.11/site-packages/instructor/patch.py:302) if isinstance(response, ChatCompletion) and response.usage is not None:
--> [303](~/dev/plus/python/.venv/lib/python3.11/site-packages/instructor/patch.py:303)     total_usage.completion_tokens += response.usage.completion_tokens
    [304](~/dev/plus/python/.venv/lib/python3.11/site-packages/instructor/patch.py:304)     total_usage.prompt_tokens += response.usage.prompt_tokens
    [305](~/dev/plus/python/.venv/lib/python3.11/site-packages/instructor/patch.py:305)     total_usage.total_tokens += response.usage.total_tokens

TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```

**To Reproduce**
It doesn't happen every time, so I'm not absolutely certain how to reproduce it. 

instructor 0.4.8
openai 1.9.0

**Expected behavior**

**Screenshots**

jxnl:
would be happy to review a PR that checks nullity before adding!

PR #359, opened 2024-01-24
Be robust to null values while updating token counts

indigoviolet:
See #358


indigoviolet:
@jxnl You've enabled auto-merge, but I believe someone also needs to approve

PR #362, opened 2024-01-25
feat:  Add support for partial literals and enums

lucasastorian:
----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :warning: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 23a3abb294f214aeae950f34b1149f2877ebafd4.  | 
|--------|--------|

### Summary:
This PR modifies the '_make_field_optional' function in 'partial.py' to convert Literal and Enum fields to Optional[str] to prevent streaming failures with OpenAI.

**Key points**:
- Added checks in '_make_field_optional' function to convert Literal and Enum fields to Optional[str].
- This change is made to prevent failures when streaming with OpenAI.
----


jxnl:
beautiful! Can you:

- [ ] add a test
- [ ] making a note of this in the docs?

jxnl:
Looking back, I feel like things should be... validated with and without strict mode. Something we should think harder about.

lucasastorian:
Sorry - haven't gotten to this yet. Will take a look tomorrow. 

I've been using Partial to stream incomplete tool completions from OpenAI (I added support for tool choice myself). The issue was that (obviously), the Pydantic schema would fail mid-stream, when OpenAI didn't stream the full Literal/Enum value at once. One way to do it, is to validate completed attributes in strict mode, and incomplete attributes ones in unstrict mode.  That might be a little overkill though - it's just as easy to validate in unstrict mode during the stream & then in strict mode once complete.

jxnl:
Yeah, let's just do unstrict, doing partial streaming in strict.

jxnl:
are you able to add tests?

PR #421, opened 2024-02-09
[DO NOT MERGE] remove generics and introduce types

savarin:


savarin:
@jxnl I had trouble adding types to `patch.py` in its current state. I created this draft PR to illustrate how we're able to add types to `patch.py` but without using generics. Would you be open to removing the generics to get `patch` to pass mypy?

jxnl:
thats fine since they are not working as expected anyways we'll have the new client in 1.0 but then we'll need the generics

PR #423, opened 2024-02-09
fix(parallel): enhance error handling in get_types_array and add test case

jxnl:
| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit d61dcaf033f1244a7452788f7a7fefb8da3fdf37.  | 
|--------|--------|

### Summary:
This PR enhances error handling in `get_types_array` function and adds a corresponding test case.

**Key points**:
- Updated `get_types_array` function in `instructor/dsl/parallel.py` to raise `TypeError` for non-iterable models.
- Added new test case `test_sync_parallel_tools__error` in `tests/openai/test_parallel.py`.
----

### DIFF
--- a/tests/openai/test_parallel.py
+++ b/tests/openai/test_parallel.py
@@ -14,6 +14,23 @@ class GoogleSearch(BaseModel):
     query: str
 
 
+def test_sync_parallel_tools__error(client):
+    client = instructor.patch(client, mode=instructor.Mode.PARALLEL_TOOLS)
+
+    with pytest.raises(TypeError):
+        resp = client.chat.completions.create(
+            model="gpt-4-turbo-preview",
+            messages=[
+                {"role": "system", "content": "You must always use tools"},
+                {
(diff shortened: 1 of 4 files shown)


tests/openai/test_parallel.py
```diff
@@ -14,6 +14,23 @@ class GoogleSearch(BaseModel):
     query: str
 
 
+def test_sync_parallel_tools__error(client):
+    client = instructor.patch(client, mode=instructor.Mode.PARALLEL_TOOLS)
+
+    with pytest.raises(TypeError):
```
jxnl, line 20: @lukevs We should definitely be switching back to regular tools mode if we can detect that it's not an iterable.

lukevs: Need it still for the `Boy | Girl` case too right (to get the name disambiguation)

PR #428, opened 2024-02-11
docs: Sync requirements-doc.txt with Poetry dev group till mkdocs works

ryanhalliday:
I tried to get setup with the docs using Poetry but it didn't work with just `poetry install` and I had to use `pip install -r requirements-docs.txt` in a different venv. 

So I did `poetry add --group dev PACKAGE` till it worked.

Added were:
- mkdocs-jupyter
- mkdocs-rss-plugin
- mkdocs-minify-plugin
- cairosvg
- pillow

I installed each of `pillow` and `cairosvg` to match `requirements-doc.txt` instead of `mkdocs-material[imaging]` but installing that imaging package might be better?

I know `mkdocs-jupyter` is a recent addition but maybe seeing as it includes a bunch of stuff a poetry `docs` group could be good? 

**Hope this was the right thing to do?**
Python packaging always feels kinda wild


jxnl:
Making a docs group would be awesome. And using imagery sounds like a good plan will merge when ready. 

Will need to change contributing.md and some deployment workflows.

jxnl (approved):

jxnl:
lets add a docs group and make sure test-docs and mkdocs works

ryanhalliday:
Made the imaging change and fixed the `--with` param in workflows. 

I Test Docs is failing because of OpenAI secret in PR workflow?

I didn't add anything to `contributing.md` - alright if I leave that for you?

PR #435, opened 2024-02-13
fix: json modes - don't add json schema again in system message

teome:
Using any of the JSON modes, the final action in `handle_response_model` adds the schema to the system message.

This PR fixes fixes a missing `else` in the logic for checking the existence of the system message. If there's no system message the code currently adds the schema twice: first by creating a system message, then checking if a system message exists and if so appending the schema.

Just needs the `else`

### DIFF
--- a/instructor/patch.py
+++ b/instructor/patch.py
@@ -159,9 +159,8 @@ def handle_response_model(
                         "content": message,
                     },
                 )
-
-            # if the first message is a system append the schema to the end
-            if new_kwargs["messages"][0]["role"] == "system":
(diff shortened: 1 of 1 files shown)


jxnl:
Will this work? I'm pretty sure the message should be part of the system message.

teome:
I believe so, unless I'm going crazy, because if `new_kwargs["messages"][0]["role"] != "system` then you were already adding the `message` with schema. If we go to the else, it must be system already and it should be added. Before, if there was no system message, it was added, then appended to, resulting in two copies of the json schema message

PR #422, opened 2024-02-09
chore: Include types to instructor.patch

savarin:
This PR adds types to the following files, as per https://github.com/jxnl/instructor/issues/390.

```
instructor/patch.py
tests/test_patch.py
```

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 26a57a899045af56ef008c46c87b8b2e8aa1dd37.  | 
|--------|--------|

### Summary:
This PR adds type annotations to `instructor/patch.py` and `tests/test_patch.py`, and includes these files in the MyPy check in `.github/workflows/mypy.yml`.

**Key points**:
- Added type annotations to `instructor/patch.py` and `tests/test_patch.py`.
- Updated `.github/workflows/mypy.yml` to include these files in MyPy check.
- Made minor code adjustments in `instructor/patch.py` for type correctness.
----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)

PR #437, opened 2024-02-15
Revert "chore: Include types to instructor.patch"

jxnl:
Reverts jxnl/instructor#422


jxnl:
@savarin the openai api call tests are all failing due to missing optionals.

savarin:
> @savarin the openai api call tests are all failing due to missing optionals.

Got it, will take a look.

PR #440, opened 2024-02-17
chore: include types to patch.py and test

savarin:
| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 71d652700888b26a9c59d7979f46652dd7fc7062.  | 
|--------|--------|

### Summary:
This PR adds type hints to `patch.py` and its corresponding test file `test_patch.py`, and updates the `mypy.yml` workflow to include these files for type checking.

**Key points**:
- Added type hints to the functions and methods in `patch.py`.
- Updated `mypy.yml` to include `patch.py` and `test_patch.py` for type checking.
- Updated `test_patch.py` to reflect the changes in `patch.py`.

----

### DIFF
--- a/instructor/patch.py
+++ b/instructor/patch.py
@@ -172,7 +172,7 @@ def process_response(
     stream: bool,
     validation_context: Optional[Dict[str, Any]] = None,
     strict: Optional[bool] = None,
-    mode: Mode = Mode.FUNCTIONS,
+    mode: Mode = Mode.TOOLS,
 ) -> Union[OpenAISchema, List[OpenAISchema]]:
     """Processes a OpenAI response with the response model, if available.
 
@@ -182,7 +182,7 @@ def process_response(
         stream (bool): Whether the response is a stream
         validation_context (dict, optional): The validation context to use for validating the response. Defaults to None.
         strict (_type_, optional): Whether to use strict json parsing. Defaults to None.
-        mode (Mode, optional): The openai completion mode. Defaults to Mode.FUNCTIONS.
+        mode (Mode, optional): The openai completion mode. Defaults to Mode.TOOLS.
 
     Returns:
         Union[T_Model, T]: The parsed response, if a response model is available, otherwise the response as is from the SDK
@@ -216,7 +216,6 @@ def process_response(
     if isinstance(response_model, ParallelBase):
         return model
 
-    assert hasattr(model, "_raw_response")
     model._raw_response = response
     return model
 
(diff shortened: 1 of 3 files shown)


savarin:
> @savarin the openai api call tests are all failing due to missing optionals.

@jxnl Trying this again. To clarify, are these the missing optionals? If yes, can I confirm it the tests now succeed? Thanks!
https://github.com/jxnl/instructor/pull/422/files#diff-1ccc47ac75961e3a4f37760e35da89919e2497efffdb67f3e5b6675a419c44b5L539-L540

instructor/patch.py
```diff
@@ -168,14 +166,14 @@ def handle_response_model(
...
-    response_model: Type[T_Model],
+    response_model: Union[Type[OpenAISchema], ParallelBase, None],
     stream: bool,
-    validation_context: dict = None,
-    strict=None,
-    mode: Mode = Mode.TOOLS,
-) -> Union[T_Model, T]:
+    validation_context: Optional[Dict[str, Any]] = None,
+    strict: Optional[bool] = None,
+    mode: Mode = Mode.FUNCTIONS,
```
jxnl, line 175: can you change this to tools for all the new calls.

instructor/patch.py
```diff
@@ -213,24 +211,27 @@ def process_response(
     # ? This really hints at the fact that we need a better way of
     # ? attaching usage data and the raw response to the model we return.
     if isinstance(model, IterableBase):
-        return [task for task in model.tasks]
+        return [task for task in model.tasks]  # type: ignore[attr-defined]
 
     if isinstance(response_model, ParallelBase):
         return model
 
+    assert hasattr(model, "_raw_response")
```
jxnl, line 219: why check it has atter before you assign it,

savarin: Actually it turns out this is not needed, since removed.

We need it in line 276, otherwise get the following error because `response_model.from_response` could also return a generator.

```shell
instructor/patch.py:277: error: Item "dict[str, Any]" of "Union[dict[str, Any], Generator[Any, None, None]]" has no attribute "_raw_response"  [union-attr]
instructor/patch.py:277: error: Item "Generator[Any, None, None]" of "Union[dict[str, Any], Generator[Any, None, None]]" has no attribute "_raw_response"  [union-attr]
```

savarin:
@jxnl Let me know if these changes are helpful. I can rebase with latest main, but also OK to close if not needed. Thanks!

jxnl:
waiting on the new client code

savarin:
> waiting on the new client code

Sounds good, will do the same.

PR #466, opened 2024-02-24
Adding new article on how to write better code

ivanleomk:
Adding a new article which talks a bit about my own learnings on how to write better python code.

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 7ccb8a15ab357a332732f801142985e14599d160.  | 
|--------|--------|

### Summary:
This PR adds a new article that provides insights and tips on writing better Python code, especially for machine learning projects.

**Key points**:
- Added a new article `write-code-to-refactor.md` in the `/docs/blog/posts/` directory.
- The article discusses common mistakes in Python code and provides tips for improvement.
- Topics covered include using abstractions, standard libraries, managing complexity, writing simple functions, and more.
- The article includes code examples for better understanding.

----

### DIFF
--- a/docs/blog/posts/write-code-to-refactor.md
+++ b/docs/blog/posts/write-code-to-refactor.md
@@ -9,7 +9,7 @@ authors:
   - jxnl
 ---
 
-# Stop shooting yourself in the foot with python
+# Write Code To Refactor
 
 In this article, we'll walk through some common mistakes people tend to make when writing python code. Then, we'll look at how these considerations should be used when working on larger machine learning training jobs and projects. Lastly, we'll conclude with some suggestions that you'll be able to apply to your project today.
 
@@ -180,10 +180,11 @@ def process():
 
 There are two main problems here
 
-- We're introducing a strange side effect that `format_data` doesn't reflect
-- We're mutating state, which makes it difficult to debug issues when our logic gets more complicated
+1. We're introducing a strange side effect that `format_data` doesn't reflect
 
-A better way to write this logic would include splitting up the saving of data and the formatting of data into two separate functions. That way our functions closely reflect the actual actions that are being performed. We'd also try to minimize the mutation of state, as seen below.
+2. We're mutating state, which makes it difficult to debug issues when our logic gets more complicated
+
+A better way to write this logic would include splitting up the saving of data and the formatting of data into two separate functions. That way our functions closely reflect the actual actions that are being performed. We'd also try to minimize the mutation of state so that it's easier to trace errors when they happen
 
 ```python
 def format_data(data):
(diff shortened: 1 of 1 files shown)


docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+```python
+def format_data(data):
+  headers = extract_headers(data)
+	data = format_dataframe(data,headers)
+	upload_to_database(data)
+  return data
+
+def process():
+	data = get_data()
+	data = format_data(data)
```
shanktt, line 178: Some inconsistent indentation

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+  headers = extract_headers(data)
+	return format_dataframe(data,headers)
+
+def save_data(data):
+  upload_to_database(data)
+
+def process():
+	data = get_data()
+	formatted_data = format_data(data)
+  save_to_database(data)
```
shanktt, line 200: Some more inconsistent indentation

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+  - Clean Code
+  - Code Quality
+authors:
+  - ivanleomk
+  - jxnl
+---
+
+# Write Code To Refactor
+
+In this article, we'll walk through some common mistakes people tend to make when writing python code. Then, we'll look at how these considerations should be used when working on larger machine learning training jobs and projects. Lastly, we'll conclude with some suggestions that you'll be able to apply to your project today.
```
jxnl, line 14: would love to see this from your perspective

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+
+This class is **very prone to errors** because we're relying on a single string to determine the logic. Imagine how much more complex we'd need to make the control flow when we add more options and complex configurations. 
+
+In order to prevent these complex errors from happening, we can use a `@classmethod` and a `enum` in order to prevent this from happening. A  `classmethod` allows us to instantiate a class with the right configuration so that we don't run into these complex edge cases. Furthermore, by pre-defining the values that a user can pass in, we can also support more complex initialisation logic.
+
+```python
+import enum
+
+class Provider(enum.Enum):
+    HUGGINGFACE = "HuggingFace"
```
jxnl, line 134: don't capitalize

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+        )
+```
+
+Small things like class methods and enums go a long way in ensuring that code that we write is safe and works as intended.
+
+### Write Simple Functions
+
+The functions I write have ideally three characteristics
+
+- They do one thing and their name reflects it
```
jxnl, line 161: function names are verbs or actions, get_x, compute_y
and types are nowns "results"

vs `def results(user_id)`

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+- They are used at the same level of abstraction
+- They are parameterized with nice type hints 
+
+#### Do one thing well 
+
+A lot of times, it's very tempting to write a function that does a lot in one step. In the example above, we can see that we're trying to both format a specific chunk of data and then save it into the database.
+
+```python
+def format_data(data):
+  headers = extract_headers(data)
```
jxnl, line 171: It helps to have one main function while a sequence of other functions and have those functions be as flat as possible.


```python
def main():
   x = F(x)
   x = G(x)
   return H(x)
```
Where f, g, and h are pretty simple and flat functions with no additional depth. 

```python
def F(x):
    x = A(x)
    x = B(x)
  
def main():
   x = F(x)
   x = G(x)
   return H(x)
```

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+def process():
+	data = get_data()
+	data = format_data(data)
+```
+
+There are two main problems here
+
+1. We're introducing a strange side effect that `format_data` doesn't reflect
+
+2. We're mutating state, which makes it difficult to debug issues when our logic gets more complicated
```
jxnl, line 185: You should use an example when you're just like mutating a dictionary behind the scenes. 

state = {}

x = f(x, state)
y = g(x, state)

return x, y, state


instead it should be 



state = {}

x, state = f(x, state)
y, state = g(x, state)

return x, y, state

This is going to be more testable because you can actually just pass in whatever state you want and check the result state. Rather than assuming mutations will happen. You should only mutate state if that state copy/mutation is expensive.

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+def save_data(data):
+  upload_to_database(data)
+
+def process():
+	data = get_data()
+	formatted_data = format_data(data)
+  save_to_database(data)
+```
+
+#### Be consistent with abstraction
```
jxnl, line 203: I don't think this is a good example, but we can really talk about it in terms of keeping the functions clear and simple.

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+    raise ValueError(f"Invalid model name of {model_name} was provided")
+
+def inference(model_name:str,data:dict):
+  model = initialise_model(model_name)
+	embeddings = model.infer(data)
+```
+
+Alternatively, we could take advantage of the `@classmethod` decorator again and define a method on the class which takes in a model_name and sets the relevant parameters.
+
+## Training Models
```
jxnl, line 256: This should almost be a second post. 

Basically saying that we should just get things wired up first and then increase data size.

docs/blog/posts/write-code-to-refactor.md
```diff
@@ -0,0 +1,402 @@
...
+    list_implementation, generator_implementation = benchmark(sz)
+    print(
+        f"Size: {sz:<15,.0f} - Generator: {generator_implementation:10,.10f}, List:{list_implementation:10,.10f}. Generator is {round(list_implementation/generator_implementation,3)} times faster"
+    )
+```
+
+We can see the output of the script above in the code block below.
+
+```bash
+Size: 100             - Generator: 0.0000052080, List:0.0000053750. Generator is 1.032 times faster
```
jxnl, line 330: It's not about being faster per se, it's about the simplicity of the code. 

And you shouldn't be writing your own benchmark code, you should use something like` timeit` You guys are not capturing any variants.

PR #486, opened 2024-03-05
feat: support anthropic function calling

shreya-51:
todo:
 - support async streaming
 - add async streaming tests


instructor/patch.py
```diff
@@ -474,6 +494,47 @@ def is_async(func: Callable) -> bool:
...
+        type_element.text = field_type.__name__
+        
+        param_description = ET.SubElement(parameter, "description")
+        param_description.text = f"The {field_name} of the {model.__name__} model"
+    
+    return ET.tostring(root, encoding="unicode")
+
+# todo: make function better (edge cases, robustness, etc.)
+# todo: super primitive parsing, make better
+def extract_xml(content: str) -> str:
```
jxnl, line 523: lets mode this to an `anthropic_utils.py`

instructor/patch.py
```diff
@@ -474,6 +494,47 @@ def is_async(func: Callable) -> bool:
         is_coroutine = is_coroutine or inspect.iscoroutinefunction(func)
     return is_coroutine
 
+# todo: make function better (edge cases, robustness, etc.)
+def model_to_xml(model: BaseModel) -> str:
```
jxnl, line 498: does this handle nested objects?

instructor/patch.py
```diff
@@ -26,6 +27,7 @@
     ChatCompletionMessageParam,
 )
 from openai.types.completion_usage import CompletionUsage
+from anthropic.types.message import Message
```
jxnl, line 30: avoid the import  via importlib

and we can make a group for anthropic in the toml

jxnl:
take a look at https://github.com/jxnl/instructor/pull/485/files

instructor/patch.py
```diff
@@ -173,11 +175,28 @@ def handle_response_model(
             # if it is, system append the schema to the end
             else:
                 new_kwargs["messages"][0]["content"] += f"\n\n{message}"
+        elif mode == Mode.ANTHROPIC_TOOLS:            
+            tool_descriptions = json_to_xml(response_model.schema_json())
+            system_prompt = f"""In this environment you have access to a set of tools you can use to answer the user's question.
```
jxnl, line 180: you can't assume a system prompt does not already exist

jxnl: but also is the system prompt also actually 

create(system=str)

shreya-51: > but also is the system prompt also actually
> 
> create(system=str)

yep, https://docs.anthropic.com/claude/docs/system-prompts#:~:text=system%3D%22Respond%20only%20in%20Spanish.%22%2C%20%23%20%3C%2D%2D%20system%20prompt

instructor/anthropic_utils.py
```diff
@@ -0,0 +1,65 @@
+import re
+import json
+import xmltodict
+from pydantic import BaseModel
+import xml.etree.ElementTree as ET
+from typing import Type, Any, Dict, TypeVar
+
+T = TypeVar('T', bound=BaseModel)
+
+def json_to_xml(json_str: str) -> str:
```
jxnl, line 10: any reason its json_str and not `BaseModel`

jxnl: cause it looks like we convert it to str just to load it again

instructor/process_response.py
```diff
@@ -281,6 +281,26 @@ def handle_response_model(
             # if it is, system append the schema to the end
             else:
                 new_kwargs["messages"][0]["content"] += f"\n\n{message}"
+        elif mode == Mode.ANTHROPIC_TOOLS:
+            tool_descriptions = response_model.anthropic_schema
+            system_prompt = f"""In this environment you have access to a set of tools you can use to answer the user's question.
```
jxnl, line 286: should use dedent here

pyproject.toml
```diff
@@ -18,6 +18,10 @@ rich = "^13.7.0"
 aiohttp = "^3.9.1"
 tenacity = "^8.2.3"
 
+[tool.poetry.group.anthropic.dependencies]
```
jxnl, line 21: should update the readme

instructor/anthropic_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    properties = model_dict.get("properties", {})
+            
+    for field_name, details in properties.items():
+        parameter = ET.SubElement(root, "parameter")
+        name = ET.SubElement(parameter, "name")
+        name.text = field_name
+        type_element = ET.SubElement(parameter, "type")
+        type_element.text = details.get("type", "unknown") # Might be better to fail here if there is no type since pydantic models require types
+        param_description = ET.SubElement(parameter, "description")
+        param_description.text = f"The {field_name} of the {model_dict['title']} model"
```
jxnl, line 42: shuoldn't this be part of the model, not the generated str we produce for them?

instructor/anthropic_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    pattern = r'<function_calls>.*?</function_calls>'
+    matches = re.findall(pattern, content, re.DOTALL)
+    return "".join(matches)
+
+# todo: make function better (edge cases, robustness, etc.)
+def xml_to_model(model: Type[T], xml_string: str) -> T:
+    """Converts XML in Anthropic's schema to an instance of the provided class."""
+    parsed_xml = xmltodict.parse(xml_string)
+    model_dict = parsed_xml['function_calls']['invoke']['parameters']
+    return model(**model_dict)
```
jxnl, line 65: this assumes single function call, is that expected?

jxnl:
@shreya-51 i added a test that fails

shreya-51:
> @shreya-51 i added a test that fails

yeah I noticed that this often fails with Lists because its not defined how the LLM should structure them - will add a fix

instructor/anthropic_utils.py
jxnl: another way to explain this is that theres literally an attribute called description we can use rather than making up `f"The {field_name} of the {model_dict['title']} model"`

jxnl:
https://github.com/jxnl/instructor/issues/529

PR #488, opened 2024-03-05
[chore] add langchain integration example

synacktraa:


jxnl:
langchain has their own structured output programs

synacktraa:
> langchain has their own structured output programs 

Yes, I'm aware. I was playing around with instructor and langchain library so I created a PR.

whitmo:
> langchain has their own structured output programs

IMO Langchain has lots of stuff, but it's business plan dictates it is an ecosystem enabler vs a singular solution.  

Speaking for myself, I would benefit from seeing examples of how to drop in or integrate instructor w/ systems I'm already developing.

PR #521, opened 2024-03-21
fix: add handling for List[non-object] types

shreya-51:
- adds handling for List[non-object] types in the model
 - adds test for List[non-object] type case
 - note: haiku tends to hallucinate the List[non-object] format, opus works better

### DIFF
--- a/tests/anthropic/test_simple.py
+++ b/tests/anthropic/test_simple.py
@@ -64,6 +64,31 @@ def test_nested_type():
     assert resp.address.street_name == "First Avenue"
 
 @pytest.mark.skip
+def test_list():
+    class User(BaseModel):
+        name: str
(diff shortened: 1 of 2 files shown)


tests/anthropic/test_simple.py
```diff
@@ -63,6 +63,31 @@ class User(BaseModel):
...
 
+@pytest.mark.skip
+def test_list():
+    class User(BaseModel):
+        name: str
+        age: int
+        family: List[str]
+    
+    resp = create(
+        model="claude-3-opus-20240229", # Fails with claude-3-haiku-20240307
```
jxnl, line 74: interesting, we should mention this ins some kind of "anthropic tips" part

PR #527, opened 2024-03-22
Better partial support

mwildehahn:
I noticed that when using `Partial[MyModel]` the model wasn't always populating required fields. With this change, the JSON schema we send to the model doesn't change (aside from having the model names updated to `Partial{...}` to make it match validation). 

The naming of all the `Partial` / `_Partial ` classes is a little weird. I was trying to keep this backwards compatible with the existing implementation. Open to suggestions on how to make this better.

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 610410797a1f8a8965d5c734dce96f267b17a3c6.  | 
|--------|--------|

### Summary:
This PR improves the handling of partial models in the `instructor` package, adds tests for the updated functionality, and updates some dependencies to optional in `poetry.lock`.

**Key points**:
- Updated `Partial` class in `instructor/dsl/partial.py` to better handle partial models.
- Added a new test file `tests/dsl/test_partial.py` to test the updated `Partial` class.
- Updated some dependencies to optional in `poetry.lock`.

----

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -176,6 +176,7 @@ class PartialBase(Generic[T_Model]):
     ) -> Generator[T_Model, None, None]:
         prev_obj = None
         potential_object = ""
+        partial_model = cls.get_partial_model()
         for chunk in json_chunks:
             potential_object += chunk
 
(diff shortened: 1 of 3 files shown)


jxnl:
this is great yeah the _Partial might need some work but this looks correct.
don't have time to give a deep review but why does it need both? 

if we break backwards compat, can we keep only one PArtial

mwildehahn:
I was able to make this cleaner with the use of another class `MakeFieldsOptional`. Users of the lib can just do `Partial[Model]` but internally when we actually want to convert nested fields to optional we can do `Partial[Model, MakeFieldsOptional]`.

shanktt (commented):
Changes look good to me! Nice approach with the `MakeFieldsOptional`, that's really clean

jxnl (approved):

PR #535, opened 2024-03-27
Clean up Anthropic tests

Cruppelt:
# Anthropic Test Cleanup

## Fix the install dependencies for anthropic/dev 
```bash
poetry install --with dev anthropic
```
to
```bash
poetry install --with dev,anthropic
```

Pipeline now actually works.

## Unit tests are now mocked for anthropic

No need to call the API, we want to test everything before and after the actual call. Integration tests are better for actual API results and we want to ensure temperate=0 so results are less variable. 

## Anthropic Unit tests don't fail
The token was never set via a secret so they'd always fail. I have now mocked the responses with expected responses so the API doesn't need to be called. 

## Poetry Lock out of date

Updated dependencies from the pyproject.toml to poetry.lock. Fixes pipeline.

## Separate out anthropic tests

So we know which tests are failing more precisely. Anthropic logic seems to need some work so good to separate out for now.

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit cd3a432d549744287fef6c0e36e05fd525e03430.  | 
|--------|--------|

### Summary:
This PR updates dependencies in the `poetry.lock` file and modifies the `anthropic` package installation in the `test.yml` file.

**Key points**:
- Updated the `poetry.lock` file with newer versions of dependencies.
- Changed the `anthropic` package installation in the `test.yml` file.
- No new `md` file in `docs`.
- No changes to the library code, so no need for updated tests or documentation.

----


tests/anthropic/test_anthropic.py
```diff
@@ -1,45 +1,62 @@
...
+    )
 
 
-@pytest.mark.skip
 def test_anthropic():
-    class Properties(BaseModel):
+    class Property(BaseModel):
         name: str
-        value: List[str]
+        value: str
```
Cruppelt, line 29: Removing complexity here, with current implementation of XML generation and non-0 temperature this is hit or miss to get correct. Will be fixed later.

instructor/retry.py
```diff
@@ -83,14 +83,15 @@ def retry_sync(
...
-                    return process_response(
+                    result = process_response(
                         response,
                         response_model=response_model,
                         stream=stream,
                         validation_context=validation_context,
                         strict=strict,
                         mode=mode,
                     )
+                    return result
```
Cruppelt, line 94: extra variable setting for easier debugging

instructor/function_calls.py
```diff
@@ -63,11 +65,7 @@ def openai_schema(cls) -> Dict[str, Any]:
     @property
     def anthropic_schema(cls) -> str:
         from instructor.anthropic_utils import json_to_xml
-
-        return "\n".join(
-            line.lstrip()
-            for line in parseString(json_to_xml(cls)).toprettyxml().splitlines()[1:]
-        )
+        return json_to_xml(cls)
```
Cruppelt, line 68: Breaks with List[x] as it inserts non-XML into the result. I'll add back the prompt clean up but this fixes current critical issue in main branch.

.github/workflows/test.yml
```diff
@@ -33,17 +33,20 @@ jobs:
         uses: snok/install-poetry@v1.3.1
 
       - name: Install dependencies
-        run: poetry install --with dev anthropic
+        run: poetry install --with dev,anthropic
```
Cruppelt, line 36: This actually works

jxnl:
I added environment variables for anthropic. We should be able to run these tests now.

Cruppelt:
> I added environment variables for anthropic. We should be able to run these tests now.

You added as secret? Doesn't seem to be picked up here https://github.com/jxnl/instructor/pull/535/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R40

PR #541, opened 2024-03-28
Preserve async function type when patching async functions

noxan:
Currently if you `patch` an async function (e.g. from `AsyncOpenAI`), the function is returned as **NOT** async. This pull requests preserves the async nature of the function (meaning fixes type checking).

Example
```python
import instructor
from openai import AsyncOpenAI

client = AsyncOpenAI()
acreate = instructor.patch(create=client.chat.completions.create)

async def main():
    response = await acreate(...)
```

Maybe there is a more clever way to do this, I don't know how.

Motivation: it's quite annoying if your type checker misleads you into removing an `await` statement which is really needed - took me way too long to the issue.

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 60740f1a30a8288ab20c204cfd0612a9435469e3.  | 
|--------|--------|

### Summary:
This PR modifies the `patch` function in `instructor/patch.py` to preserve the async nature of a function when it is patched, by introducing a new protocol and a new overload for the function.

**Key points**:
- Introduced a new protocol `AsyncInstructorChatCompletionCreate` in `instructor/patch.py`.
- Added a new overload for the `patch` function that returns `AsyncInstructorChatCompletionCreate` when an async function is patched.
- The changes preserve the async nature of a function when it is patched.

----

### DIFF
--- a/instructor/patch.py
+++ b/instructor/patch.py
@@ -38,6 +38,17 @@ class InstructorChatCompletionCreate(Protocol):
     ) -> T_Model: ...
 
 
+class AsyncInstructorChatCompletionCreate(Protocol):
+    async def __call__(
+        self,
+        response_model: Type[T_Model] = None,
+        validation_context: dict = None,
+        max_retries: int = 1,
+        *args: T_ParamSpec.args,
+        **kwargs: T_ParamSpec.kwargs,
+    ) -> T_Model: ...
+
+
 @overload
 def patch(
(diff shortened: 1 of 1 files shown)


jxnl:
take a look at 1.0!

PR #565, opened 2024-04-04
Allow extras dependencies to truly be optional

ameade:
Installing via poetry install with no extras resulted in:

> Traceback (most recent call last):
>   File "summarizer.py", line 9, in <module>
>     import instructor
>   File "instructor/instructor/__init__.py", line 56, in <module>
>     from .client_groq import from_groq
>   File "instructor/instructor/client_groq.py", line 9, in <module>
>     client: groq.Groq,
>             ^^^^^^^^^
> AttributeError: module 'groq' has no attribute 'Groq'

This is because groq 0.1.1 (which I had installed) does not have groq.Groq. So it needs to be pinned to ^0.4.2. However, I am not using groq but the 'test-docs' extra installs both groq and anthropic, so it should probably be an extra itself. groq and anthropic being installed with 'test-docs' causes them to be imported in __init__.py

> if importlib.util.find_spec("groq") is not None:
> from .client_groq import from_groq
>     __all__ += ["from_groq"]

Furthermore, all of the dependencies for these extras need to be marked optional or else they are installed anyways.

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 52f95bff0c1d53cd9676edd9db6afb9dea23db84.  | 
|--------|--------|

### Summary:
This PR makes 'groq' and 'anthropic' optional dependencies, removes unnecessary dependencies, and updates the 'groq' version.

**Key points**:
- Made 'groq' and 'anthropic' optional dependencies.
- Removed unnecessary dependencies from 'poetry.lock'.
- Updated 'groq' version to 0.4.2 in 'pyproject.toml'.

----

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -125,7 +125,7 @@ files = [
 name = "anthropic"
 version = "0.21.3"
 description = "The official Python library for the anthropic API"
-optional = false
+optional = true
 python-versions = ">=3.7"
 files = [
     {file = "anthropic-0.21.3-py3-none-any.whl", hash = "sha256:5869115453b543a46ded6515c9f29b8d610b6e94bbba3230ad80ac947d2b0862"},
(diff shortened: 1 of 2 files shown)


jxnl:
thanks so much!

PR #579, opened 2024-04-08
Refactor anthropic message format to support tool messages & Allows Anthropic Bedrock clients

lazyhope:
Instead of concatenating strings with "\n\n", we use the following format to merge consecutive Anthropic messages of the same roles:
```json
{
  "role": "user", 
  "content": [
    {"type": "text", "text": "Hello, Claude"},
    {"type": "text", "text": "How is it going?"}
  ]
}
```
Which also allows other message types like image objects, tool invocation (for assistant), tool results, etc. As a comparsion, this version of tool support is able to pass the anthropic tests which currently fails.

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 9705a6f3b4cf00ddfb04c9d999f53f39c44312a2.  | 
|--------|--------|

### Summary:
This PR refactors the anthropic message format to support tool messages, updates the `reask_messages` and `merge_consecutive_messages` functions, and modifies the corresponding tests.

**Key points**:
- Refactored `reask_messages` function in `/instructor/retry.py` to include the original response and format error messages differently.
- Updated `merge_consecutive_messages` function in `/instructor/utils.py` to handle the new message format.
- Updated tests in `/tests/test_utils.py` to reflect these changes.

----


tests/test_new_client.py
```diff
@@ -176,3 +177,68 @@ def test_client_anthropic_response():
     )
     assert user.name == "Jason"
     assert user.age == 10
+
+
+def test_client_anthropic_bedrock_response():
```
jxnl, line 182: can we mark this test as skipped since i dont have permissions

tests/test_utils.py
```diff
@@ -137,8 +137,20 @@ def test_merge_consecutive_messages():
     ]
     result = merge_consecutive_messages(messages)
```
jxnl, line 138: maybe have a provider keyword so it doesn't mess up openai stuff

jxnl (changes requested):
1. skip anthropic bedrock tests
2. add some logic to avoid doing the merge for openai keys

tests/test_utils.py
palako: I reckon the previous implementation above (with the \n instead of the extend method for merging messages) was better and didn't break openAI things. I'd be in favour of not needing to pass provider keywords if possible, the more generic the code the better in my opinion. Ideally once created the initial instructor client, the rest of the code should not have references to the underlying model being code

lazyhope: > maybe have a provider keyword so it doesn't mess up openai stuff
OpenAI has a similar way of merging messages of the same role, but this function can also be used by Anthropic exclusively (like what current implementation does), do you mean you also want openai to use this function?

tests/test_new_client.py
lazyhope: Done

tests/test_utils.py
lazyhope: > I reckon the previous implementation above (with the \n instead of the extend method for merging messages) was better and didn't break openAI things. I'd be in favour of not needing to pass provider keywords if possible, the more generic the code the better in my opinion. Ideally once created the initial instructor client, the rest of the code should not have references to the underlying model being code

For the sake of not messing up with the current code base I can see you point, but I suppose OpenAI clients doesn't need to merge message anyway, and this merge function is mainly used for Anthropic message API which requires alternating roles in each message, and it's a convenient way to introduce image/tool use messages.

jxnl (approved):

jxnl:
is this good to merge?

lazyhope:
> is this good to merge?

Yeah, I was waiting for it to be merged these days 😂

PR #585, opened 2024-04-09
Add support for cohere models

lidiyam:
Trying to add support for cohere models (`command-r` and `command-r-plus`) leveraging their function calling ability.

Looking at the tool-use prompt template from the docs [here](https://docs.cohere.com/docs/prompting-command-r), the tools are defined as python functions and so I had the following idea:

1. generate python code that defines the desired pydantic object based on it's `.model_json_schema()`
2. insert it in the tool-use prompt that defines the tool function (which will take in the pydantic object as the sole parameter)
3. let llm generate the required arguments for the tool function as usual
4. parse it
5. retry if failed to parse till `max_retries` is reached

This leads to less retries than asking the model to output json based on `.model_json_schema()` directly, but is more costly for simple use cases (non-nested pydantic objects) where the extra llm call is not really necessary.

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit ba6c2462af7e07d745bd6e00392aca8ea8f7861f.  | 
|--------|--------|

### Summary:
This PR adds support for Cohere models in the Instructor library, introducing a new mode and function to handle Cohere's function calling ability, along with necessary updates in various files, tests, and dependencies.

**Key points**:
- Added support for Cohere models in `instructor` library.
- Introduced a new mode, `COHERE_TOOLS`, in `instructor/mode.py`.
- Added a new function, `from_cohere`, in `instructor/client_cohere.py` to handle Cohere's function calling ability.
- Updated `instructor/__init__.py` to import `from_cohere` if Cohere is available.
- Updated `instructor/function_calls.py` to handle `COHERE_TOOLS` mode.
- Updated `instructor/process_response.py` to handle `COHERE_TOOLS` mode.
- Updated `instructor/retry.py` to handle `COHERE_TOOLS` mode.
- Updated `instructor/utils.py` to include `COHERE` in `Provider` enum.
- Updated `pyproject.toml` and `requirements.txt` to include Cohere as a dependency.
- Added tests for the new functionality in `tests/test_new_client.py`.

----


instructor/client_cohere.py
```diff
@@ -0,0 +1,160 @@
...
+        client, (cohere.Client)
+    ), "Client must be an instance of cohere.Cohere"
+
+    def create_cohere_structured_output(**kwargs):
+        """
+        (1) generates a tool snippet for a pydantic object
+        (2) formats a tool use prompt with the tool snippet
+        (3) sends the tool use prompt to cohere client chat
+        """
+        if "messages" in kwargs and "response_model" in kwargs:
```
jxnl, line 40: oh my god this is brutal

instructor/client_cohere.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def format_tool_use_prompt(conversation: list, tool_snippet: str) -> str:
+    """Returns tool use prompt with code for tools being provided by `tool_snippet`."""
+    preamble = """\
+<BOS_TOKEN><|START_OF_TURN_TOKEN|><|SYSTEM_TOKEN|># Safety Preamble
+The instructions in this section override those in the task description and style guide sections. Don't answer questions that are harmful or immoral.
+
+# System Preamble
+## Basic Rules
```
jxnl, line 135: is this their system message???

instructor/function_calls.py
```diff
@@ -137,6 +141,22 @@ def parse_anthropic_json(
...
+        completion,
+        validation_context: Optional[Dict[str, Any]] = None,
+        strict: Optional[bool] = None,
+    ) -> BaseModel:
+        text = completion.text
+        json_text = extract_json_from_codeblock(text)
+        function_dict = json.loads(json_text)
+        params = function_dict['parameters'] if 'parameters' in function_dict else function_dict
+        arg_name = cls.__name__.lower()
+        return cls.model_validate_json(
```
jxnl, line 156: this can just be model_validate() without the dumps

tests/test_new_client.py
```diff
@@ -176,3 +178,55 @@ def test_client_anthropic_response():
     )
     assert user.name == "Jason"
     assert user.age == 10
+
+
+def test_client_cohere_response():
```
jxnl, line 183: can you mark these as skip for now since i dont have cohere apis

jxnl:
Do you think we're better off just asking for a json?

lidiyam:
Ok asking for json directly removes a lot of the gnarliness 😅

instructor/client_cohere.py
```diff
@@ -0,0 +1,37 @@
+import cohere
+import instructor
+
+from typing import overload
+
+
+@overload
+def from_cohere(
+    client: cohere.Client,
```
jxnl, line 9: is there an async client we need to watch out for?

lidiyam: oh yes there's [`cohere.AsyncClient`](https://github.com/cohere-ai/cohere-python/blob/c7d5d5ea8451d35ea054187e24356b99cb484c88/src/cohere/client.py#L248)

for some reason `instructor.utils.is_async` is not recognizing its `client.chat` function as async and so patching doesn't work...will send a follow up a bit later today

lidiyam: Should support both now!

jxnl:
do you mind just update the docs  or adding a generic  `/hub` example? so we can tweet this out and het some cohere attention :)

lidiyam:
Ok added a few basic examples (like with `User` or `Group` extraction) and then a "tutorial" on doing document segmentation with a cohere client (since `cmd-r+` has 128k context length it's a good use case for it).

docs/examples/document_segmentation.md
```diff
@@ -0,0 +1,143 @@
+# Document Segmentation
```
jxnl, line 1: you'll have to add this in the mkdocs.yml file
or even reference it in examples/index.md

Issue #600, opened 2024-04-14
Support for Python 3.9


talperetz:
**Is your feature request related to a problem? Please describe.**
I am currently working on a project that utilizes **Vercel Functions**, which only supports [Python 3.9](https://github.com/orgs/vercel/discussions/639).

**Describe the solution you'd like**
Support for python 3.9 for until other major cloud providers will align on > 3.9 (AWS Lambda, Vercel, Netlify etc..)

**Describe alternatives you've considered**
Using json mode with openai instead of instructor.

PR #601, opened 2024-04-14
Support python 3.9

RedTachyon:
I've been tracking down similar bugs for way too long, so figured it shouldn't be a huge problem.

Note: CI isn't passing right now because it seems like it hasn't passed in quite some time on the main branch. That being said, the errors on 3.9 seem to be the same as >=3.10, so this should work as support.

Re: #600 and

image

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 10951e4c24304c65edb723f2e7a6adbe487e4aad.  | 
|--------|--------|

### Summary:
This PR adds support for Python 3.9, updates the Python version in the GitHub workflows and `pyproject.toml`, and makes changes to type annotations.

**Key points**:
- Added support for Python 3.9.
- Updated Python version in GitHub workflows and `pyproject.toml`.
- Imported `annotations` from `__future__` in multiple files for postponed evaluation of type annotations.
- Modified usage of `Union` in type hints.

----


instructor/client_anthropic.py
```diff
@@ -6,15 +8,19 @@
 
 @overload
 def from_anthropic(
-    client: anthropic.Anthropic | anthropic.AnthropicBedrock | anthropic.AnthropicVertex,
+    client: anthropic.Anthropic
+    | anthropic.AnthropicBedrock
```
jxnl, line 12: Does this need to be union

instructor/client_anthropic.py
```diff
@@ -6,15 +8,19 @@
...
     mode: instructor.Mode = instructor.Mode.ANTHROPIC_JSON,
     **kwargs,
 ) -> instructor.Instructor: ...
 
 
 @overload
 def from_anthropic(
-    client: anthropic.AsyncAnthropic | anthropic.AsyncAnthropicBedrock | anthropic.AsyncAnthropicVertex,
+    client: anthropic.AsyncAnthropic
+    | anthropic.AsyncAnthropicBedrock
```
jxnl, line 22: Union again?

instructor/client_anthropic.py
RedTachyon: You mean explicitly rewriting it as `Union[...]` as opposed to `X | Y | Z`? It would be one way of fixing it, the other way is adding `from __future__ import annotations` on top of the file, which is what I opted for because `X | Y | Z` is just such a nice syntax.

jxnl: pertect i didnt kno!

jxnl:
Looks good! Email jason@jxnl.co

PR #622, opened 2024-04-22
Fix anthropic usage and tools

lazyhope:
This PR fixes two issues with the current implementation:
- In `ANTHROPIC_TOOLS` mode when Claude didn't provide a tool invocation, an index error occurs
- `utils.update_total_usage` only works with OpenAI's `ChatCompletion`, so Anthropic's message usage doesn't update

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 7ceb1b5fa28bca972bd4d882c27b57e8b9eee1f3  | 
|--------|

### Summary:
This PR fixes issues related to Anthropic's usage and tools, updates the `update_total_usage` function to work with Anthropic's message usage, and adds a test case for retry error scenario.

**Key points**:
- Fixed index error in `parse_anthropic_tools` function in `function_calls.py` when there are no tool calls.
- Updated `update_total_usage` function in `utils.py` to work with Anthropic's message usage.
- Modified `retry.py` to handle cases when there is no tool invocation.
- Added a test case in `test_simple.py` to test the retry error scenario.

----

### DIFF
--- a/instructor/function_calls.py
+++ b/instructor/function_calls.py
@@ -4,7 +4,7 @@ from typing import Annotated, Any, Dict, Optional, Type, TypeVar
 
 from docstring_parser import parse
 from openai.types.chat import ChatCompletion
-from pydantic import BaseModel, Field, create_model, validate_call
+from pydantic import BaseModel, Field, TypeAdapter, create_model
 
 from instructor.exceptions import IncompleteOutputException
 from instructor.mode import Mode
@@ -120,13 +120,11 @@ class OpenAISchema(BaseModel):  # type: ignore[misc]
     ) -> BaseModel:
         tool_calls = [c.input for c in completion.content if c.type == "tool_use"]
 
-        @validate_call
-        def check_tool_call(
-            tool_calls: Annotated[list, Field(min_length=1, max_length=1)],
-        ):  # raise ValidationError if tool_calls is not of length 1
-            return tool_calls[0]
+        tool_calls_validator = TypeAdapter(
+            Annotated[list, Field(min_length=1, max_length=1)]
+        )
+        tool_call = tool_calls_validator.validate_python(tool_calls)[0]
 
-        tool_call = check_tool_call(tool_calls)
         return cls.model_validate(tool_call, context=validation_context, strict=strict)  # type:ignore
 
     @classmethod
--- a/instructor/function_calls.py
+++ b/instructor/function_calls.py
@@ -119,9 +119,12 @@ class OpenAISchema(BaseModel):  # type: ignore[misc]
         validation_context: Optional[Dict[str, Any]] = None,
         strict: Optional[bool] = None,
     ) -> BaseModel:
-        tool_call = [c.input for c in completion.content if c.type == "tool_use"][0]
+        tool_calls = [c.input for c in completion.content if c.type == "tool_use"]
+        tool_call = tool_calls[0] if tool_calls else {}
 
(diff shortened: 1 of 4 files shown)


instructor/utils.py
```diff
@@ -93,11 +94,16 @@ async def extract_json_from_stream_async(
...
-        total_usage.completion_tokens += response.usage.completion_tokens or 0
-        total_usage.prompt_tokens += response.usage.prompt_tokens or 0
-        total_usage.total_tokens += response.usage.total_tokens or 0
+    response_usage = getattr(response, "usage", None)
+    if isinstance(response_usage, OpenAIUsage):
+        total_usage.completion_tokens += response_usage.completion_tokens or 0
+        total_usage.prompt_tokens += response_usage.prompt_tokens or 0
+        total_usage.total_tokens += response_usage.total_tokens or 0
         response.usage = total_usage  # Replace each response usage with the total usage
+    elif isinstance(response_usage, AnthropicUsage):
```
jxnl, line 103: will this break if anthropic is not installed

lazyhope: You are right, this can be avoided using try import, but models of other providers will still have undefined behavior.

lazyhope:
Also the `RetryError` exception handling is not working in the `retry_async` function: `'RetryError' object has no attribute 'n_attempts'`, and I suppose `raise InstructorRetryException` within the try block appears to be a mistake. @jxnl

instructor/function_calls.py
```diff
@@ -119,9 +119,12 @@ def parse_anthropic_tools(
         validation_context: Optional[Dict[str, Any]] = None,
         strict: Optional[bool] = None,
     ) -> BaseModel:
-        tool_call = [c.input for c in completion.content if c.type == "tool_use"][0]
+        tool_calls = [c.input for c in completion.content if c.type == "tool_use"]
+        tool_call = tool_calls[0] if tool_calls else {}
```
jxnl, line 123: we should throw an error there right?

like assert len(tool_calls) == 1, "Please call exactly one tool"

instructor/utils.py
```diff
@@ -93,11 +93,24 @@ async def extract_json_from_stream_async(
...
+
+    # Anthropic usage
+    try:
+        from anthropic.types import Usage as AnthropicUsage
+
+        if isinstance(response_usage, AnthropicUsage):
+            total_usage.input_tokens += response_usage.input_tokens or 0
+            total_usage.output_tokens += response_usage.output_tokens or 0
+            response.usage = total_usage
+    except ImportError:
```
jxnl, line 111: log the exception here.

jxnl (changes requested):
this looks great, left 2 comments

instructor/function_calls.py
lazyhope: I left the error indication here:
https://github.com/jxnl/instructor/blob/230276e7067ec232448ed7394c95d1b697a87173/instructor/retry.py#L84
By passing an empty object, the language model will also understand what fields are missing

instructor/utils.py
lazyhope: I suppose we should log outside the exception handling part to indicate that no compatible usage was found? (The usage is neither OpenAI's `ChatCompletionUsage` nor `Anthropic` Usage)

instructor/function_calls.py
lazyhope: But it may miss models whose fields all has a default value, let me change that

jxnl (approved):

Issue #623, opened 2024-04-23
Run Pyright for CI Checks


max-muoto:
**Is your feature request related to a problem? Please describe.**
[Pyright](https://github.com/microsoft/pyright) is effectively the default type-checker for Python in VSCode, through Pylance. It's likely many Instructor users have Pylance/Pyright enabled through VSCode, which can lead to a variety of type-checking errors in places where Pyright is stricter than MyPy

**How is Pyright stricter than MyPy?**
Often times instructor will deal with generic classes without passing parameters, MyPy is usually fine with this, just inferring `Any` for anything not filled in. Let's take this example:

```python
from typing import Callable, reveal_type
    
def foo(create: Callable):
    reveal_type(create)
```

For MyPy, `def (*Any, **Any) -> Any` is revealed, while for Pyright `create: (...) -> Unknown` is revealed. As a result, those with stricter type-checking settings on will get run into errors whenever Instructor, does something like this, usually with just `Callable` or `Awaitable`. This same principle also principle also applies for `dict`, `list`, and other ggenerics. `dict == dict[unknown, unknown`], etc.

**Describe the solution you'd like**
Additionally run Pyright, alongside MyPy in CI checks, and fix existing Pyright violations

**Fixing existing violations**

I think fixing most existing issues wouldn't be overly difficult, in my cases it would just be providing more context, `Callable -> Callable[..., Any]`, `dict -> dict[str, Any]`, etc., however Pyright does catch typing errors that MyPy hasn't.

With the "standard" as the type-checking mode on, it catches about 40 errors, which I think is pretty doable, there might need to be some ignores in places with Instructor is dealing with third-party libraries that could be better typed, but I think there will be a lot of valuable in the primary API being fully compliant with Pyright.

**Possible issues**

I would say one partial issue is that Pyright actually has far fewer bugs, which means we'll have to be careful about writing code that is compatible across both. I would in the long-term, you could just run Pyright.

MyPy currently sits at about 2.7K issues, with Pyright at 30. There's also a repo going over the issues said issues in MyPy, to ensure they weren't present in Pyright: https://github.com/erictraut/mypy_issues

I would argue for the most accurate type-checking ossible, and so being able to lean in typing that's broken in MyPy, but works in Pyright, you could just run the later long-term. But for the sake of getting the most compatible across users, I think think running both would suffice, and add typing ignores where there is some key difference across the two.

[...]

PR #630, opened 2024-04-29
Implement Pyright for Type-Checking

max-muoto:
Fixes: https://github.com/jxnl/instructor/issues/623

This PR implements Pyright for type-checking, with the strictest settings enabled, fully replacing MyPy, which is partially broken at the moment. 

Realize this has gotten pretty large, just let me know if you would like me to break out any individual changes into separate PRs (Ruff config tweak, CI, etc) and can do that.

## MyPy vs Pyright

[Pyright](https://github.com/microsoft/pyright)is the default type-checker for VSCode from Microsoft, and exists as its own CLI tool that you can utilize for type-checking your codebase. It's the only Python type-checker that's fully supports the [latest Python typing spec](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html). Outside of that, it has several key differentiating factors from MyPy, which are touched [on here](https://microsoft.github.io/pyright/#/mypy-comparison).

Some main things that I think are worth calling out:

### Inferred Returned Types/Type-Checking for Unannotated Code

Return-types are always inferred in Pyright.

```python
def foo():
    return "bar"

result = foo()
reveal_type(result) # Revealed type is `Literal[bar]`
```

With this, Pyright will always type-check external libraries, using whatever internal type hints they provided, if stubs aren't given to take precedence. Unlike MyPy, type-checking is performed at all times unless a library is specifically stubbed out to always return `Any`.

### Unknown vs Any

When a type isn't provided, MyPy will put an `Any` in its place. Pyright on the other hand forces you to explicitly use `Any`, inferring `Unknown` is no generic param is provided. Some examples of the implications of this are discussed in this [issue](https://github.com/jxnl/instructor/issues/623).

### Narrowing/TypeGuards

Pyright supports a variety of type-guards that MyPy doesn't. See [here](https://microsoft.github.io/pyright/#/type-concepts-advanced?id=type-guards).

### Support for New Features

Pyright already has support for 3.13 features that we can start to utilize through `type_extensions`, for example generic defaults and support for the deprecated decorator. On the other MyPy still does not have full support for all 3.12 features, see [here](https://github.com/python/mypy/issues/15238).

### Less Bugs

Generally from my experience Pyright has far fewer bugs (34 GitHub issues) than MyPy (2.6k GitHub issues), and MyPy bugs are cross-checked against Pyright to ensure they aren't present.

### Editor Support

Assuming you have Pylance installed in VSCode or any other editor, you'll get type-checking right as you modify code that will simply pull from the config in our `pyproject.toml`.

## This PR

What exactly is this PR doing? It enables Pyright, with its strictest settings, with the exception of allowing deprecated methods (we're still using `classmethod`, which is deprecated), and allowing for redundant `isinstance` checks. The later is mostly around the fact that we can't ensure that users are doing static type-checking, so "redundant" runtime type-checking still serves a purpose.

I'll go over some of the general fixes to maybe try and help answer questions that might come from this PR.

### Typed Dicts

You'll see the replacement of some inline `dict` types with TypeDicts, this mostly around the fact that kwargs unpacking is not type-safety without them.

Let's take this example:

```python
additional_params: Dict[str, Union[str, Dict[str, Union[int, float, str]]]] = {}
if hyperparameters_dict:
    additional_params["hyperparameters"] = hyperparameters_dict
if validation_file:
    additional_params["validation_file"] = validation_file
if model_suffix:
    additional_params["suffix"] = model_suffix
job = client.fine_tuning.jobs.create(
    training_file=file_id,
    model=model,
    **additional_params,
)
```

Pyright doesn't consider this type-safe, since it doesn't know the types of the specific fields being passed into `create`, since the value is simply one of int, float, str, regardless of the key. To make this type-safe, we can create a typed dictionary defining the fields and their respective types.

### Consistent Type Syntax

I enabled the Ruff rule [UP006](https://docs.astral.sh/ruff/rules/non-pep585-annotation/) to ensure we're using consistent type syntax. As of 3.9, explicitly importing `Dict`, `List`, etc is deprecated, as you can simply use the primitives as generics. This rule is auto-fix for the most part.

### Typer/CLI Fixes

Removed unnecessary typing ignores by avoiding explicitly passing in `...`, as it is the default, and corrected types that should have been `Optional`.

### Pydantic

Had to add a variety of typing ignores toimports for `create_model`, due to missing generics. [Made a PR fixing this](https://github.com/pydantic/pydantic/commit/37fa1bce922d1849e700ac102d3062a51d4c2ca2), which was pushed, so once we upgrade to their next release, we can remove these.

### CI / Pre-Commit

Pyright needs our extra dependencies to properly type-check everything, so tweaking the CI to simply do an install with Poetry, instead of the `requirements.txt`. I additionally updated the `pre-commit` to run Pyright locally.

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 55ac2b96f4f1c171302f266cdc5bf654c138d78d  | 
|--------|

### Summary:
This PR implements Pyright for type-checking, updates various files for type safety, uses new Python 3.9 typing syntax, and handles kwargs updates.

**Key points**:
- Implements Pyright for type-checking, replacing MyPy.
- Adds a new GitHub workflow for Pyright.
- Updates various files to ensure type safety with Pyright.
- Uses new typing syntax introduced in Python 3.9.
- Updates handling of kwargs in several places.
- Adds type ignores for some imports from `pydantic.create_model` due to missing generics.
- Updates `requirements.txt` and `pyproject.toml` files to include Pyright.

----


max-muoto:
@jxnl Still have some cleaning up here to do, but was wondering if you might be able to run the actions for me? Hoping to test out the new Pyright check.

.github/workflows/pyright.yml
```diff
@@ -0,0 +1,47 @@
...
+    matrix:
+      os: [ubuntu-latest, macos-latest]
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
```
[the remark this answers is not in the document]
max-muoto: https://github.com/jxnl/instructor/pull/630/commits/168c8f61288f72038c171c97487b70f8d7df14b7

instructor/client.py
```diff
@@ -60,16 +57,38 @@ def completions(self) -> Self:
     def messages(self) -> Self:
         return self
 
+    @overload
+    def create(
```
max-muoto, line 61: This usage of overloading might seem funky. But it's the only way we can get accurate type-checking here without refactoring `AsyncInstructor`, as the core issue here is thatthe subclass is doing an incompatible method override, since adding the `async` keyword implicitly adds `Awaitable` to the return-type.

Thus, `create` should return `Union[T, Awaitable[T]]:` However, doing this without overloads will lead to an inaccurate unioned type whenever calling `create` on `Instructor`. Overloading like this, we can specify that for `Instructor`, it will return the non-awaited version, whileour subclass will return the awaited version.

The main alternatives would be 1) refactor `AsyncInstructor` so it simply inherits from a common base class that `Instructor` also inherits from. 2) add typing ignores.

Happy to go with either or, depending on what you prefer, or leave this as is, and refactor it in a follow-up @jxnl. Let me know what you think.

instructor/cli/files.py
```diff
@@ -1,6 +1,8 @@
+# type: ignore - stub mismatched
```
max-muoto, line 1: We're really not following the stubs at all here, so not too sure if this was just from when we were on a previous version of the OpenAI client? Mainly, that we treat the pydantic models they return as dictionaries, but maybe there's something I'm missing. Thought it would be easiest to ignore this for now, and go back to it later.

instructor/cli/usage.py
```diff
@@ -46,13 +46,7 @@ async def get_usage_for_past_n_days(n_days: int) -> List[Dict[str, Any]]:
...
 
 # Define the cost per unit for each model
-# Add temporary body type hint here because mypy may infer the dict type
-# from the first few items (?) in the dict, which may not be representative of
-# the entire dict.
-MODEL_COSTS: Dict[
-    ModelNames,
-    Union[Dict[str, float], float],
-] = {
+MODEL_COSTS = {
```
max-muoto, line 49: Annotation is inferred properly now.

instructor/client_anthropic.py
```diff
@@ -35,9 +35,10 @@ def from_anthropic(
         | anthropic.AnthropicBedrock
         | anthropic.AsyncAnthropicBedrock
         | anthropic.AsyncAnthropicVertex
+        | anthropic.AnthropicVertex
```
max-muoto, line 38: Presentin an overload, so we needed this here.

instructor/distil.py
```diff
@@ -5,19 +5,36 @@
...
+
+class OpenAIChatKwargs(TypedDict):
+    messages: list[ChatCompletionMessageParam]
+    functions: NotRequired[list[dict[str, Any]]]
 
 
 class FinetuneFormat(enum.Enum):
-    MESSAGES: str = "messages"
-    RAW: str = "raw"
+    MESSAGES = "messages"
```
max-muoto, line 36: Inaccurate to type these with `str`, since they're considered literals.

instructor/distil.py
```diff
@@ -149,33 +166,38 @@ def distil(
...
                     fn=fn,
                     args=args,
                     kwargs=kwargs,
                     base_model=return_base_model,
                 )
                 return self.client.chat.completions.create(
-                    **openai_kwargs, model=model, response_model=return_base_model
+                    **openai_kwargs,
+                    model=model,
+                    response_model=return_base_model,  # type: ignore - TODO figure out why `response_model` is not recognized
```
max-muoto, line 187: Missing from the OpenAI stubs.

instructor/dsl/maybe.py
```diff
@@ -53,23 +53,22 @@ def __bool__(self):
     Returns:
         MaybeModel (Type[BaseModel]): A new Pydantic model that includes fields for `result`, `error`, and `message`.
     """
-
-    fields = {
-        "result": (
+    return create_model(
```
max-muoto, line 56: We would have to use a type-dict to unpack in a type-safe manner, so it's just going to be less verbose to directly pass everything in.

examples/avail/run_mixtral.py
```diff
@@ -25,7 +25,7 @@ class DateRange(BaseModel):
         default=None,
         description="If the date range repeats, and how often, this way we can generalize the date range to the future., if its special, then we can assume it is a one time event.",
     )
-    days_of_week: List[
+    days_of_week: list[
```
jxnl, line 28: oh interesting, i have a habit of using List vs list, whats the reasoning?

tests/test_function_calls.py
```diff
@@ -12,7 +12,7 @@
 
 
 @pytest.fixture  # type: ignore[misc]
-def test_model() -> Type[OpenAISchema]:
+def test_model() -> type[OpenAISchema]:
```
jxnl, line 15: same here type vs Type

jxnl:
quick look over 3.10 tests and it looks like theres some subclass issues, that point to list[int] vs List[int]. Maybe i'm mistaken

instructor/client_anthropic.py
jxnl: +1

examples/avail/run_mixtral.py
max-muoto: Before Python 3.9, you weren't able to use the primitive `list`, `dict`, or `type` as generics. Now that that you can, technically `typing.List` is deprecated: https://peps.python.org/pep-0585/

max-muoto: This was just an autofix from Ruff, which will help everything stay consistent in regards to this.

max-muoto:
> the base tests (without openai calls) should be passing on main

Not able to run all of the tests locally, but should be good now!

tests/test_function_calls.py
max-muoto: Same answer as above.

max-muoto:
Thanks for the quick review @jxnl I'll try and follow-up with some additional clean-up in regards to the overloads, and try and improve some of the Ruff rules regarding our typing config.

PR #633, opened 2024-04-30
Enable PyUpgrade Ruff Rule-Set

max-muoto:
As a follow-up to rolling out Pyright, this PR enables the full Ruff Pyupgrade [rule-set](https://docs.astral.sh/ruff/rules/#pyupgrade-up), to help make sure we're using the most modern typing syntax in a consistent matter, and avoiding deprecated imports. 

### Avoiding Typing Imports

Imports for types such as `Callable`, and `Awaitable` should be imported from `collections.abc` instead of `typing` post Python 3.9.

### Modern Union Syntax

Using the pipe operator `|` instead of `Union`, when `from __future__ import annotations` is present. (This can be all files once the base is 3.10)

### Caching

Replacing `@lru_cache(maxsize=None)` with `functools.cache`: https://docs.astral.sh/ruff/rules/lru-cache-with-maxsize-none/. The later being a drop-in replacement.

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 4ad3a9c18f856e3cf299087a231ca51151b2edb9  | 
|--------|

### Summary:
This PR applies the full Ruff Pyupgrade rule-set, updating imports, decorators, and file handling methods across multiple scripts to enforce modern Python coding standards.

**Key points**:
- Enabled full Ruff Pyupgrade rule-set in `.ruff.toml`.
- Updated imports, replacing `typing.Iterable` with `collections.abc.Iterable` across various files.
- Replaced `@functools.lru_cache()` with `@functools.cache` in applicable instances.
- Modified file handling by removing mode 'r' in `open()` function calls.
- Adjusted example scripts and instructor modules to align with modern Python practices.

----


jxnl:
theres explicit request for 3.9, will this still run?

max-muoto:
> theres explicit request for 3.9, will this still run? 

In regards to the union syntax? Yes, since Ruff is only adding it to files with deferred annotations.

PR #662, opened 2024-05-12
fix: remove URLs in pydantic error messages

### LINKED ISSUE #509: Pydantic includes URLs with validation errors which aren't useful for the LLM

Pydantic includes a URL by default to look up more information on the error which isn't useful for the LLM and also takes up token space.

```
1 validation error for Boop
name
  Value error, 'bla' is not a valid boop. [type=value_error, input_value='bla', input_type=str]
    For further information visit https://errors.pydantic.dev/2.6/v/value_error
```

It'd be nice to have more control over what happens here. Pydantic let's you dump the errors if you catch the exception via `errors(` and then specify `include_url`. It feels like the default should be to not include that.

Just giving more control over the reask flow could let people solve this themselves.

ssonal:
Fixes #509

Pydantic includes error URLs in ValidationErrors which means extra tokens used in LLM calls.

Pydantic relies on an environment variable to control this setting and is on by default. https://github.com/pydantic/pydantic-core/blob/e1fc99dd3207157aad77defc20ab6873fd268b5b/python/pydantic_core/_pydantic_core.pyi#L818

This change sets the environment variable dynamically when the package is invoked.

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 5c2a35961c141c39afd1c83286195548dbf86b24  | 
|--------|

### Summary:
This PR sets an environment variable to remove URLs from Pydantic error messages and includes a test to verify the change.

**Key points**:
- Set `PYDANTIC_ERRORS_INCLUDE_URL` to `0` in `set_env_variable` in `/instructor/utils.py`
- Call `set_env_variable` in `/instructor/__init__.py`
- Add test in `/tests/test_function_calls.py` to verify URLs are not in Pydantic error messages

----

### DIFF
--- a/docs/concepts/reask_validation.md
+++ b/docs/concepts/reask_validation.md
@@ -200,6 +200,45 @@ The docs are currently incomplete, but we have a few advanced validation techniq
 1. Validate the entire object with all attributes rather than one attribute at a time
 2. Using some 'context' to validate the object: In this case, we use the `context` to check if the citation existed in the original text.
 
+## Optimizing Token usage
+
+Pydantic automatically includes a URL within the error message itself when an error is thrown so that users can learn more about the specific error that was thrown. Some users might want to remove this URL since it adds extra tokens that otherwise might not add much value to the validation process.
+
+We've created a small helper function that you can use below which removes this url in the error message
+
+```python hl_lines="6"
+from instructor.utils import disable_pydantic_error_url
+from pydantic import BaseModel, ValidationError
+from typing_extensions import Annotated
+from pydantic import AfterValidator
+
+disable_pydantic_error_url() # (1)!
+
+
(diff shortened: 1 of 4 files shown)


jxnl:
I would prefer this to be a piece of the documentation rather than something that we force users to use.

ssonal:
This does sound like a sane default to me, especially when retries come into the picture. Would be cumbersome to have to set this env variable each time. But maybe something like a config file makes more sense instead?

Mr-Ruben:
I have been waiting for this fix since March-2024.

I am with @ssonal   
Who wants a useless url in the error message?  (+noise, +tokens, no benefit)
Just document how to enable if someone ever wants and move on.

jxnl:
see #757

PR #682, opened 2024-05-18
Anthropic streaming support

ssonal:
Streaming is now live on Anthropic. This change introduces json parsing support for streamed chunks. 

https://docs.anthropic.com/en/api/messages-streaming

### DIFF
--- a/instructor/dsl/iterable.py
+++ b/instructor/dsl/iterable.py
@@ -79,7 +79,12 @@ class IterableBase:
     ) -> Generator[str, None, None]:
         for chunk in completion:
             try:
-                if chunk.choices:
+                if mode == Mode.ANTHROPIC_JSON:
+                    if json_chunk := chunk.delta.text:
+                        yield json_chunk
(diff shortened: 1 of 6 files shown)


instructor/dsl/iterable.py
```diff
@@ -79,7 +79,12 @@ def extract_json(
     ) -> Generator[str, None, None]:
         for chunk in completion:
             try:
-                if chunk.choices:
+                if mode == Mode.ANTHROPIC_JSON:
+                    if json_chunk := chunk.delta.text:
+                        yield json_chunk
+                if mode == Mode.ANTHROPIC_TOOLS:
+                    yield chunk.model_extra.get("delta", "").get("partial_json", "")
```
jxnl, line 86:
```suggestion
                    yield chunk.model_extra.get("delta", {}).get("partial_json", "")
```

should it be {} since `"".get` would error?

jxnl:
can you try and update the documentation on anthropic docs?

@ellipsis-dev

jxnl (approved):

PR #706, opened 2024-05-27
fixed incorrect argument

Elektra58:
Argument `openai_client` is incorrect and needs to be replaced by `client`


Elektra58:
Replaced by #707 because there are two instances where the incorrect argument is used.

PR #719, opened 2024-05-31
async gemini support

Cruppelt:
| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit b1602cb76a9d13029d08cc0f3bbc8b92e34dedaf  | 
|--------|--------|

### Summary:
This PR introduces asynchronous support to the `from_gemini` function in the `instructor/client_gemini.py`, allowing it to handle both synchronous and asynchronous operations.

**Key points**:
- Added async support in `from_gemini` function in `instructor/client_gemini.py`
- Overloads added for `from_gemini` to handle sync and async modes
- Conditional logic based on `use_async` to return appropriate instructor type
- Uses `generate_content_async` for async and `generate_content` for sync

----

### DIFF
--- a/instructor/client_gemini.py
+++ b/instructor/client_gemini.py
@@ -32,9 +32,7 @@ def from_gemini(
     use_async: bool = False,
     **kwargs: Any,
 ) -> instructor.Instructor | instructor.AsyncInstructor:
-    assert (
-        mode == instructor.Mode.GEMINI_JSON
-    ), "Mode must be instructor.Mode.GEMINI_JSON"
+    assert mode == instructor.Mode.GEMINI_JSON, "Mode must be instructor.Mode.GEMINI_JSON"
 
     assert isinstance(
         client,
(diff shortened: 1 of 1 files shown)


Cruppelt:
lmk thoughts on the use_async - The other clients have different Async clients that we can check the instance of but in this case Google/Gemini has the same client.

jxnl (approved):

jxnl:
Honestly, this looks good. Have you tested it locally yet?

Cruppelt:
> Honestly, this looks good. Have you tested it locally yet?

Yes, typehead working too.

image

ssonal:
retry.py retry_async logic needs to also be updated to handle ValidationErrors. Gemini uses a different messages format.

Cruppelt:
> retry.py retry_async logic needs to also be updated to handle ValidationErrors. Gemini uses a different messages format.

Have an example that will break? Happy to add

ssonal:
> > retry.py retry_async logic needs to also be updated to handle ValidationErrors. Gemini uses a different messages format.
> 
> Have an example that will break? Happy to add

Just saying that https://github.com/jxnl/instructor/blob/3d8d4495edd6740fff2f1bfbb0335b198f01f0dd/instructor/retry.py#L159-L169 and https://github.com/jxnl/instructor/blob/3d8d4495edd6740fff2f1bfbb0335b198f01f0dd/instructor/retry.py#L225-L232 should match. 

`GEMINI_JSON` mode requires different handling of args for retries.

PR #726, opened 2024-05-31
feat: gemini tool calling support

ssonal:
- added `GEMINI_TOOLS` mode for data extraction through function calling
- added compatibility for streaming, partials, iterables
- updated tests
- updated docs

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit a6c95aae65f36fac9b8ccc283e275120bae7a247  | 
|--------|--------|

### Summary:
Added `GEMINI_TOOLS` mode to the `instructor` library for structured data extraction, updated documentation, tests, and added `jsonref` dependency.

**Key points**:
- Introduced `GEMINI_TOOLS` mode for structured data extraction via function calling
- Updated `README.md` and `docs/concepts/patching.md` for new mode
- Modified Python modules to support `GEMINI_TOOLS` mode
- Updated tests to include new mode
- Added `jsonref` dependency for `GEMINI_TOOLS` mode
- Ensured compatibility with streaming, partials, and iterables

----


tungalbert99:
Is GEMINI_TOOLS mode similar to the new `v1.5.3` that was released today with the `response_schema` specification?

docs/concepts/patching.md
```diff
@@ -19,6 +19,26 @@ from openai import OpenAI
 client = instructor.from_openai(OpenAI(), mode=instructor.Mode.TOOLS)
 ```
 
+### Gemini Tool Calling
+
+Gemini supports tool calling for stuctured data extraction. Gemini tool calling requires `jsonref` to be installed.
+
+!!! Warning
```
jxnl, line 26:
```suggestion
!!! warning "Limitations"
```

jxnl:
fuck i think i merged vertex ai PR and fucked up this one, give me some time and i try to resolve it

gokturkDev:
Hey any updates on this? We are getting a lot of json validation errors with gemini which is preventing us to go into production with it, so would absolutely love to get tools support. Would you be able to give a approx. date for this? @jxnl

jxnl:
@ivanleomk can you take ownership of this? test this locally and make sure everything works?

ivanleomk:
@jxnl yep I can do it.

ssonal:
@ivanleomk my bad, looks like I missed a case. Updated & should work now.

ivanleomk:
Not super familiar with Gemini, possible to show an example of how to setup GenerationConfig or how it's typically used? 

Want to make sure it works in both cases.

ssonal:
> Any idea what's up with this? Also, have you tried the native tool calling from gemini ( https://ai.google.dev/gemini-api/docs/function-calling ) , wondering how it stacks up against your specific implementation here?

@ivanleomk this is the native gemini tool calling implementation. See https://github.com/jxnl/instructor/blob/2a34d08314f902a765a21b95284404e4dc0d2636/instructor/process_response.py#L437-L443
Rest of this PR is config & response handling related.

ivanleomk:
@ssonal how much would it take to implement function calling using this instead?

https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/call-gemini-using-openai-library

Wondering if it might be easier to ensure libs remain consistent especially since now we are doing this weird prompt injection in the PR

```python
  message = dedent(
                f"""
                As a genius expert, your task is to understand the content and provide arguments to the functions provided. Make sure to provide the right function name and an openAPI compatible response!
                """
            )
            # check that the first message is a system message
            # if it is not, add a system message to the beginning
            if new_kwargs["messages"][0]["role"] != "system":
                new_kwargs["messages"].insert(
                    0,
                    {
                        "role": "system",
                        "content": message,
                    },
                )
```

ssonal:
> @ssonal how much would it take to implement function calling using this instead?
> 
> https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/call-gemini-using-openai-library
> 
> Wondering if it might be easier to ensure libs remain consistent especially since now we are doing this weird prompt injection in the PR
> 
> ```python
>   message = dedent(
>                 f"""
>                 As a genius expert, your task is to understand the content and provide arguments to the functions provided. Make sure to provide the right function name and an openAPI compatible response!
>                 """
>             )
>             # check that the first message is a system message
>             # if it is not, add a system message to the beginning
>             if new_kwargs["messages"][0]["role"] != "system":
>                 new_kwargs["messages"].insert(
>                     0,
>                     {
>                         "role": "system",
>                         "content": message,
>                     },
>                 )
> ```

@ivanleomk unfortunately openai library compatibility is only for Google VertexAI, and not for the Google Cloud AI platform as of today.
With regards to the system prompt, I didn't want to overwrite system prompts that the user may have included but maybe this snippet could be more robust.  Should I approach this differently?

ivanleomk:
@ssonal Hmm have you tried not including a system prompt itself? I think that might not be a bad experiment to try.

Also for some reason when I run this snippet

```python
import instructor
import google.generativeai as genai
import logfire

logfire.configure()

client = instructor.from_gemini(
    genai.GenerativeModel(), mode=instructor.Mode.GEMINI_TOOLS
)


class UserExtractValidated(BaseModel):
    name: str
    age: int

    @field_validator("name")
    @classmethod
    def validate_name(cls, v: str) -> str:
        if v.upper() != v:
            raise ValueError(
                "Name should be uppercase, make sure to use the `uppercase` version of the name"
            )
        return v


model = client.chat.completions.create(
    response_model=UserExtractValidated,
    strict=False,
    messages=[
        {"role": "user", "content": "Extract jason is 25 years old"},
    ],
)

print(model.model_dump_json(indent=2))
```

I get the error that the Tool name does not match. Gemini seems to be returning a function call called `run` for some reason

ssonal:
> @ssonal Hmm have you tried not including a system prompt itself? I think that might not be a bad experiment to try.
> 
@ivanleomk Sure, let me make this change and report back.

> I get the error that the Tool name does not match. Gemini seems to be returning a function call called `run` for some reason

Try with the 1.5-pro model. Flash struggles with validations and retries. Also the [default model](https://github.com/google-gemini/generative-ai-python/blob/efead6bea6768f6f4a3d90d348647b0a54fe2435/google/generativeai/generative_models.py#L72) is `gemini-pro` which is also much further behind in terms of perf. 

```python
client = instructor.from_gemini(
    client=genai.GenerativeModel(
        model_name="models/gemini-1.5-pro",
    ),
    mode=instructor.Mode.GEMINI_TOOLS,
)
```

ssonal:
> @ssonal Hmm have you tried not including a system prompt itself? I think that might not be a bad experiment to try.
> 
@ivanleomk updated with this change. All good - tests pass.

krish-bell:
What's left for this PR to be merged? I see that last activity was 3 weeks ago

ivanleomk:
Will look at this PR later in the week and update the code.

instructor/function_calls.py
```diff
@@ -249,6 +247,22 @@ def parse_gemini_json(
...
+        strict: Optional[bool] = None,
+    ) -> BaseModel:
+        tool_call = completion.parts[0].function_call
+        assert (
+            tool_call.name == cls.openai_schema["name"]  # type: ignore[index]
+        ), "Tool name does not match"
+        # reference https://ai.google.dev/gemini-api/tutorials/extract_structured_data
+        response = type(tool_call).to_dict(tool_call)["args"]
+
+        return cls.model_validate(response, context=validation_context, strict=strict)
```
ivanleomk, line 264: @jxnl i've been playing around with this and it seems like perhaps we might want to set strict to be False by default for gemini tools?

Let's take the simple example below

```
import instructor
import google.generativeai as genai

client = instructor.from_gemini(
    genai.GenerativeModel("gemini-1.5-pro"),
    mode=instructor.Mode.GEMINI_TOOLS,
)


class Weather(BaseModel):
    location: str
    hour: int
    temperature: float


    response_model=Weather,
    messages=[
        {
            "role": "user",
            "content": "The weather in Singapore at 5pm is 34.5 degree celcius",
        },
    ],
    strict=False,
)
```

When we process this in `process_response`, we get an error if strict isn't true. This seems to happen because when we coerce the protobuf to a json dictionary, it coerces all the ints into floats for some reason.

```python
@classmethod
    def parse_gemini_tools(
        cls: type[BaseModel],
        completion: ChatCompletion,
        validation_context: Optional[dict[str, Any]] = None,
        strict: Optional[bool] = None,
    ) -> BaseModel:
        tool_call = completion.parts[0].function_call
	print(tool_call)
        assert (
            tool_call.name == cls.openai_schema["name"]  # type: ignore[index]
        ), "Tool name does not match"
        # reference https://ai.google.dev/gemini-api/tutorials/extract_structured_data
        response = type(tool_call).to_dict(tool_call)["args"]
	print(response)

        return cls.model_validate(response, context=validation_context, strict=strict)
```

This in turn gives u the following output when we print out the value at each stage

```
name: "Weather"
args {
  fields {
    key: "temperature"
    value {
      number_value: 34.5
    }
  }
  fields {
    key: "location"
    value {
      string_value: "Singapore"
    }
  }
  fields {
    key: "hour"
    value {
      number_value: 17
    }
  }
}

{'hour': 17.0, 'location': 'Singapore', 'temperature': 34.5}
location='Singapore' hour=17 temperature=34.5
```

ivanleomk:
@ssonal do you know of a way to extract the value of a int from a protobuf without it becoming a float? I'm a bit worried about turning off the strict parsing for all our tests because of this one change.

Also, in terms of the tool_config, I think it makes sense to force a function call by passing in the `allowed_function_names` parameter

```
new_kwargs["tool_config"] = {
                "function_calling_config": {
                    "mode": "ANY",
                    "allowed_function_names": [response_model.__name__],
                },
            }
```

instructor/function_calls.py
```diff
@@ -135,6 +130,9 @@ def from_response(
         if mode == Mode.GEMINI_JSON:
             return cls.parse_gemini_json(completion, validation_context, strict)
 
+        if mode == Mode.GEMINI_TOOLS:
```
ivanleomk, line 133: This does functionally the same thing as the vertex_ai_tools function above. Why not we just reuse the vertexai_tools mode

ivanleomk:
```
if mode in {Mode.VERTEXAI_TOOLS, Mode.GEMINI_TOOLS}:
            return cls.parse_vertexai_tools(completion, validation_context, strict)
```

might be good to just rename that to parse_gemini_tools too

instructor/function_calls.py
ivanleomk: I realised that we're doing that already for the vertexai_tools so that's not entirely inconsistent with the existing approach. Then we can throw an error if users set strict=True in the function call

```python
@classmethod
    def parse_vertexai_tools(
        cls: type[BaseModel],
        completion: ChatCompletion,
        validation_context: Optional[dict[str, Any]] = None,
        strict: Optional[bool] = None,
    ) -> BaseModel:
        strict = False
        tool_call = completion.candidates[0].content.parts[0].function_call.args  # type: ignore
        model = {}
        for field in tool_call:  # type: ignore
            model[field] = tool_call[field]
        return cls.model_validate(model, context=validation_context, strict=strict)
```

ivanleomk:
I ran tests on the new GEMINI_TOOLS mode with pro and flash locally and this is the result

```
tests/llm/test_gemini/test_modes.py ........                                                    [  7%]
tests/llm/test_gemini/test_patch.py ........                                                    [ 14%]
tests/llm/test_gemini/test_retries.py ........                                                  [ 22%]
tests/llm/test_gemini/test_simple_types.py ...                                                  [ 25%]
tests/llm/test_gemini/test_stream.py ............                                               [ 36%]
tests/llm/test_gemini/evals/test_classification_enums.py ....................                   [ 55%]
tests/llm/test_gemini/evals/test_classification_literals.py ....................                [ 73%]
tests/llm/test_gemini/evals/test_entities.py ....                                               [ 77%]
tests/llm/test_gemini/evals/test_extract_users.py ............                                  [ 88%]
tests/llm/test_gemini/evals/test_sentiment_analysis.py ............                             [100%]
```

ivanleomk (approved):

jxnl (approved):

ivanleomk:
Ran tests once more

## VertexAI

```bash
tests/llm/test_vertexai/test_message_parser.py ....                                             [ 16%]
tests/llm/test_vertexai/test_modes.py ......                                                    [ 41%]
tests/llm/test_vertexai/test_retries.py ....                                                    [ 58%]
tests/llm/test_vertexai/test_simple_types.py ......                                             [ 83%]
tests/llm/test_vertexai/test_stream.py ....                                                     [100%]
```

## Gemini

```bash
tests/llm/test_gemini/test_modes.py ........                                                    [  7%]
tests/llm/test_gemini/test_patch.py ........                                                    [ 14%]
tests/llm/test_gemini/test_retries.py ........                                                  [ 22%]
tests/llm/test_gemini/test_simple_types.py ...                                                  [ 25%]
tests/llm/test_gemini/test_stream.py ............                                               [ 36%]
tests/llm/test_gemini/evals/test_classification_enums.py ....................                   [ 55%]
tests/llm/test_gemini/evals/test_classification_literals.py ....................                [ 73%]
tests/llm/test_gemini/evals/test_entities.py ....                                               [ 77%]
tests/llm/test_gemini/evals/test_extract_users.py ............                                  [ 88%]
tests/llm/test_gemini/evals/test_sentiment_analysis.py ............                             [100%]
```

Issue #668, opened 2024-05-14
`create_iterable` is broken


bryanhpchiang:
`create_iterable` from the docs does not work.

1. Go here: https://python.useinstructor.com/#streaming-iterables-create_iterable
2. Copy + paste + run the example
3. Error
```
Traceback (most recent call last):
  File "/Users/bryan/work/flow/iterable.py", line 14, in <module>
    users = client.chat.completions.create_iterable(
  File "/Users/bryan/work/flow/.venv/lib/python3.10/site-packages/instructor/client.py", line 184, in create_iterable
    return self.create_fn(
  File "/Users/bryan/work/flow/.venv/lib/python3.10/site-packages/instructor/patch.py", line 140, in new_create_sync
    response_model, new_kwargs = handle_response_model(
  File "/Users/bryan/work/flow/.venv/lib/python3.10/site-packages/instructor/process_response.py", line 191, in handle_response_model
    if is_simple_type(response_model):
  File "/Users/bryan/work/flow/.venv/lib/python3.10/site-packages/instructor/dsl/simple_type.py", line 40, in is_simple_type
    if isclass(response_model) and issubclass(response_model, BaseModel):
  File "/Users/bryan/miniconda3/envs/turbo/lib/python3.10/abc.py", line 123, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a class
```

```
(turbo) bryan@Bryans-MacBook-Pro-7 flow % python -V
Python 3.10.12
(turbo) bryan@Bryans-MacBook-Pro-7 flow % pip show openai
Name: openai
Version: 1.30.1
Summary: The official Python library for the openai API
Home-page:
Author:
Author-email: OpenAI <support@openai.com>
License:
Location: /Users/bryan/miniconda3/envs/turbo/lib/python3.10/site-packages
Requires: anyio, distro, httpx, pydantic, sniffio, tqdm, typing-extensions
Required-by: instructor, litellm, open-interpreter, self-operating-computer
(turbo) bryan@Bryans-MacBook-Pro-7 flow % pip show instructor
Name: instructor
Version: 1.2.6
Summary: structured outputs for llm
Home-page: https://github.com/jxnl/instructor
Author: Jason Liu
Author-email: jason@jxnl.co
License: MIT
Location: /Users/bryan/miniconda3/envs/turbo/lib/python3.10/site-packages
[...]

PR #737, opened 2024-06-09
Iterable Workaround

ivanleomk:
This helps solve issue #668 which highlighted an issue with how we're handling Iterables in Python 3.9. There's a bunch of new tests and a new wrapper function which checks the system python version before running the iterable check.


tests/dsl/test_simple_type.py
```diff
@@ -0,0 +1,51 @@
+import unittest
+from instructor.dsl.simple_type import is_simple_type
+from pydantic import BaseModel
+from enum import Enum
+import typing
+
+
+class SimpleTypeTests(unittest.TestCase):
```
jxnl, line 8: don't use unitests

instructor/dsl/simple_type.py
```diff
@@ -31,14 +31,27 @@ def __class_getitem__(cls, response_model: type[BaseModel]) -> type[BaseModel]:
         )
 
 
+def validateIsSubClass(response_model: type):
```
jxnl, line 34: snake case

tests/llm/test_openai/evals/test_entities.py
```diff
@@ -54,6 +54,7 @@ def ask_ai(content, model, client) -> DocumentExtraction:
                 "content": content,
             },
         ],
+        max_retries=4,
```
jxnl, line 57: doubt 4 is the answer...

Issue #627, opened 2024-04-24
Any valid Pydantic type should be supported (e.g. TypedDict)


ADR-007:
**Is your feature request related to a problem? Please describe.**
Currently, only Pydantic models are supported as response_model.
But in some cases, I want to use TypedDict instead. For example, I don't want to do massive refactoring, so I just want to add response schema validation to an existing job.

**Describe the solution you'd like**
I would like to able to use TypedDict as response_model. E.g.:
```py
class MyModel(TypedDict):
    my_key: str

result = client.chat.completions.create(
    messages=[...],
    response_model=MyModel,
)
```

It is very simple to implement in Pydantic V2:
```py
from pydantic import TypeAdapter

adapter = TypeAdapter(MyModel)
json_schema = adapter.json_schema()

result = prcess_llm(prompt, json_schema, ...)

my_model = adapter.validate_python(result)
```

jxnl:
Happy to take or for this in process response.

ivanleomk:
@ADR-007 just pushed up a PR which introduces this. Is this what you had in mind for your use case?

```python
from typing_extensions import TypedDict
from openai import OpenAI
import instructor


class User(TypedDict):
    name: str
    age: int


client = instructor.from_openai(OpenAI())

print(
    client.chat.completions.create(
        model="gpt-3.5-turbo",
        response_model=User,
        messages=[
            {
                "role": "user",
                "content": "Timothy is a man from New York who is turning 32 this year",
            }
        ],
    )
)

"""
name='Timothy' age=32
"""
```

ADR-007:
@ivanleomk yes, thank you!

ADR-007:
Oh. That PR is not yet merged, so I should leave this issue open

ADR-007:
I would really like to have this implemented. It is currently the only thing that blocks me from using this library :(

jxnl:
we're close!

ivanleomk:
@ADR-007 PR #758  has been merged so going to close this issue

PR #758, opened 2024-06-13
feat: add support for typeddicts

ivanleomk:
This addresses #627 and adds support for typed dicts in Instructor. Users can now use the following syntax

```python
from typing_extensions import TypedDict
from openai import OpenAI
import instructor


class User(TypedDict):
    name: str
    age: int


print(
    client.chat.completions.create(
        model="gpt-3.5-turbo",
        response_model=User,
        messages=[
            {
                "role": "user",
                "content": "Timothy is a man from New York who is turning 32 this year",
            }
        ],
    )
)

"""
name='Timothy' age=32
"""
```

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit e7341fd0d4489e6fc25834bf81071b023f248fd3  | 
|--------|--------|

### Summary:
This PR adds support for TypedDict in Instructor by converting them to Pydantic models and includes comprehensive tests for these features.

**Key points**:
- Added support for `TypedDict` in `Instructor` by converting them to Pydantic models.
- Modified `tests/llm/test_openai/test_patch.py` to include tests for `TypedDict` support.
- Updated `test_typed_dict` to assert the response is an instance of `BaseModel` instead of `UserExtract`.

----


tests/llm/test_openai/test_patch.py
```diff
@@ -12,6 +13,30 @@ class UserExtract(BaseModel):
...
+    client = instructor.patch(client, mode=mode)
+    model = client.chat.completions.create(
+        model=model,
+        response_model=UserExtractTypedDict,
+        max_retries=2,
+        messages=[
+            {"role": "user", "content": "Extract jason is 25 years old"},
+        ],
+    )
+    assert isinstance(model, BaseModel), "Should be instance of a pydantic model"
```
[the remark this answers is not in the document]
ivanleomk: We can only assert that it's an instance of a BaseModel...

ivanleomk:
Small downside is that we don't get type hints anymore but that is expected since we are using the `create_model` method to take in arbitrary typed dicts

tests/test_process_response.py
```diff
@@ -0,0 +1,18 @@
...
+def test_typed_dict_conversion() -> None:
+    class User(TypedDict):  # type: ignore
+        name: str
+        age: int
+
+    _, user_tool_definition = handle_response_model(User)
+
+    class User(BaseModel):
+        name: str
+        age: int
```
ADR-007, line 15: Does it work with Required/NotRequired types? 🤔

instructor/process_response.py
```diff
@@ -193,6 +201,12 @@ def handle_response_model(
         if is_simple_type(response_model):
             response_model = ModelAdapter[response_model]
 
+        if is_typed_dict(response_model):
+            response_model: BaseModel = create_model(
```
ADR-007, line 205: Maybe it would be easier to use RootSchema?
```
response_model = RootModel[response_model]
```

ADR-007: But in this case, the value should taken from `response_model.root`, so we need to have another `if` 🤔

ADR-007: Nevermind, it doesn't work

tests/test_process_response.py
ivanleomk: Hmm currently with my approach, I assume all keys are required but I see where you're coming from. Will try to see what I can do

jxnl: is optional not supported?

jxnl (approved):
Please make sure to add this in concepts. typeddict.md or something in concepts

PR #762, opened 2024-06-15
feat: prompt docs

TwistingTwists:
This is in relation to tweet : https://x.com/jxnlco/status/1801666332343287920

One is done. more coming.


jxnl:
I've found someone to work on this already ty!

Issue #715, opened 2024-05-30
Count tokens used in retried requests due to Pydantic validation failures


Elijas:
**Is your feature request related to a problem? Please describe.**
I'm frustrated that only the last request tokens are included when retrying to extract the Pydantic model after a validation failure. This limits the effectiveness of retries, especially when using multiple LLMs like Litellm and Anthropic. It would be helpful to have a complete token count for all retried requests to better manage context and improve accuracy.

**Describe the solution you'd like**
I would like the retry logic to count and include tokens from all previous requests, not just the last one. This will ensure a comprehensive context for retries, improving the effectiveness of the model extraction process and better handling validation failures.

**Describe alternatives you've considered**
- Using the max_retries Parameter: This currently only includes the last request tokens, which doesn't provide a full context.
- Custom Retry Logic with Tenacity: This approach could work but requires manually integrating token counts from all previous requests, which is complex and not straightforward without built-in support.

**Is your feature request related to a problem? Please describe.**
I'm frustrated that only the last request tokens are included when retrying to extract the Pydantic model after a validation failure. This limits the effectiveness of retries, especially when using multiple LLMs like Litellm and Anthropic. It would be helpful to have a complete token count for all retried requests to better manage context and improve accuracy.

**Additional context**
Including the token count from all retried requests will help maintain full context, making the retry process more reliable and accurate. This enhancement would be particularly beneficial for users leveraging multiple LLMs, ensuring comprehensive error handling and model extraction.

Thank you for considering this feature request. It would greatly improve the utility and reliability of the Instructor library.

---

I'm using `Instructor -> LiteLLM (library, not proxy) -> Anthropic Claude` stack

jxnl:
Good call. happy to take a PR; otherwise, most big use cases I've seen with the instructor already lean on third-party systems to track the token usage.

Elijas:
> Good call. happy to take a PR; otherwise, most big use cases I've seen with the instructor already lean on third-party systems to track the token usage.

third party systems:
litellm proxy? Or youve seen something else

jxnl:
helicone, langsmith, logfire, parea,

ionflow:
I need this as well. Will look at implementing something.

cpfiffer:
+1 on this one. It seems like I'd need to write custom hooks to track each retry?

cpfiffer:
For anyone who stumbles on this, I wrote a hook to track this:

```python
from rich import print

client.clear()

class UsageTracker:
[...]

Issue #667, opened 2024-05-14
Support Usage Tokens Output in Claude API


exa256:
**Is your feature request related to a problem? Please describe.**
Currently, reporting usage dictionary from OpenAI API is supported as seen in this document and usage dictionary. https://python.useinstructor.com/concepts/usage/?h=token+usage

However, Claude API patch does not have this functionality, even though usage is available from a successful 200 response from Anthropic's server:
200 Response from https://docs.anthropic.com/en/api/messages
```
{
  "content": [
    {
      "text": "Hi! My name is Claude.",
      "type": "text"
    }
  ],
  "id": "msg_013Zva2CMHLNnXjNJJKqJ2EF",
  "model": "claude-3-opus-20240229",
  "role": "assistant",
  "stop_reason": "end_turn",
  "stop_sequence": null,
  "type": "message",
  "usage": {
    "input_tokens": 10,
    "output_tokens": 25
  }
}
```

**Describe the solution you'd like**
Instructor should patch Claude's API and surface the `usage` dictionary as part of the output in the second tuple like so:
```
structure_output, completion = client.chat.completions.create_with_completion(...)
completion.usage # should returns usage, consists of input and output tokens
```

Elijas:
Use Anthropic Claude through LiteLLM, the usage and cost gets reported

```
import instructor
from litellm import completion
from litellm import completion, completion_cost, cost_per_token
from pydantic import BaseModel


class User(BaseModel):
    name: str
    age: int


client = instructor.from_litellm(completion)

resp, completion = client.chat.completions.create_with_completion(
    model="claude-3-opus-20240229",
    max_tokens=1024,
    messages=[
        {
            "role": "user",
            "content": "Extract Jason is 25 years old.",
        }
    ],
    response_model=User,
)

assert isinstance(resp, User)
assert resp.name == "Jason"
assert resp.age == 25

usage = completion.usage
input_tokens = usage.prompt_tokens
output_tokens = usage.completion_tokens
total_tokens = usage.total_tokens
    input_cost_usd, output_cost_usd = cost_per_token(model, prompt_tokens=input_tokens, completion_tokens=output_tokens)
    completion_cost_usd = completion_cost(completion_response=raw_result)
```

ssonal:
> **Describe the solution you'd like**
> Instructor should patch Claude's API and surface the `usage` dictionary as part of the output in the second tuple like so:

`structured_output._raw_response.usage` works but doesn't take retries into account. 

[...]

PR #763, opened 2024-06-15
feat: bugfix accumulating token usage

ionflow:
In addition to now accumulating tokens for LiteLLM client calls, this fixes a couple issues that I discovered implementing this fix. 

Issue: https://github.com/jxnl/instructor/issues/715

1. InstructorRetryExceptions were not being thrown from retry.py. Now you can capture these exceptions and do something with total_usage data (e.g. calculate the cost of all failed completions)
2. InstructorRetryExceptions had `total_usage` as an `int` type and would never be able to accept `total_usage`. Now it accepts `Usage` (from anthropic.types) and `CompletionUsage` (used by both OpenAI and LiteLLM).

Also can confirm that the raw response from the anthropic client (`self.client.chat.completions.create_with_completion`) is now accumulating tokens, reported in this issue: https://github.com/jxnl/instructor/issues/667

Here is some output to demonstrate:
```
10:47:42.645 POST /generate/questionnaire
10:47:42.656   FastAPI arguments [fastapi]
model in generate_questionnaire_service: Models.CLAUDE_3_HAIKU
10:47:42.687 Pydantic CompletionUsage validate_python
10:47:42.688 Pydantic Usage validate_python
10:47:42.695 Message with 'claude-3-haiku-20240307' [LLM]
instructor accumulated usage before: Usage(input_tokens=0, output_tokens=0)
trying a completion...
accumulated usage after: Usage(input_tokens=3737, output_tokens=428)
10:47:46.166 Pydantic Questionnaire validate_json
10:47:46.179 Retry failed after 1 attempts with exception: 1 validation error for Questionnaire
personal_assistant
  ...ormation visit https://errors.pydantic.dev/2.7/v/literal_error
10:47:47.190 Retrying... Attempt number: 2
10:47:47.195 Message with 'claude-3-haiku-20240307' [LLM]
instructor accumulated usage before: Usage(input_tokens=3737, output_tokens=428)
trying a completion...
accumulated usage after: Usage(input_tokens=8064, output_tokens=872)
10:47:50.164 Pydantic Questionnaire validate_json
instructor response total usage: Usage(input_tokens=8064, output_tokens=872)
10:47:50.167 $0.003106
```


instructor/utils.py
```diff
@@ -121,10 +122,12 @@ async def extract_json_from_stream_async(
 
 def update_total_usage(
     response: T_Model,
-    total_usage: OpenAIUsage | AnthropicUsage,
+    total_usage: CompletionUsage | AnthropicUsage,
 ) -> T_Model | ChatCompletion:
     response_usage = getattr(response, "usage", None)
-    if isinstance(response_usage, OpenAIUsage) and isinstance(total_usage, OpenAIUsage):
+    if isinstance(response_usage, (CompletionUsage, LiteLLMUsage)) and isinstance(
```
jxnl, line 128: this will make litellm a dependaence, is there a prototype or inferface we can use isntead

instructor/exceptions.py
```diff
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 from typing import Any
+from anthropic.types import Usage
```
jxnl, line 4: this will break imports for those who did not instal anthropic

ionflow:
@jxnl i'm finally getting back to this. any guidance on where I should define these types? we already have a dependency on openai types for CompletionUsage, but if we don't want any more dependencies then we'll need usage types for [Anthropic](https://docs.anthropic.com/en/api/messages), [Google](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/get-token-count) and [Cohere](https://docs.cohere.com/docs/chat-api#response-structure) to support the different structures.

ionflow:
@jxnl after some more explorations and thinking about how I would want this to work for retries I've come up with a `UnifiedUsage` class and this additional `create_with_completion` argument `with_usage` to support backwards compatibility. Basically we need some universal way to track usage during retries. This unified usage class allows us to support different usage reporting structures.

Please let me know what you think about this approach. I don't want to overcomplicate it, but couldn't think of a more simple way to move forward with full support of all clients.

If you like this I will add tests and address any other feedback you might have.

instructor/usage.py
```diff
@@ -0,0 +1,70 @@
...
+        self.input_tokens += other.input_tokens
+        self.output_tokens += other.output_tokens
+
+
+class Response(Protocol):
+    usage: Any
+    unified_usage: UnifiedUsage
+
+
+def convert_to_unified_usage(usage: Any) -> UnifiedUsage:
```
jxnl, line 24: this is great!

jxnl:
happy with this! can you just make sure to update the docs for patch and for usage?

ionflow:
@jxnl I think this is ready.

jxnl (approved):

jxnl (changes requested):
it looks like the types are not correct overloaded 

Screenshot 2024-07-02 at 10 18 55 PM
Screenshot 2024-07-02 at 10 18 48 PM

I can take a stab but look at the `from_openai` to see how overload can be used

matt-bernstein:
Hi, is this PR still in a mergeable state? I need this fix as well and would take a shot at getting it merged if the codebase hasn't diverged too far.

ivanleomk:
@matt-bernstein I haven't taken a look at this for a while sorry, had a whole bunch of other issues to fix. What is the specific use case you're looking for the accumulating of token usage?

PR #764, opened 2024-06-16
feat: prompt engineering cookbooks

shreya-51:


mkdocs.yml
```diff
@@ -203,6 +203,13 @@ nav:
     - Validation: 'tutorials/4-validation.ipynb'
   - Jobs Board (External):
     - Jobs: 'jobs.md'
+  - Prompt engineering:
```
shreya-51, line 206: @jxnl would you prefer having these categories as the pages and then prompting techniques inside the table of content of each page? or pull them all out and make a page per prompting technique?

jxnl: I dont know what the optimal SEO route is

jxnl: >SEO is another crucial deciding factor for single-page vs. multi-page web design. A single-page design is adequate if you only need to target one primary keyword, but a multi-page design is the better option for a diversified SEO strategy.Sep 8, 2022

suggests that its better to have single pages, so say 'step back prompting' on google will go to instructor page. vs 

'prompting tips'

jxnl: This also makes sense since i unitest the docs, we can make sure each page has unit tested examples

shreya-51: okay cool, will adjust

jxnl (approved):

PR #801, opened 2024-07-02
Fix: Resolve create_model() TypeError in Partial class

slavakurilyak:
This pull request addresses an issue with the `create_model()` function in the `Partial` and `PartialBase` classes. The error was encountered while following the streaming partial responses example in the [Instructor documentation](https://python.useinstructor.com/concepts/partial/).

### Changes Made:

- Added missing 'model_name' argument to `create_model()` calls in both `PartialBase` and `Partial` classes
- Refactored model name creation for better readability and consistency
- Fixed errors in `PartialBase` and `Partial` classes that were causing TypeError exceptions

### Problem Description:

When trying to use the `create_partial` functionality as described in the documentation, the following error was encountered:

```python
TypeError: create_model() missing 1 required positional argument: 'model_name'
```

This error prevented the proper functioning of the partial streaming feature, which is crucial for providing incremental snapshots of the response model state.

### Impact:

This fix ensures that the partial streaming functionality works as expected, allowing users to:

- Create dynamic classes that treat all original model fields as Optional
- Stream partial responses for immediate use in contexts like UI rendering
- Iterate over incremental updates of the response model

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit d6c5e085a0862466f2741baeae5336199174f5f2  | 
|--------|--------|

### Summary:
Fixes `TypeError` in `Partial` and `PartialBase` by adding missing `model_name` argument to `create_model()` calls.

**Key points**:
- Fixes `TypeError` in `Partial` and `PartialBase` classes by adding missing `model_name` argument to `create_model()` calls.
- Refactors model name creation for readability and consistency.
- Ensures partial streaming functionality works as expected.
- Changes made in `instructor/dsl/partial.py` file.

----

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -83,12 +83,14 @@ class PartialBase(Generic[T_Model]):
             cls, BaseModel
         ), f"{cls.__name__} must be a subclass of BaseModel"
 
+        model_name = (
+            cls.__name__
+            if cls.__name__.startswith("Partial")
+            else f"Partial{cls.__name__}"
+        )
+
         return create_model(
-            __model_name=(
-                cls.__name__
-                if cls.__name__.startswith("Partial")
-                else f"Partial{cls.__name__}"
-            ),
+            model_name,
             __base__=cls,
             __module__=cls.__module__,
             **{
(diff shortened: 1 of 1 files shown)


jxnl:
do we know why this happened? is this a pydantic version or a python version gap?

slavakurilyak:
Looks like a pydantic issue with some traces [here](https://github.com/pydantic/pydantic/commit/eeea647f0b2a1107a923d4c0e3369344c985d38a)

The model name was changed `model_name` / `__model_name`

v1 code [here](https://github.com/pydantic/pydantic/blob/main/pydantic/v1/main.py)
v2 code [here](https://github.com/pydantic/pydantic/blob/main/pydantic/main.py)

jxnl:
ok so it looks like with this update we should also pin a new version of pydantic right?

jxnl:
@ivanleomk 

can you increment version and pin to pydantic > 2.8.0 and release

vikyw89:
this bug still exists on my end

largomst:
I can also reproduce this issue

vikyw89:
sent  PR to fix this
https://github.com/jxnl/instructor/pull/844

ivanleomk:
Just tested against instructor 1.3.7 and this doesn't seem to happen so going to close #844 along the way

PR #856, opened 2024-07-18
New Async Validator

ivanleomk:
Adding a new async validator

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit d25f55766c96089656708740cb856c58353ae147  | 
|--------|--------|

### Summary:
Introduced asynchronous validation for Pydantic models with new decorators, context handling, methods, tests, and documentation updates.

**Key points**:
- Added `async_field_validator` and `async_model_validator` decorators in `instructor/validators.py` for async validation.
- Introduced `AsyncValidationContext` and `AsyncValidationError` in `instructor/validators.py`.
- Updated `OpenAISchema` in `instructor/function_calls.py` with methods: `get_async_validators`, `get_model_coroutines`, `model_async_validate`, `execute_field_validator`, and `execute_model_validator`.
- Modified `process_response_async` in `instructor/process_response.py` to handle async validation.
- Updated `retry_async` in `instructor/retry.py` to catch `AsyncValidationError`.
- Added tests in `tests/llm/test_openai/test_async_decorator.py` for async validator functionality.
- Updated documentation in `docs/concepts/reask_validation.md` to include examples of async validation.

----


ivanleomk:
@jonastemplestein I've implemented a version of the async validator, would love to hear your thoughts on it.

See Documentation here : https://1586cbb1.instructor.pages.dev/concepts/reask_validation/#async-validation

instructor/decorators.py
```diff
@@ -0,0 +1,71 @@
+from typing import Callable, Any, TypeVar
```
jxnl, line 1: lets rename this to validators 
then also bring it to the main namespace in __init__.py

docs/concepts/reask_validation.md
```diff
@@ -193,6 +193,253 @@ except ValidationError as e:
...
+
+This is especially useful if you are making multiple api calls in parallel. This implementation is heavily inspired by [pydantic-async-validation](https://github.com/team23/pydantic-async-validation/tree/64e4190494141c881ade30963390f91c3286dafc).
+
+### Field Validators
+
+To use our Async Validators, you'll need to import in the `OpenAISchema` class that we expose in instructor. Here's a quick example of how we can validate multiple fields at the same time
+
+```python hl_lines="12"
+from instructor.decorators import async_field_validator
+from instructor.function_calls import OpenAISchema
```
jxnl, line 208: why not base model? we should try to not expose this externally

jxnl: would almost rather 

`from instructor import BaseModel`

jxnl: or from `instructor import AsyncBaseModel`

ivanleomk: Managed to get this working with the default pydantic base model! :)

instructor/decorators.py
ivanleomk: Renamed and added to the main namespace

ivanleomk:
Implemented the warning when a user uses the synchronous client @jonastemplestein. I'll need to think a bit about how to implement a field-level validator which re-runs when a specific validator returns a changed value.

Will need to change a significant amount of code to support that :<

docs/concepts/reask_validation.md
```diff
@@ -193,6 +193,255 @@ except ValidationError as e:
...
+1.  You can set the fields to run the same validator on. We don't support wildcard characters (eg. '\*' ) at this moment but might in the future.
+
+### Model Validators
+
+We can also use a model validator to validate a combination of fields. In this case, we want to make sure that every field is in caps. This helps us to prevent errors.
+
+```python hl_lines="11"
+from pydantic import BaseModel
+import asyncio
+from instructor import from_openai,async_model_validator
```
jxnl, line 254:
```suggestion
from instructor import from_openai, async_model_validator
```

jxnl:
i think this just needs ruff on the docs. 

`pytest --update-examples`

ivanleomk:
@jxnl have updated the repo and ran ruff on it

jxnl (approved):

instructor/function_calls.py
```diff
@@ -295,14 +434,39 @@ def parse_json(
...
+def openai_schema_helper(cls: T) -> T:
+    origin = get_origin(cls)
+
+    if origin is list:
+        return list[openai_schema_helper(cls.__args__[0])]
+
+    if origin is Literal:
+        return cls
+
+    if issubclass(cls, (str, int, bool, float, bytes)):
```
michael-quinlan, line 446: I think this code doesnt handle UnionTypes, for example a BaseModel with something like
```
  name: str | None
```
At this point if you call ```issubclass(cls, (str, int, bool, float, bytes))``` on. UnionType you get ```TypeError: issubclass() arg 1 must be a class```.


Our code started crashing and I bisected it to this PR. In particular if I revert openai_schema to 
```
shema = wraps(cls, updated=())(
        create_model(
            cls.__name__ if hasattr(cls, "__name__") else str(cls),
            __base__=(cls, OpenAISchema),
        )
    )
    return cast(OpenAISchema, shema) 
```
and bypass this whole new method then it works again.

FYI we use Hakiu and i haven't validated if this code works on OpenAi, however are they any tests that cover UnionTypes?

I can also see if I can come up with a fix, but those who know the entire code base better might be faster.

michael-quinlan: So i think if you add this
```
    if isinstance(cls, UnionType):
        for arg in get_args(cls):
            return openai_schema_helper(arg)
```
on line 444 (i.e above ```if issubclass(cls, (str, int, bool, float, bytes)): ...'```) then it works.

You will also need to 
```
from typing import Annotated, Any, Optional, TypeVar, cast, get_origin, Literal, get_args
from types import UnionType
```

lazyhope: Not only that, it doesn't work for types like dict[str, str] either. The current implementation really needs to be carefully reviewed and fixed. @jxnl @ivanleomk

PR #857, opened 2024-07-18
Call `model.validate_async()` during async response processing (if it exists)

jonastemplestein:
This PR makes instructor call `model.validate_async()` during response processing, if such a method exists.

**Context:**

We often have multiple LLM based validators on a single response model, so we want to run the validators asynchronously in parallel.

This is not possible using pydantic's built-in validation mechanism, because this runs during a model's `__init__`, which is never asynchronous.

In our model base class, we've implemented the following API for defining async validators:

```python
class Model(HexModel):
    foo: str
    bar: str

    @async_field_validator("foo", "bar")
    @classmethod
    async def validate_foobar(cls, v: str) -> str:
        return v + "_123"

    @async_model_validator
    async def validate_model(self):
        if self.foo != self.bar:
            raise ValueError("foo must be equal to bar")

m = Model(foo="abc", bar="def")
await m.validate_async() # raises a ValidationError just like pydantic's synchronous validators
```

To benefit from instructors retry loop, we need instructor to call `validate_async()` for us.

I'm opening this as a PR, because a) we believe this functionality would be useful for other users who make heavy use of `llm_validator` and b) we don't want to maintain an instructor fork

We're happy to change the implementation to be more to your liking, too. 

Happy to share the implementation of the validation logic itself, too, if useful. 

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit bd6e7720e1a41c127882253703e5ce8e26008256  | 
|--------|--------|

### Summary:
Add hook to call `model.validate_async()` if it exists during response processing in `instructor/process_response.py`.

**Key points**:
- **Feature Added**: Call `model.validate_async()` if it exists in `instructor/process_response.py`.
- **File Affected**: `instructor/process_response.py`.
- **Function Modified**: `process_response_async`.
- **Behavior**: If `model` has a `validate_async` method, it is called asynchronously during response processing.
- **Use Case**: Enables asynchronous validation for models, useful for running multiple LLM-based validators in parallel.

----

### DIFF
--- a/instructor/process_response.py
+++ b/instructor/process_response.py
@@ -81,6 +81,9 @@ async def process_response_async(
         mode=mode,
     )
 
+    if hasattr(model, "validate_async") and callable(model.validate_async):
+        await model.validate_async()
+
     # ? This really hints at the fact that we need a better way of
     # ? attaching usage data and the raw response to the model we return.
     if isinstance(model, IterableBase):
--- a/instructor/retry.py
+++ b/instructor/retry.py
@@ -246,7 +246,7 @@ async def retry_async(
                         mode=mode,
                     )
                 except (ValidationError, JSONDecodeError) as e:
-                    logger.debug(f"Error response: {response}", e)
+                    logger.debug(f"Error response: {response}")
                     kwargs["messages"].extend(reask_messages(response, mode, e))
                     if mode in {Mode.ANTHROPIC_TOOLS, Mode.ANTHROPIC_JSON}:
                         kwargs["messages"] = merge_consecutive_messages(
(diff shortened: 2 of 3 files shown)


jonastemplestein:
@ivanleomk @jxnl Looks like wires crossed with your effort in #856

jonastemplestein:
Yea, that's easier for us as we know all our models derive from the same base class

We've also implemented `@async_model_validator` in our implementation

I wonder if it would make sense for you to implement the async validators in a separate class that can be mixed into any `pydantic.BaseModel`. This would make the implementation more straightforward and re-usable, but the cost is that anyone using async Validators would have to use that mixin class

Here's how we've implemented the async validators on our model base class: https://gist.github.com/jonastemplestein/fae719b82cc516a1bf6701ce2ee2a242

(My cofounder @obeattie built this, not me)

PR #872, opened 2024-07-23
Fix dedent issue in process_response.py

Mazyod:
Hello,

I noticed that the prompts being generated have weird indentations. This is due to the `dedent` function being applied with f-string interpolation, which causes it to effectively do nothing since the JSON schema being dumped doesn't have indentations in new lines.

I have created this draft PR using Copilot Workspace, but happy to polish it and make any necessary changes if this if of interest to the maintainers.

---

Update `handle_response_model` function to fix dedent issue

* Apply `dedent` function only on the literal string and concatenate it with the `json.dumps` result in the `handle_response_model` function.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/jxnl/instructor?shareId=b27788dc-42e0-450f-b3f2-d7258d5073f3).

### DIFF
--- a/instructor/process_response.py
+++ b/instructor/process_response.py
@@ -264,15 +264,19 @@ def handle_response_model(
         elif mode in {Mode.JSON, Mode.MD_JSON, Mode.JSON_SCHEMA}:
             # If its a JSON Mode we need to massage the prompt a bit
             # in order to get the response we want in a json format
-            message = dedent(
-                f"""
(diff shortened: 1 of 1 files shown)


ivanleomk:
@jxnl i took a look at this and the fix does work. I'm just wondering if (1) We should merge this in or if (2) We should look towards standardizing prompts with just a Jinja environment that applies an dendent automatically

PR #907, opened 2024-08-06
add(docs): add llm input debug log

### LINKED ISSUE #906: [Feature]Add log for llm input/prompt text data
Sometimes, we want to know what's the prompt/input which is sent to llm is exactly. For example, as a new user of pydantic or instructor, we don't know how these pkgs work under the hood, so it's will be helpful to check the prompt itself.

We could add a debug message simply.

Maybe a info level message?

**Additional context**
I'd like to create a PR to do this.

shenxiangzhuang:
Close #906 

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit f68392c19ad1708dd85203131d5b1d0726629cc8  | 
|--------|--------|

### Summary:
Added debug logging for input arguments in `retry_sync` and `retry_async` functions in `instructor/retry.py`.

**Key points**:
- Added debug logging for input arguments in `instructor/retry.py`.
- Updated `retry_sync` to log `args` and `kwargs` before function execution.
- Updated `retry_async` to log `args` and `kwargs` before function execution.
- Helps in tracing inputs passed to the LLM during retries.

----

### DIFF
--- a/instructor/retry.py
+++ b/instructor/retry.py
@@ -158,6 +158,7 @@ def retry_sync(
         for attempt in max_retries:
             with attempt:
                 try:
+                    logger.debug(f"sync llm input: {args=}, {kwargs=}")
                     response = func(*args, **kwargs)
                     stream = kwargs.get("stream", False)
                     response = update_total_usage(response, total_usage)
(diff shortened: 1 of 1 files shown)


ivanleomk:
Hmm have you tried using logfire or langfuse? I find that the logging is pretty good there and should work out of the box since we're using the base clients itself.

shenxiangzhuang:
> Hmm have you tried using logfire or langfuse? I find that the logging is pretty good there and should work out of the box since we're using the base clients itself.

Well, thanks for your timely reply and I haven't test these observability tools with instructor. IMHO, it will be very convenient if the instrctor library itself could provide some basic log. For example, in some simple project, the basic build-in logging library is enough. So I think these logs should be helpful in the situations where we don't(or can't) use some  observability tools.

jxnl:
openai has debug logs that call log everything.
that said i dont want instructor to be logging everything to debug, not all input data should be logged so I'll have to think harder about this

Issue #880, opened 2024-07-24
Optional[List] as field type crashes with streaming


ahuang11:
- [x] This is actually a bug report.
- [ ] I am not getting good LLM Results
- [ ] I have tried asking for help in the community on discord or discussions and have not received a response.
- [ ] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

- [x] gpt-3.5-turbo
- [ ] gpt-4-turbo
- [ ] gpt-4
- [ ] Other (please specify)

**Describe the bug**

In `openai_schema_helper`, `issubclass(cls)` crashes when `list[str] | None`

**To Reproduce**
Steps to reproduce the behavior, including code snippets of the model and the input data and openai response.

```python
import instructor
from typing import Optional
from instructor import Partial
from pydantic import BaseModel
from openai import AsyncOpenAI



# Define your desired output structure
class UserInfo(BaseModel):
    name: Optional[list[str]]
    age: int


# Patch the OpenAI client
client = instructor.from_openai(AsyncOpenAI())

# Extract structured data from natural language
user_info = await client.chat.completions.create(
    model="gpt-3.5-turbo",
    response_model=Partial[UserInfo],
    messages=[{"role": "user", "content": "John Doe is 30 years old."}],
    stream=True
)

async for chunk in user_info:
    print(chunk)
```

**Expected behavior**

Works like without streaming.

**Screenshots**

michael-quinlan:
I think this is the same as https://github.com/jxnl/instructor/issues/878 but this issue has a nice small reproduction step.

ahuang11:
Thanks for letting me know. I'll close this as duplicate.

michael-quinlan:
I think it's fine to also leave it open, this is my first PR in this repo, so I don't know the protocols they follow. I'm just trying to help get a fix out as more people will hit this now it's in the 'latest version'

Btw I'm more than happy for the real fix not being my PR, that's just my attempt, so others who have better solutions should go ahead.

Issue #878, opened 2024-07-24
issubclass() arg 1 must be a class


ShawnToron:
- [x] This is actually a bug report.
- [ ] I am not getting good LLM Results
- [ ] I have tried asking for help in the community on discord or discussions and have not received a response.
- [ ] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

- [ ] gpt-4
- [ ] Other (please specify)

**Describe the bug**
I have a Flask App and using Instructor with GPT4o.  It works perfectly on my windows machine.  But I have a new windows machine with python, etc. installed but when I run the same code it's all working except the functions with the Instructor library where i get the following error message:

'Error fetching recommendations: issubclass() arg 1 must be a class'

**To Reproduce**
Steps to reproduce the behavior, including code snippets of the model and the input data and openai response.

**Expected behavior**

**Screenshots**

michael-quinlan:
I beleive this is what i fixed in 

https://github.com/jxnl/instructor/pull/877

Do you pydantic models include optional values i.e str | None ?

ShawnToron:
Hi Michael:

Yes my model includes Optional, str, none.

e.g.
class InvestingProfile(BaseModel):
    """
    This model includes detailed information about a client's investing
profile.
    """
    investor_type: Optional[str] = Field(None, description="Investor's risk
profile categorized high, medium, or low based on their risk tolerance.")
    assetclass_preferences: Optional[str] = Field(None,
description="Comma-separated
list of preferred asset classes, such as stocks, bonds, or real estate.")

Can you please guide me on how I should I apply the fix?

Thank you so much!

michael-quinlan:
So hopefully the someone will review and merge the changes or make the appropriate changes themselves.

For you i think the best thing to do is install a version prior the bug being introduced, Version 1.3.5 should be good.

ShawnToron:
It works now, Michael.  Thank SO much.  Really appreciate it.

ivanleomk:
Sorry about this, I pushed up a new version that breaks some of these integrations. But hopefully most of these will be ironed out soon. Will take a look at @michael-quinlan's PR in the week when I get the time

ShawnToron:
No worries, Ivan.  All good and thanks for to you all and esp Jason for the
awesome library!

ivanleomk:
@ShawnToron  I would recommend 1.3.4 to be safe before trying 1.3.5

ShravanSunder:
[...]

PR #908, opened 2024-08-06
Gracefully handle unsupported types in OpenAI function call schema

bencrouse:
## Related

https://github.com/jxnl/instructor/issues/878
https://github.com/jxnl/instructor/issues/880

## Description

This is an improvement related to https://github.com/jxnl/instructor/issues/878. With this change, when using `str | None` with OpenAI, instead of a `TypeError: issubclass() arg 1 must be a class` you get a `ValueError: Unsupported Class of str | None` which can lead you to a direct resolution more quickly.

Note that it seems like [the desired fix was merged](https://github.com/jxnl/instructor/issues/880), but I figured this was quick enough to open that it would at least help others understand what's really going on. If not, feel free to close as irrelevant! Thanks.

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 335dacd3f1167b202310f22d1e7d7f7212836c3f  | 
|--------|--------|

### Summary:
Improved error handling for unsupported types in `openai_schema_helper` function in `instructor/function_calls.py`.

**Key points**:
- Updated `openai_schema_helper` in `instructor/function_calls.py` to handle unsupported types more gracefully.
- Replaced `TypeError` with `ValueError` for unsupported types.
- Improved error message to `ValueError: Unsupported Class of {cls}`.
- Specifically handles `str | None` type scenario.

----

### DIFF
--- a/instructor/function_calls.py
+++ b/instructor/function_calls.py
@@ -472,24 +472,25 @@ def openai_schema_helper(cls: T) -> T:
     if origin is Union:
         return Union[tuple(openai_schema_helper(arg) for arg in cls.__args__)]
 
-    if issubclass(cls, (str, int, bool, float, bytes, Enum)):
-        return cls
+    if isinstance(cls, type):
+        if issubclass(cls, (str, int, bool, float, bytes, Enum)):
+            return cls
 
-    if isinstance(cls, type) and issubclass(cls, BaseModel):
-        new_types = {}
-        for field_name, field_info in cls.model_fields.items():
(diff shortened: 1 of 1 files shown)


ivanleomk:
@bencrouse thanks for the PR! I think we're going to just revert this recursive parsing of the OpenAI Schema in the next release which should fix this error. Just moving a bit slower and adding more testing before we do another release

ivanleomk:
Going to close this PR since we won't be doing more work on the openai_helper function

PR #938, opened 2024-08-20
Adding support for structured outputs

ivanleomk:
This seems to be a much easier integration

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit ea1509bd96e3bdaef993972d8b0cca938e47c9a7  | 
|--------|--------|

### Summary:
This PR adds `Mode.STRUCTURED_OUTPUTS` and `Mode.TOOLS_STRICT`, updates JSON handling, and modifies tests and dependencies.

**Key points**:
- Add `Mode.STRUCTURED_OUTPUTS` and `Mode.TOOLS_STRICT` in `instructor/mode.py`.
- Update `handle_response_model` for both modes.
- Modify JSON extraction in `instructor/dsl/iterable.py` and `instructor/dsl/partial.py`.
- Adjust `from_response` in `instructor/function_calls.py`.
- Update `instructor/client.py` for both modes.
- Skip tests for both modes in `tests/llm/test_openai/test_validators.py`.
- Update OpenAI dependency from `^1.40.0` to `^1.1.0`.

----


ivanleomk:
A bit skeptical if this works but https://openai.com/index/introducing-structured-outputs-in-the-api/ we can see that the easiest way to do this is really

CleanShot 2024-08-20 at 09 24 19@2x


Which we can just hook into using our `process_response_model`  function

ivanleomk:
Most changes are due to the pyproject.toml being regenerated

instructor/dsl/partial.py
```diff
@@ -170,7 +170,7 @@ def extract_json(
                     elif mode in {Mode.JSON, Mode.MD_JSON, Mode.JSON_SCHEMA}:
                         if json_chunk := chunk.choices[0].delta.content:
                             yield json_chunk
-                    elif mode == Mode.TOOLS:
+                    elif mode == Mode.TOOLS or mode == Mode.STRUCTURED_OUTPUTS:
```
jxnl, line 173: use a set like `elif mode in {Mode.JSON, Mode.MD_JSON, Mode.JSON_SCHEMA}`

instructor/function_calls.py
```diff
@@ -264,7 +264,7 @@ def from_response(
             Mode.warn_mode_functions_deprecation()
             return cls.parse_functions(completion, validation_context, strict)
 
-        if mode in {Mode.TOOLS, Mode.MISTRAL_TOOLS}:
+        if mode in {Mode.TOOLS, Mode.MISTRAL_TOOLS, Mode.STRUCTURED_OUTPUTS}:
```
jxnl, line 267: why does structured outputs using parse tools?

instructor/mode.py
```diff
@@ -19,6 +19,7 @@ class Mode(enum.Enum):
     VERTEXAI_JSON = "vertexai_json"
     GEMINI_JSON = "gemini_json"
     COHERE_JSON_SCHEMA = "json_object"
+    STRUCTURED_OUTPUTS = "structured_output"
```
jxnl, line 22: one is plural and one is singular

instructor/process_response.py
```diff
@@ -249,6 +250,15 @@ def handle_response_model(
             Mode.warn_mode_functions_deprecation()
             new_kwargs["functions"] = [response_model.openai_schema]
             new_kwargs["function_call"] = {"name": response_model.openai_schema["name"]}
+        elif mode in {Mode.STRUCTURED_OUTPUTS}:
```
jxnl, line 253: why not call it tools_strict 

since it looks like we're not using the parse method?

jxnl (changes requested):
this is strict tool calling, not structured outputs as defined by the parse method

ivanleomk:
> this is strict tool calling, not structured outputs as defined by the parse method 

Hmm I was following the instructions in their announcement article here which toggles it by default when you use strict = True when calling tools

 [openai.com/index/introducing-structured-outputs-in-the-api](https://openai.com/index/introducing-structured-outputs-in-the-api/) 

See the Docs here too : https://platform.openai.com/docs/guides/function-calling

I think this is a better way to do structured output since it allows us to hook into all of the existing functionality that we've already built instead of rebuilding new functionality around the new `.parse` and `.stream` method

instructor/process_response.py
ivanleomk: That's a good point, let me rename the method to TOOLS_STRICT then

instructor/function_calls.py
ivanleomk: They've enabled structured output support for function calling if we use their strict:true https://platform.openai.com/docs/guides/function-calling. That's why we use the pydantic helper in order to enerate the relevant schema.

In this case, I tried using the `.parse` method and found it difficult to figure out how to do validation with + support the partials and iterables that we had already supported. The Function Calling approach was a lot more simple and easy to understand

instructor/dsl/partial.py
ivanleomk: Sure, that makes it more consistent. Is there a reason why we should prefer a set over say a normal if-else? Is that just so that it's easier to integrate/make changes down the line

instructor/process_response.py
```diff
@@ -249,6 +250,15 @@ def handle_response_model(
             Mode.warn_mode_functions_deprecation()
             new_kwargs["functions"] = [response_model.openai_schema]
             new_kwargs["function_call"] = {"name": response_model.openai_schema["name"]}
+        elif mode == Mode.TOOLS_STRICT:
+            response_model_schema = pydantic_function_tool(response_model)
```
ivanleomk, line 254: This functionality here uses the OpenAI Function tool. we set the strict to True to configure the Structured Outputs mode

```
from openai import pydantic_function_tool


class User(BaseModel):
    name: str
    age: int


class Users(BaseModel):
    users: list[User]


print(pydantic_function_tool(Users))

# {
#     "type": "function",
#     "function": {
#         "name": "Users",
#         "strict": False,
#         "parameters": {
#             "$defs": {
#                 "User": {
#                     "properties": {
#                         "name": {"title": "Name", "type": "string"},
#                         "age": {"title": "Age", "type": "integer"},
#                     },
#                     "required": ["name", "age"],
#                     "title": "User",
#                     "type": "object",
#                     "additionalProperties": False,
#                 }
#             },
#             "properties": {
#                 "users": {
#                     "items": {"$ref": "#/$defs/User"},
#                     "title": "Users",
#                     "type": "array",
#                 }
#             },
#             "required": ["users"],
#             "title": "Users",
#             "type": "object",
#             "additionalProperties": False,
#         },
#     },
# }

```

jxnl (approved):

PR #941, opened 2024-08-21
Add LLAMA-3.1 JSON mode

kerem-essential:
Llama 3.1 has its own tool calling modes including. This PR adds support for that mode as otherwise LLAMA-3.1 models have low performance in more complicated completions.

### DIFF
--- a/instructor/client.py
+++ b/instructor/client.py
@@ -411,6 +411,7 @@ def from_openai(
             instructor.Mode.JSON,
             instructor.Mode.JSON_SCHEMA,
             instructor.Mode.MD_JSON,
+            instructor.Mode.LLAMA3_JSON,
         }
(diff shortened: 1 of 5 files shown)


ivanleomk:
@kerem-essential didn't know this was the case. Do you have a link I could read a bit more about this? Woul like to understand more about this response format before approving the PR

kerem-essential:
> @kerem-essential didn't know this was the case. Do you have a link I could read a bit more about this? Woul like to understand more about this response format before approving the PR

Absolutely! I used this page as reference: https://llama.meta.com/docs/model-cards-and-prompt-formats/llama3_1/#json-based-tool-calling
Should've posted in the original PR description but was in a hurry.

PR #948, opened 2024-08-24
Bugfix: Literal while streaming does not work

roeybc:
Trying to fix literal validation while streaming by waiting for the object to fully be received.

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -130,7 +130,7 @@ class PartialBase(Generic[T_Model]):
         for chunk in json_chunks:
             potential_object += chunk
             obj = from_json(
-                (potential_object or "{}").encode(), partial_mode="trailing-strings"
+                (potential_object or "{}").encode(), partial_mode="on"
             )
             obj = partial_model.model_validate(obj, strict=None, **kwargs)
             yield obj
@@ -144,7 +144,7 @@ class PartialBase(Generic[T_Model]):
         async for chunk in json_chunks:
             potential_object += chunk
             obj = from_json(
-                (potential_object or "{}").encode(), partial_mode="trailing-strings"
+                (potential_object or "{}").encode(), partial_mode="on"
             )
             obj = partial_model.model_validate(obj, strict=None, **kwargs)
             yield obj


mrdkucher:
Is there another solution for streaming Literals? It's nice to be able to see trailing strings in streaming responses (I just created an issue for this: https://github.com/jxnl/instructor/issues/989).

Perhaps Literals should be modified to accept any string value when streaming back partials, and validation can be done afterwards.

ivanleomk:
Hmm this really isn't something that we can change at the moment, I personally prefer the trailing strings too in streaming responses but that breaks Literal support. 

Don't think a flag that you pass in six levels down is the solution too. We might just have to wait for jitter to modify its parsing strategy to support both

mrdkucher:
> Hmm this really isn't something that we can change at the moment, I personally prefer the trailing strings too in streaming responses but that breaks Literal support.
> 
> Don't think a flag that you pass in six levels down is the solution too. We might just have to wait for jitter to modify its parsing strategy to support both

Agreed on the flag, that doesn't feel like the right solution. I'm confused about the Literal case, though. Can you provide a test case that fails when you use `"trailing-strings"`? I thought validation would just be disabled for streamed responses. I'd opt to make `instructor.Partial` accept values outside the Literals?

mrdkucher:
@ivanleomk , what do you think about this solution for `Literal`s?

https://stackoverflow.com/questions/76669927/pydantic-how-to-ignore-invalid-values-when-creating-model-instance

Since validation on many fields is disabled, it feels appropriate to do the same for Literal fields rather than having those be a special case which require validation. Further, disabling `"trailing-strings"` just for Literals seems heavy-handed. I'd propose that we simply return `None` for the Literal field in the case the model generates something which isn't an acceptable Literal value, since that's more in line with how other fields are "validated".

I've tested this approach locally, and it allows us to keep "trailing-strings" enabled. Copied from stack overflow:
```python
from collections.abc import Callable
from typing import Annotated, Any

from pydantic import BaseModel, ValidationError, WrapValidator


def invalid_to_none(v: Any, handler: Callable[[Any], Any]) -> Any:
    try:
        return handler(v)
    except ValidationError:
        return None


class Foo(BaseModel):
    age: Annotated[int | None, WrapValidator(invalid_to_none)]
    name: str | None
```

If we take this approach, we'd simply return `None` values until we get a valid Literal value, which helps when we're receiving partials of the Literal.

PR #991, opened 2024-09-13
Adding support for O1

ivanleomk:
This adds a new mode which should support the o1 model family using the openai client sdk. We'll need to upgrade the supported version to support the new `max_completion_tokens` parameter though. This allows us to do something like what we see below.

```python
from instructor import from_openai, Mode
from openai import OpenAI

client = from_openai(OpenAI(), mode=Mode.JSON_O1)


class UserDetails(BaseModel):
    name: str
    age: int


    model="o1-mini",
    messages=[{"role": "user", "content": "Jason is 10"}],
    response_model=UserDetails,
    max_completion_tokens=300, # This has to have a minimum of 200+ because of reasoning tokens
)

print(resp)
```

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit f35b04c32fe447855b1af5885bebac732b9deed6  | 
|--------|--------|

feat: add support for JSON_O1 mode with OpenAI SDK

### Summary:
Adds support for `JSON_O1` mode using the OpenAI client SDK, requiring an updated `openai` dependency.

**Key points**:
- **Behavior**:
  - Adds support for `JSON_O1` mode in `from_openai()` in `client.py`.
  - Updates `handle_response_model()` in `process_response.py` to handle `JSON_O1` mode, ensuring no system messages are included.
  - Updates `from_response()` in `function_calls.py` to parse `JSON_O1` mode using `parse_json()`.
- **Dependencies**:
  - Updates `openai` dependency to `^1.45.0` in `pyproject.toml` to support `max_completion_tokens`.
- **Enums**:
  - Adds `JSON_O1` to `Mode` enum in `mode.py`.

----

### DIFF
--- a/instructor/client.py
+++ b/instructor/client.py
@@ -426,6 +426,7 @@ def from_openai(
             instructor.Mode.PARALLEL_TOOLS,
             instructor.Mode.MD_JSON,
             instructor.Mode.TOOLS_STRICT,
+            instructor.Mode.JSON_O1,
         }
 
     if isinstance(client, openai.OpenAI):
--- a/instructor/function_calls.py
+++ b/instructor/function_calls.py
@@ -146,7 +146,7 @@ class OpenAISchema(BaseModel):
         if mode in {Mode.TOOLS, Mode.MISTRAL_TOOLS, Mode.TOOLS_STRICT}:
             return cls.parse_tools(completion, validation_context, strict)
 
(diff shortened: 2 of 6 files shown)


ivanleomk:
Important here to note that the current `o1-preview` doesn't support things like tool-calling, streaming and so this is really just supporting a simple json mode support

jxnl (approved):

PR #1056, opened 2024-10-09
fix: update gemini's `safety_settings`

alxpez:
Addressing the issue where instructor's "required" `safety_settings` are overriding user-provided or pre-existing settings in the Gemini model configuration.

These settings are applied unconditionally. The updates ensure that default settings are applied only if they represent a *more restrictive* threshold than the existing setting for a given harm category.

This preserves any pre-existing settings that have:
* A higher (less restrictive) threshold (e.g. `BLOCK_NONE` compared to `BLOCK_ONLY_HIGH`).
* Additional categories not included in the default settings (e.g. `HARM_CATEGORY_SEXUALLY_EXPLICIT`).

This change ensures that user-specified safety configurations are respected while still providing required settings levels when none are provided.

----

> [!IMPORTANT]
> Updates `update_gemini_kwargs` in `instructor/utils.py` to apply default `safety_settings` only if more restrictive than existing settings.
> 
>   - **Behavior**:
>     - Updates `update_gemini_kwargs` in `instructor/utils.py` to apply default `safety_settings` only if they are more restrictive than existing settings.
>     - Preserves user-provided settings with higher thresholds or additional categories.
>   - **Imports**:
>     - Removes `type: ignore` comment from `google.generativeai.types` import.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=jxnl%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 6fa5494b93654edc89842892d1b50e83999464c9. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/instructor/utils.py
+++ b/instructor/utils.py
@@ -345,7 +345,7 @@ def update_gemini_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
     kwargs["contents"] = transform_to_gemini_prompt(kwargs.pop("messages"))
 
     # minimize gemini safety related errors - model is highly prone to false alarms
-    from google.generativeai.types import HarmCategory, HarmBlockThreshold
+    from google.generativeai.types import HarmCategory, HarmBlockThreshold  # type: ignore
 
     minimum_safety_settings = {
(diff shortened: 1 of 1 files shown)


alxpez:
redundant

PR #1070, opened 2024-10-11
Enhance multimodality

arcaputo3:
## Summary:
I think `multimodal.py` is a huge step in the right direction for Instructor, especially as vision becomes more mainstream. Since every provider handles images uniquely, the hope is that the interface will only need to define `to_<provider>` methods, and the `Image` class can handle serialization and routing as needed.

Though experimental, I propose image autodetection to streamline vision use cases. Whether image data comes from a path, URL, or base64 encoded string, it should always be possible to automatically figure it out from a source string (bar edge cases where a user can be more explicit if needed). This allows users to specify content as a list of strings, and Instructor can handle the rest by setting `autodetect_images=True`.

## Change notes:
- Upgrade `Image` to support all valid OpenAI and Anthropic vision mime types
- Use mime type detection rather than extensions
- Add image autodetection for ease of use

### DIFF
--- a/docs/concepts/multimodal.md
+++ b/docs/concepts/multimodal.md
@@ -33,3 +33,21 @@ response = client.chat.completions.create(
 The `Image` class takes care of the necessary conversions and formatting, ensuring that your code remains clean and provider-agnostic. This flexibility is particularly valuable when you're experimenting with different models or when you need to switch providers based on specific project requirements.
 
 By leveraging Instructor's multimodal capabilities, you can focus on building your application logic without worrying about the intricacies of each provider's image handling format. This not only saves development time but also makes your code more maintainable and adaptable to future changes in AI provider APIs.
+
+Alternatively, by passing `autodetect_images=True` to `client.chat.completions.create`, you can pass file paths, URLs, or base64 encoded content directly as strings.
+
+```python
+import instructor
+import openai
+
+client = instructor.from_openai(openai.OpenAI())
+
+response = client.chat.completions.create(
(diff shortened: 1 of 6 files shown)


jxnl:
this looks great but mind cleaning up the diff

arcaputo3:
> this looks great but mind cleaning up the diff

yes, I ruffed the whole repo by mistake lol. fixed now

arcaputo3:
Ideally I'd like to be able to support this feature with Anthropic prompt caching more natively as currently this would require some gymnastics. My thought is that we could support a generic content dict such as 
```python
{"type": "image", "source": <generic_str_source>}
```
and that way users could specify the `"cache_control"` key if they like

jxnl:
oh good idea on the cache key. as long as its well documented im +1

arcaputo3:
@jxnl let me know what you think. I am not actively using Gemini, but it would be great if we could add Google support at a later date.

jxnl:
why not use functools.lru_cache looks like a lot of extra code to implement the cache

arcaputo3:
> why not use functools.lru_cache looks like a lot of extra code to implement the cache

A few reasons:

- Cannot be modified or disabled at runtime
- Cannot be accessed after the function is declared

although it is slop, it is functionally better. I tried to make `lru_cache` work similarly and it was terrible. I'm open to going back and just hard enabling caching at 128, but I know some power users might want to disable it

arcaputo3:
Okay I realized that I didn't even need to hash the cache control logic and refactored everything to be much more readable

PR #1103, opened 2024-10-21
fix: hotfix `create_with_completion` failing for `AdapterBase`, `ParallelBase`, `IterableBase` and `PartialBase`

ivanbelenky:
## Describe your changes

The change solve `create_with_completion` failing with `AdapterBase`, `ParallelBase`, `IterableBase` and `PartialBase` types. Every single `instructor.create_fn` call is wrapped with a `Creation` object. The idea for this wrapper is to be as simple as a container for both the raw response from the LM client, and the processed response by `instructor`.

```python
T = TypeVar("T", bound=Union[BaseModel, "Iterable[Any]", "Partial[Any]"])


class Creation(Generic[T]):
    raw: Any # probably should collapse to the ChatCompletion model
    processed: T

    def __init__(self, raw: Any, processed: T) -> None:
        self.raw, self.processed = raw, processed
```

Then `retry_sync` and `retry_async` return types were altered from `T_Model` to `Creation[T_Model]`.

For further traceability in instructor calls, a runtime accessible Inspector|Monitor could be set up. Capturing calls as a postfix for every single `create_fn` call. 

### [Related Issue](https://github.com/instructor-ai/instructor/discussions/1078)

----

> [!IMPORTANT]
> Fixes `create_with_completion` failures by introducing `Creation` class to wrap responses, updating return types in `instructor/client.py`, `instructor/patch.py`, and `instructor/retry.py`.
> 
>   - **Behavior**:
>     - Fixes `create_with_completion` failing for `AdapterBase`, `ParallelBase`, `IterableBase`, and `PartialBase` by introducing `Creation` class in `instructor/client.py`.
>     - `Creation` class wraps raw and processed responses.
>     - Updates `retry_sync` and `retry_async` in `instructor/retry.py` to return `Creation[T_Model]`.
>   - **Functions**:
>     - Modifies `create`, `create_partial`, `create_iterable`, and `create_with_completion` in `Instructor` and `AsyncInstructor` classes to use `Creation`.
>     - Updates `new_create_sync` and `new_create_async` in `instructor/patch.py` to handle `Creation`.
>   - **Misc**:
>     - Adds `Generic` import in `instructor/client.py` for `Creation` class.
>     - Adjusts type hints and return types across affected functions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 1f30dce2c550d109d6a2cf833b50e387a0174eb0. It will automatically update as commits are pushed.</sup>


jxnl:
will this break for folks who access _raw_response or type inference?

ivanbelenky:
Hey @jxnl good questions
- access to `_raw_response` will not break, I did not eliminate it's assignment on `process_response`
- not sure I know what you mean in the context of these changes by type inference. But I guess the answer is no, mainly because the changes are only wrapping responses from internal `create_fn` calls with a `Creation` object. If you can describe a test case scenario, I am happy to add it to the tst suite as well.

conradlee:
Hey @ivanbelenky great that you tackled this problem.  At the moment my code is starting to get polluted with all sorts of 'dummy models` that look like

```
class ListOfMyModel(BaseModel):
    items: list[BaseModel]
```
and the addition of this boilerplate seems to very much against the ethos of this library.
Any chance you could review this @jxnl ?

ivanbelenky:
@jxnl @ivanleomk any timeline on this?

jxnl:
Closing as part of repository maintenance for pull requests created before 2025.

PR #1135, opened 2024-10-30
feat: fix `max_retries` typing

jordyjwilliams:
When working on a project and [reviewing the docs](https://python.useinstructor.com/concepts/retrying/?h=tenacity#retry-callbacks) I noticed that there were `# type: ignore` references when using a `tenacity`  `Retrying | AsyncRetrying` object.

This PR aims to fix the underlying `type issues` as it seems `max_retries` can always be one of: `int | Retrying | AsyncRetrying`

### DIFF
--- a/instructor/client.py
+++ b/instructor/client.py
@@ -13,6 +13,10 @@ from typing import (
     Literal,
     Any,
 )
+from tenacity import (
+    AsyncRetrying,
+    Retrying,
+)
 from collections.abc import Generator, Iterable, Awaitable, AsyncGenerator
 from typing_extensions import Self
 from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


jordyjwilliams:
@jxnl I am aware that this may be far too simplistic of an approach. from `Contributions` and existing issues I am not sure if this was a known issue or I could be missing something here. 

I am just trying to fix the need for a `type: ignore` within a project where we use `instructor`. Happy to address this in a different way or completely re-evaluate if I have missed something here!

Cheers!

ivanleomk:
Modifying the typehints since AsyncRetrying is really only for async clients so we should be selective about the typing here

Link : https://python.useinstructor.com/concepts/retrying/#other-features-of-tenacity

jordyjwilliams:
@ivanleomk ahh I missed this, thanks a lot. Appreciate the commits and review

ivanleomk (approved):
CI is good to go, going to approve this. Thanks for helping to fix the typing.

jordyjwilliams:
No worries at all. Thanks for the approval!

PR #1194, opened 2024-11-19
Fix tools for OpenAI messages

chris-sanders:
Fixes two different issues with tool use and OpenAI
* OpenAI can (will?) return an empty context when it asks for tool_call. This context is Null and failed to iterate. Null context is replaced with empty list instead.
* Extra arguments were stripped keeping only role content and type. This prevents OpenAI from recognizing tool calls, the non-processed arguments are passed through untouched to preserve them.

----

> [!IMPORTANT]
> Fixes handling of null contexts and preserves additional arguments in OpenAI messages in `convert_messages()` in `instructor/multimodal.py`.
> 
>   - **Behavior**:
>     - Replace null `content` with an empty list in `convert_messages()` to handle empty contexts from OpenAI.
>     - Preserve additional message arguments in `convert_messages()` by including them in the message dictionary.
>   - **Misc**:
>     - Minor changes in `convert_messages()` to ensure compatibility with OpenAI tool calls.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 586300c89598bcce167b0596f71997b0285374fa. It will automatically update as commits are pushed.</sup>


jxnl (approved):

chris-sanders:
That fixes black, note that I added it to the dev requirements as well, if you don't want that it's all in one commit easy to revert.

I looked and the failures now seem to be unrelated, running black on non-python files, connection errors in a tests that appear to reach out to the API direct but are failing long before the changes made here.

PR #1228, opened 2024-11-30
Support individual partial enum values

mwildehahn:
If you use literal values with partial it fails as the model is generating partial values. `PartialLiteralMixin` is one approach to fixing this but then means no other string values will stream as they're processed.

This is another approach that doesn't change how we parse the JSON, but just means the partial model won't validate these enums until they're fully resolved.

----

> [!IMPORTANT]
> Adds support for handling partial enum values in JSON parsing using `PartialLiteralValidator` and updates tests accordingly.
> 
>   - **Behavior**:
>     - Introduces `PartialLiteralValidator` in `partial.py` to handle partial enum values.
>     - Updates `_make_field_optional()` to use `PartialLiteralValidator` for `Literal` types with `Partial`.
>   - **Tests**:
>     - Adds `test_partial_enums()` in `test_partial.py` to validate partial enum handling.
>     - Ensures enums are not validated until fully resolved, except for non-partial fields.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 15d0093c97210264a10fd030ac501744a55e956a. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/instructor/dsl/parallel.py
+++ b/instructor/dsl/parallel.py
@@ -16,6 +16,11 @@ from instructor.mode import Mode
 
 T = TypeVar("T", bound=OpenAISchema)
 
+class MissingToolCall(Exception):
+    def __init__(self, response: Any, *args: Any) -> None:
+        super().__init__(*args)
+        self.response = response
+
 
 class ParallelBase:
     def __init__(self, *models: type[OpenAISchema]):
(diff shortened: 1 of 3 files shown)


jxnl:
Closing as part of repository maintenance for pull requests created before 2025.

PR #1318, opened 2025-01-25
adding support for async validation

tvt173:
Hi,

First of all, thanks for the great project. 

While working with it, I eventually found a need to use an async validator for my response model. This PR adds support for async validation.

I noticed that there is some code in this repo regarding async validators (https://github.com/instructor-ai/instructor/pull/856), which i tried using, but even when using the included async validator on my model, it never seems to be executed.

I am using https://github.com/team23/pydantic-async-validation, but the only requirement for this to work is that the response model has defined on it a `model_async_validate` method, and that the user is using an async client.

----

> [!IMPORTANT]
> Adds async validation support in `process_response_async()` by calling `model_async_validate` if present on `response_model`.
> 
>   - **Behavior**:
>     - Adds async validation support in `process_response_async()` by checking for `model_async_validate` method on `response_model` and calling it if present.
>     - This change is specific to async processing and does not affect synchronous processing.
>   - **Functions**:
>     - Modifies `process_response_async()` in `process_response.py` to include async validation step.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 76a74781ed42e9dc48512e3ca307bb4af404787a. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/instructor/process_response.py
+++ b/instructor/process_response.py
@@ -84,6 +84,8 @@ async def process_response_async(
             response,
             mode=mode,
         )
+        if hasattr(response_model, "model_async_validate"):
+            await response_model.model_async_validate(model)
         return model
 
     model = response_model.from_response(
@@ -92,6 +94,8 @@ async def process_response_async(
         strict=strict,
         mode=mode,
     )
+    if hasattr(response_model, "model_async_validate"):
+        await response_model.model_async_validate(model)
 
     # ? This really hints at the fact that we need a better way of
     # ? attaching usage data and the raw response to the model we return.


tvt173:
@jxnl are you happy with this PR? please let me know if it needs any modifications

ivanleomk:
Hey @tvt173 , the main issue with implementing asynchronous validation is that it needs to be done using an external Mixin. Python's type checking and Pydantic don't play nicely if we put it on the BaseModel.

This current fix might work but I think it is not the best moving forward. I'm happy to work with you on a separate PR where we implement it using a Mixin instead. The code I have it somewhere in an older PR!

Ideally this is separated and isolated to a specific Mixin that people opt in too

tvt173:
hi @ivanleomk , thank you for your review. maybe i don't fully understand your feedback, but i _am_ using a Mixin, the `AsyncValidationModelMixin` defined  in the [pydantic-async-validation](https://github.com/team23/pydantic-async-validation) library. in this PR i simply use duck typing though, so i don't bring in a dependency to that library... i only run async validation if the `model_async_validate` method is defined on the model. since i only run the async validation if the method is defined, it has no effect when the mixin is not used, as you desire.

i do have this implemented and working in a real-life use case. it looks something like this

```python
from pydantic_async_validation import async_model_validator, AsyncValidationModelMixin,

class ClientValidatedModel(AsyncValidationModelMixin, MyModel):
    @async_model_validator()
    async def async_validate_my_model(self) -> Self:
        return await my_async_validation_function(self)

... 

resp: ClientValidatedModel = await client.chat.completions.create(
        model=model,
        response_model=ClientValidatedModel,
        temperature=DEFAULT_TEMPERATURE,
        messages=[
            {"role": "developer", "content": prompt},
            {"role": "user", "content": user_prompt},
        ],
        max_retries=5,
    )
```
please let me know your thoughts.

tvt173:
hi @ivanleomk , thanks for your reply. yes, it looks like they've added support for `ValidationInfo`:
https://github.com/team23/pydantic-async-validation?tab=readme-ov-file#field-validators

image

ivanleomk:
@tvt173 I think this requires you to specify this as a kwarg and uses their own ValidationInfo class rather than the pydantic context variable. 

```
import asyncio

import pydantic
from pydantic_async_validation import (
    async_field_validator,
    AsyncValidationModelMixin,
    ValidationInfo,
)


class SomethingModel(AsyncValidationModelMixin, pydantic.BaseModel):
    name: str

    @async_field_validator("name", some_other_field="some_other_field")
    async def validate_name(
        self, value: str, field: str, config: ValidationInfo
    ) -> None:
        print(config.extra)
        return value


class ParentModel(AsyncValidationModelMixin, pydantic.BaseModel):
    child: SomethingModel


async def main():
    invalid_instance = ParentModel.model_validate_json(
        '{"child": {"name": "invalid"}}', context={"config": {"extra": "forbid"}}
    )
    await invalid_instance.model_async_validate()


if __name__ == "__main__":
    asyncio.run(main())
```

if you look at this, the config context in validation is never passed and instead this validation info requires you to write these kwarg at run time. This isn't great because you can't use information at run-time such as a database query or whatnot.

We can verify that this is the case below when we list the methods and fields on the ValidationInfo object.

```
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', 'extra', 'func']
```

Let me think about this for a bit. I wrote an implementation 5 months ago which I closed the PR on, might use that instead

Issue #1005, opened 2024-09-18
Perplexity API doesn't work


anjor:
- [x] This is actually a bug report.
- [ ] I am not getting good LLM Results
- [ ] I have tried asking for help in the community on discord or discussions and have not received a response.
- [ ] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

- [ ] gpt-3.5-turbo
- [ ] gpt-4-turbo
- [ ] gpt-4
- [x] Other (please specify)
pplx-70b-chat

**Describe the bug**
Using the Perplexity API throws a 400:
```
openai.BadRequestError: Error code: 400 - {'error': {'message': '["At body -> response_format -> ResponseFormatText -> type: Input should be \'text\'", "At body -> response_format -> ResponseFormatJSONSchema -> type: Input should be \'json_schema\'", "At body -> response_format -> ResponseFormatJSONSchema -> json_schema: Field required", "At body -> response_format -> ResponseFormatRegex -> type: Input should be \'regex\'", "At body -> response_format -> ResponseFormatRegex -> regex: Field required"]', 'type': 'bad_request', 'code': 400}}
```

**To Reproduce**

Run the example script from the repo: https://github.com/jxnl/instructor/blob/main/examples/open_source_examples/perplexity.py

**Expected behavior**
The script should run successfully.

jxnl:
This is a known issue, perplexity does not do structured outputs well

rooterkyberian:
It does support them, but with json schema https://docs.perplexity.ai/guides/structured-outputs#1-get-a-response-in-json-format; which seems to be related to #909

Issue #909, opened 2024-08-06
Integrate with new strict OpenAI structured outputs


asim-shrestha:
**Is your feature request related to a problem? Please describe.**
OpenAI just released more strict structured output adherence: https://openai.com/index/introducing-structured-outputs-in-the-api/. 

Taking a look at the client, it seems we're already passing in `strict=True` by default but it would be good to also support `json_schema` for the gpt-4o models that support it 

Quote from the blogpost 
```
2. A new option for the response_format parameter: developers can now supply a JSON Schema via json_schema, a new option for the response_format parameter. This is useful when the model is not calling a tool, but rather, responding to the user in a structured way. This feature works with our newest gpt-4o models: gpt-4o-2024-08-06, released today, and gpt-4o-mini-2024-07-18. When a response_format is supplied with strict: true, model outputs will match the supplied schema.
```

AlmogBaku:
That was fast!


it looks like oai took much inspiration from here 😬

mrdkucher:
+1. Looks like support was added for adding the `strict=True` argument for tool calls (https://github.com/jxnl/instructor/pull/938), but will there be a corresponding `Mode.JSON_STRICT` update using the new "json_schema" support in OpenAI's `response_model` parameter? It would be nice to get guaranteed structured output back as well, besides just in tool calls.

ChristianWeyer:
> +1. Looks like support was added for adding the `strict=True` argument for tool calls (#938), but will there be a corresponding `Mode.JSON_STRICT` update using the new "json_schema" support in OpenAI's `response_model` parameter? It would be nice to get guaranteed structured output back as well, besides just in tool calls.

I would second this. Any opinion on that @jxnl ?

ChristianWeyer:
> i think structured outputs will be better than strict json

Would you elaborate a bit on this?
Thanks :-)

M-Kerr:
Any movement on this? Instructor still submits requests to OpenAI as tool function calls instead of json_schema.

From the [OAI docs](https://platform.openai.com/docs/guides/structured-outputs?context=without_parse#how-to-use):
```
response = client.chat.completions.create(
    model="gpt-4o-2024-08-06",
    messages=[
        {"role": "system", "content": "You are a helpful math tutor. Guide the user through the solution step by step."},
        {"role": "user", "content": "how can I solve 8x + 7 = -23"}
    ],
    response_format={
        "type": "json_schema",
        "json_schema": {
            "name": "math_response",
[...]

PR #1319, opened 2025-01-25
feat: Add support for perplexity sonar

charlieyou:
Perplexity released their Sonar API with structured outputs which is not compatible with any of the existing function calling methods currently implemented. Rest of the API is similar to OpenAI.

See references: 
- https://docs.perplexity.ai/api-reference/chat-completions
- https://docs.perplexity.ai/guides/structured-outputs

Implemented it with a new client, but don't have a strong opinion on keeping it vs only keeping the new mode.

----

> [!IMPORTANT]
> Add support for Perplexity Sonar API with new client and mode, including JSON handling and tests.
> 
>   - **Behavior**:
>     - Add support for Perplexity Sonar API with new client `from_perplexity` in `client_perplexity.py`.
>     - Introduce `PERPLEXITY_JSON` mode in `mode.py` for handling structured outputs.
>     - Update `process_response.py` and `reask.py` to handle `PERPLEXITY_JSON` mode.
>   - **Functions**:
>     - Add `handle_perplexity_json()` in `process_response.py` for processing JSON responses.
>     - Add `reask_perplexity_json()` in `reask.py` for reasking logic.
>   - **Tests**:
>     - Add `test_perplexity` directory with `conftest.py` and `test_modes.py` for testing Perplexity integration.
>     - Define test models and modes in `util.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 73a90619a6dd7be588e6047d5909f67393157c7d. It will automatically update as commits are pushed.</sup>


rooterkyberian:
this resolves #1005 , but the structured outputs problem has been also discussed here #909

Maybe it would be worth to make it more universal?

BTW, if you add "name" to the json_schema object, then the same `mode` would also work with nebius API (which also follows OpenAI API but I think for some reason requires "name") https://openai.com/index/introducing-structured-outputs-in-the-api/ .
AFAIK "name' field in "json_schema" should not break perplexity API response.

PR #1348, opened 2025-02-14
Add support for OpenRouter SO

daswer123:
Hi, recently found your tool and I really like it.

I mainly use openrouter so I was a bit confused when using the `from_openai` client I got errors on some models. 

I looked at the code and saw that I had to enable another mod: `JSON`, `JSON_SCHEMA` or `MD_JSON`. However, none of these mods are used in any of these 

```json
response_format: {
      type: ``json_schema'',
      json_schema: {}
}
```

So according to the documentation from [OpenAI ](https://platform.openai.com/docs/guides/structured-outputs) and [Openrouter ](https://openrouter.ai/docs/features/structured-outputs) I implemented a new mod that uses ``json_schema``
This allowed to use almost all models in Openrouter.

I am still new to SO posture so I might have missed something, I will be glad if you can point out possible errors.

----

> [!IMPORTANT]
> Add `JSON_SCHEMA_OPENROUTER` mode to support OpenRouter's JSON schema format across various functions and files.
> 
>   - **Behavior**:
>     - Adds `JSON_SCHEMA_OPENROUTER` mode to support OpenRouter's JSON schema format.
>     - Updates `from_openai()` in `client.py` to include `JSON_SCHEMA_OPENROUTER` in supported modes.
>   - **Functions**:
>     - Updates `extract_json()` and `extract_json_async()` in `iterable.py` and `partial.py` to handle `JSON_SCHEMA_OPENROUTER`.
>     - Modifies `from_response()` in `function_calls.py` to parse `JSON_SCHEMA_OPENROUTER` mode.
>     - Updates `handle_json_modes()` in `process_response.py` to handle `JSON_SCHEMA_OPENROUTER`.
>   - **Enums**:
>     - Adds `JSON_SCHEMA_OPENROUTER` to `Mode` enum in `mode.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 8af3ed5b04fe1930b3bb6b03623d5506df008dad. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/instructor/client.py
+++ b/instructor/client.py
@@ -508,6 +508,7 @@ def from_openai(
             instructor.Mode.TOOLS,
             instructor.Mode.JSON,
             instructor.Mode.JSON_SCHEMA,
+            instructor.Mode.JSON_SCHEMA_OPENROUTER,
             instructor.Mode.MD_JSON,
         }
 
--- a/instructor/dsl/iterable.py
+++ b/instructor/dsl/iterable.py
@@ -104,6 +104,7 @@ class IterableBase:
                         Mode.JSON,
                         Mode.MD_JSON,
                         Mode.JSON_SCHEMA,
+                        Mode.JSON_SCHEMA_OPENROUTER,
                         Mode.CEREBRAS_JSON,
                         Mode.FIREWORKS_JSON,
                     }:
@@ -145,6 +146,7 @@ class IterableBase:
                         Mode.JSON,
                         Mode.MD_JSON,
                         Mode.JSON_SCHEMA,
+                        Mode.JSON_SCHEMA_OPENROUTER,
                         Mode.CEREBRAS_JSON,
                         Mode.FIREWORKS_JSON,
                     }:
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -268,6 +268,7 @@ class PartialBase(Generic[T_Model]):
                         Mode.JSON,
                         Mode.MD_JSON,
                         Mode.JSON_SCHEMA,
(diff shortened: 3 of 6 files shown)


ivanleomk:
Hmm seems like we already have a JSON_SCHEMA mode in the main branch that was added a year ago  : https://github.com/instructor-ai/instructor/commit/8ced667cc21bbbebb7c2a22c76e0a3e314cb27a7

For instance, if we use the following code with OpenAI

```
from openai import OpenAI
import os
import instructor


class User(BaseModel):
    name: str
    age: int
    number: str = Field(
        None, pattern=r"^\d+$"
    )  # Field for a number that will be validated with regex


client = OpenAI()

client = instructor.from_openai(client, mode=instructor.Mode.TOOLS_STRICT)

for model in [
    "gpt-4o-mini",
]:
    print(f"Testing {model}")
        model=model,
        messages=[
            {
                "role": "system",
                "content": "Extract a user please",
            },
            {
                "role": "user",
                "content": "Ivan is 20 years old and his number is 1234567890",
            },
        ],
        response_model=User,
    )

    assert isinstance(resp, User)
```

This throws the following error

```
instructor.exceptions.InstructorRetryException: Error code: 400 - {'error': {'message': "Invalid schema for function 'User': In context=('properties', 'number'), 'pattern' is not permitted.", 'type': 'invalid_request_error', 'param': 'tools[0].function.parameters', 'code': 'invalid_function_parameters'}}
```

PR #1386, opened 2025-03-09
feat: Adding support for Open Router

ivanleomk:
This PR adds examples for how we can support OpenRouter and experimental support for Structured Outputs with Open Router. 

I tried to add explicit callouts too that you must make sure that your model supports tool calling or structured outputs, or else you will get unforeseen behaviour. This happens because we cannot set open router specific configurations using the OpenAI client.

So users can for instance make tool calls against models.. that don't support structured outputs at all for some reason.

CleanShot 2025-03-09 at 21 41 02@2x

But otherwise, the way I've verified this is by making sure that OpenAI throws an error when we try to pass in a regex pattern ( as written in the docs ) 

----

> [!IMPORTANT]
> Adds support for OpenRouter, including structured outputs and tool calling, with documentation and examples.
> 
>   - **Behavior**:
>     - Adds support for OpenRouter in `from_openai()` in `client.py`, allowing modes `TOOLS`, `OPENROUTER_STRUCTURED_OUTPUTS`, and `JSON`.
>     - Implements `handle_openrouter_structured_outputs()` in `process_response.py` to manage structured outputs.
>     - Updates `from_response()` in `function_calls.py` to handle OpenRouter errors and structured outputs.
>   - **Documentation**:
>     - Adds `openrouter.md` with examples for using OpenRouter with Instructor.
>     - Updates `index.md` to include OpenRouter in the list of integrations.
>     - Updates `mkdocs.yml` to include OpenRouter in the navigation.
>   - **Misc**:
>     - Adds `OPENROUTER` to `Provider` enum in `utils.py`.
>     - Adds `OPENROUTER_STRUCTURED_OUTPUTS` to `Mode` enum in `mode.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 14bfba8a80736538d77399a94169c9a9edeeac80. It will automatically update as commits are pushed.</sup>


instructor/function_calls.py
```diff
@@ -245,6 +252,7 @@ def from_response(
             Mode.CEREBRAS_JSON,
             Mode.FIREWORKS_JSON,
             Mode.PERPLEXITY_JSON,
+            Mode.OPENROUTER_STRUCTURED_OUTPUTS,
```
ivanleomk, line 255: The weird thing is that when we use OpenRouter's structured outputs implementation, the completion comes back as this

This is for Gemini

```
choices=[Choice(finish_reason='stop', index=0, logprobs=None, message=ChatCompletionMessage(content='{"age": 25, "name": "Jason"}', refusal=None, role='assistant', audio=None, function_call=None, tool_calls=None),
```

This is for OpenAI

```
choices=[Choice(finish_reason='stop', index=0, logprobs=None, message=ChatCompletionMessage(content='{"name":"Jason","age":25}', refusal=None, role='assistant', audio=None, function_call=None, tool_calls=None)
```

But the same use of OpenAI's structured outputs with the default OpenAI api returns it as a function call instead.

```
choices=[Choice(finish_reason='stop', index=0, logprobs=None, message=ChatCompletionMessage(content=None, refusal=None, role='assistant', audio=None, function_call=None, tool_calls=[ChatCompletionMessageToolCall(id='call_nrxEG6W16RNL1UDpTVQsIxVr', function=Function(arguments='{"name":"Jason","age":25}', name='User'), type='function')]))]
```

jxnl (approved):

PR #1392, opened 2025-03-11
Fixing Adaptor Base create_with_completion error

ivanleomk:
This adds a fix for the Adaptor Base models so that when users specify either a simple_type or a list[Model] it will preserve the Completion object with the create_with_completion method

----

> [!IMPORTANT]
> Fix `create_with_completion` in `Adaptor Base` to preserve `Completion` object for specific response types and update tests accordingly.
> 
>   - **Behavior**:
>     - Fix `create_with_completion` in `instructor/client.py` to preserve `Completion` object for `simple_type` or `list[Model]`.
>     - Update `create()` and `create_with_completion()` to handle `AdapterBase` responses correctly.
>   - **Testing**:
>     - Add `test_completion_create.py` to test `create_with_completion` with various clients and models.
>     - Modify `test.yml` to ignore `tests/llm/test_completions/` during certain test runs.
>     - Add tests for `simple_type` handling in `test_simple_types.py`.
>   - **Misc**:
>     - Add `python-dotenv` to `dev` dependencies in `pyproject.toml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 572e71acc97b326ebd56f2212e58148b0b2e607d. It will automatically update as commits are pushed.</sup>


jxnl:
I think this can be done way less repetitively

.github/workflows/test.yml
```diff
@@ -31,12 +31,6 @@ jobs:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
           COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
 
-      - name: Run Gemini Tests
```
jxnl, line 34: you deleted tests?

tests/llm/test_writer/test_completion.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Verify we got a valid response
+    assert isinstance(response, bool)
+    assert response is True
+
+    assert isinstance(completion, Chat)
+
+
+@pytest.mark.asyncio
+async def test_create_with_completion_async():
```
jxnl, line 49: I feel like this should be parameterized for basic call

.github/workflows/test.yml
ivanleomk: Let me add it back in for gemini and GenAI SDK

tests/llm/test_writer/test_completion.py
ivanleomk: Hmm perhaps we can parameterise it with the clients and the modes itself in a single test with modelId and some parameters

ivanleomk:
Passing tests

```
tests/llm/test_completion_create.py ...............s..............                                 [100%]
```

ivanleomk:
Added some async tests too

```
tests/llm/test_completions/test_completion_create.py ...............s.............................s..............                     [100%]
```

tests/llm/test_completions/test_completion_create.py
```diff
@@ -0,0 +1,218 @@
...
+class User(BaseModel):
+    name: str
+    age: int
+
+
+response_model_and_prompt = [
+    (User, "Ivan is 20 and lives in Singapore?"),
+    (bool, "Is 30 greater than 20?"),
+]
+clients = [
```
jxnl, line 26: Would this work? This would mean that I have to install every single client right?

ivanleomk: yes this means you need to install every single client with this so we cannot run this in CI. Would you prefer just testing a subset of clients in CI and having the rest commented out?

PR #1441, opened 2025-03-24
add async partial streaming support for genai

oegedijk:
This PR adds support for partial async streaming for google genai sdk. Noticed that async streaming was broken so added a test and then put in the fix.

Main cause of the bug was that the genai client was not returning an AsyncGenerator, but already awaited the result. 
Then had to copy the chunking strategies from the relevant genai modes over from the sync variant. 

Added a test that was first broken and is now passing. 

----

> [!IMPORTANT]
> Fix async streaming in Google GenAI SDK by ensuring correct return type and add tests for verification.
> 
>   - **Behavior**:
>     - Fixes async streaming in `from_genai()` in `client_genai.py` by ensuring `client.aio.models.generate_content_stream` returns an `AsyncGenerator`.
>     - Adds handling for `Mode.GENAI_STRUCTURED_OUTPUTS` and `Mode.GENAI_TOOLS` in `extract_json_async()` in `iterable.py` and `partial.py`.
>   - **Tests**:
>     - Adds `test_partial_model_async()` in `test_stream.py` to verify async streaming functionality with `Partial[UserExtract]` model.
>     - Ensures test passes with the new async streaming fix.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=instructor-ai%2Finstructor&utm_source=github&utm_medium=referral)<sup> for b8e75e0ea3625738436b9914d0052ac2b2a678b6. It will automatically update as commits are pushed.</sup>


instructor/client_genai.py
```diff
@@ -49,7 +48,7 @@ def from_genai(
 
         async def async_wrapper(*args: Any, **kwargs: Any):  # type:ignore
             if kwargs.pop("stream", False):
-                return client.aio.models.generate_content_stream(*args, **kwargs)  # type:ignore
+                return await client.aio.models.generate_content_stream(*args, **kwargs)  # type:ignore
```
oegedijk, line 51: this is the main fix

instructor/dsl/partial.py
```diff
@@ -354,6 +356,18 @@ async def extract_json_async(
                     yield json.dumps(
                         chunk.candidates[0].content.parts[0].function_call.args
                     )
+                if mode == Mode.GENAI_STRUCTURED_OUTPUTS:
```
oegedijk, line 359: copied from extract_json

Issue #1542, opened 2025-05-22
Accept api_key in from_provider


jeroenvds:
**Is your feature request related to a problem? Please describe.**
It would be helpful if you could provide the API key as part of the from_provider constructor, as settings are often managed outside the environment variables. Currently the only option is to pass it to the client, but that removes the LLM provider-agnostic solution of from_provider()

**Describe the solution you'd like**
This code should be accepted
```
    instructor_client = instructor.from_provider(
        model=settings.LLM_MODEL,
        api_key=settings.LLM_API_KEY
    )
```

Today this results in an Exception:
`The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable`

**Describe alternatives you've considered**
- Environment variables: Not easy to change at runtime etc.
- Setting the API key directly on the client: Impossible to write LLM provider-agnostic code

**Additional context**
n/a

Gable-github:
@jxnl i would like to work on this enhancement. Could you assign this issue to me? Thanks!

jxnl:
> [@jxnl](https://github.com/jxnl) i would like to work on this enhancement. Could you assign this issue to me? Thanks!

ty!

jeroenvds:
@jxnl, any guidance on how we can proceed with this feature request? 

@Gable-github had created a PR, refined by @ivanleomk , but you seem to have closed the PR without a comment. If it would be helpful, I'm happy to make some updates myself. But I would first need to understand what wasn't working with the proposed updates in the PR (https://github.com/567-labs/instructor/pull/1556)

PR #1556, opened 2025-05-23
feat: implemented support for accepting api_key in form_provider

Gable-github:
## This PR adds api_key parameter to from_provider function

Description:
This PR addresses issue #1542 by adding an api_key parameter to the from_provider function. This allows users to provide API keys directly to the function rather than having to set them in environment variables or pass them to the client directly.

Yes its the same description as the bot PR. However I noticed some possible errors in its implementation, specifically for these providers:
1. Vertexai
  - it does this
  ```
  if api_key and "api_key" not in kwargs:
                  kwargs["api_key"] = api_key
  ```
  and then passes the `**kwargs` to the genai.Client object which is wrong. `**kwargs` is only supposed to be passed to the from_genai() function where it dosen't include the `api_key` variable. Doing it the bot's way would pass the `api_key` into from_genai() which is not what we want.
  - vertex also uses the Client which takes in an `api_key` and handles it in the package.

2. Perplexity
  - Its handling both cases where api_key is in `**kwargs` but I think its a wrong way to handle it. I handled it in the case where `api_key` is not in `**kwargs`. 

3. generative-ai
  - I've updated it to handle `api_key` according to [this](https://ai.google.dev/gemini-api/docs/migrate) documentation

Other change descriptions:
  - Made error message for missing `api_key` arguments that we have to handle ourselves in this file clearer.

Apologies I failed to follow any convention and seek your kind comments/review :)

#1542 

- [X] I have performed a self-review of my code

### DIFF
--- a/instructor/auto_client.py
+++ b/instructor/auto_client.py
@@ -29,6 +29,7 @@ supported_providers = [
 @overload
 def from_provider(
     model: KnownModelName,
+    api_key: str | None = None,
     async_client: Literal[True] = True,
     **kwargs: Any,
 ) -> AsyncInstructor: ...
@@ -37,6 +38,7 @@ def from_provider(
 @overload
 def from_provider(
     model: KnownModelName,
+    api_key: str | None = None,
     async_client: Literal[False] = False,
     **kwargs: Any,
 ) -> Instructor: ...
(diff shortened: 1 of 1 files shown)


jxnl:
shouldn't it just be in **kwargs and passed into all client initiliazation?

ivanleomk:
Ok I've cleaned this up a fair bit. It seems like vertexai with genai does not work even on their own docs so not going to worry about that. 

Would it be possible to help update the docs with this new API key? You should be able to run this locally with mkdocs serve. 

If you get some errors with `Cairo`, I normally find that I have to set this path

```bash
export DYLD_LIBRARY_PATH=/opt/homebrew/lib
```

ivanleomk:
@Gable-github thanks for this PR!

Gable-github:
i think there are other places i could have added this (individual provider pages, advanced examples, etc) but maybe better to do a follow-up docs PR for the rest? didn't want to make this one huge and hard to review / too out of scope

Gable-github:
I noticed my PR was eventually closed after some time, and as of right now I'm not sure if there was anything specific I could have done differently. Totally understand if my code wasn't up to par. But I’m eager to learn and try contributing again in the future. 

Would you be able to share any feedback or point me toward what I could improve on next time? I really appreciate your time and any guidance you can offer @jxnl

PR #1578, opened 2025-06-03
fix: filter out Gemini thought parts in genai tool parsing

indigoviolet:
The parse_genai_tools method was incorrectly asserting that there should only be one Part in the response, but Gemini can include additional Parts with "thought: true" for its internal reasoning process. 

This change filters out thought parts before checking for multiple function calls, preventing runtime errors when using GENAI_TOOLS mode.

### DIFF
--- a/instructor/function_calls.py
+++ b/instructor/function_calls.py
@@ -305,10 +305,17 @@ class OpenAISchema(BaseModel):
         assert isinstance(completion, types.GenerateContentResponse)
         assert len(completion.candidates) == 1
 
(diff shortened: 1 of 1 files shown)


levzem:
related to #1579

jxnl (commented):
this would not break create_with_completion right?

indigoviolet:
> this would not break create_with_completion right?

I don't think so, this is a very local intervention in the function call parsing -> response model

noby3:
I added to the prompt: "Do not answer in parts." - still works fine for me, but not always

Issue #1531, opened 2025-05-16
GenAI SDK Config isn't applied


ivanleomk:
We don't apply the generation config for the GenAI SDK integration for now

PR #1592, opened 2025-06-06
pop generation_config for genai models and use as additional kwargs

oegedijk:
This PR fixed the passing genai generation_config down to the genai client issue #1531


oegedijk:
ah, hadnt pulled the latest, and didnt realize this was already fixed. closing.

PR #1657, opened 2025-07-06
Json schema fix

Canttuchdiz:
Note: I couldn't find the PR template.

When using a local llm, I was unable to make a successful request. Upon further research, it seems that the format for the json schema is incorrect. In accordance with OpenAI documentation and LM Studio, I changed the response_model format, which now works as intended when using `instructor.Mode.JSON_SCHEMA`.

### DIFF
--- a/instructor/process_response.py
+++ b/instructor/process_response.py
@@ -396,8 +396,11 @@ def handle_json_modes(
         new_kwargs["response_format"] = {"type": "json_object"}
     elif mode == Mode.JSON_SCHEMA:
         new_kwargs["response_format"] = {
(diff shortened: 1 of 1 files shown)


DaveOkpare:
@jxnl Looks good to me

jxnl:
openai says json_object: https://platform.openai.com/docs/guides/structured-outputs#json-mode

jxnl:
oh i see json mode vs json schema in the api

jxnl:
@claude can you write a test in test_openai's basic api that tests every JSON type mode and make a PR into this PR? 

I think right now we only test tools

PR #1720, opened 2025-07-18
Attempt 2 at a refactor

jxnl:
This is one of my attempts at re-factoring the instructor library by providing an adapter that can be swapped in for multiple clients. So far, I've only implemented an OpenAI adapter that only supports a synchronous, non-streaming, single-tool call. 

I will give someone $6,000 if they can build everything else using AI. 

> [!IMPORTANT]
> Refactors instructor library to an adapter-based architecture, focusing on OpenAI support with new client and adapter classes, and adds tests for OpenAI client functionality.
> 
>   - **Architecture**:
>     - Introduces adapter-based architecture in `instructor/v2` for better maintainability.
>     - Adds `InstructorClient`, `ProviderAdapter`, and `from_provider()` for handling LLM provider interactions.
>   - **OpenAI Support**:
>     - Implements `OpenAIAdapter` in `openai.py` for handling OpenAI API requests and responses.
>     - Supports modes: TOOLS, TOOLS_STRICT, JSON, JSON_SCHEMA, PARALLEL_TOOLS.
>     - Adds usage tracking and retry logic.
>   - **Testing**:
>     - Adds `test_openai.py` for testing OpenAI client with different modes and context handling.
>   - **Misc**:
>     - Adds `README.md` for migration overview and guidelines.
>     - Registers `OpenAIAdapter` in `__init__.py` and `registry.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=567-labs%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 3d01a7bbf5993fcbef627afb7d2175655ce895b9. You can [customize](https://app.ellipsis.dev/567-labs/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tests/test_v2/test_openai.py
+++ b/tests/test_v2/test_openai.py
@@ -1,7 +1,6 @@
 from pydantic import BaseModel
 import pytest
-from typing import Literal, Union
-from collections.abc import Iterable
+from typing import Literal
 from instructor.v2.auto_client import from_provider
 from instructor.mode import Mode
 


enyst:
Interesting! Very tempted to do this. The entire thing will take some time, but the fascinating thing about putting our agent to work on it is that
- it's a refactoring (they're hit or miss TBH, "it depends")
- it's a refactoring with a dozen providers, sharing a lot of patterns!

Is the Readme.md good enough for the description of this refactoring, or does it have hallucinations?

Cons/notes:
- Bedrock and maybe even Vertex are a little problematic because I don't have ready accounts. Vertex should be fine though.

Would you be interested in splitting this, like, high priority first, then medium, then low? Or better yet, PR per provider!

Stupid question: if it works for litellm, doesn't that mean that most if not all providers could go through litellm client ? Or what is the reason for native clients for what works via litellm?

PR #1751, opened 2025-07-28
only use thinking_config in GenerateContentConfig

jonbuffington:
If "thinking_config" is not removed from kwargs, the genai
Models.generate_content_stream() function call will raise an
"an unexpected keyword argument" exception.

After making this change, code like the following succeeds:

```
client.chat.completions.create_iterable(
    messages=…,
    response_model=…,
    thinking_config=types.ThinkingConfig(thinking_budget=0),
)
```

### DIFF
--- a/instructor/providers/gemini/utils.py
+++ b/instructor/providers/gemini/utils.py
@@ -824,6 +824,7 @@ def handle_genai_structured_outputs(
     new_kwargs.pop("messages", None)
     new_kwargs.pop("generation_config", None)
     new_kwargs.pop("safety_settings", None)
+    new_kwargs.pop("thinking_config", None)
 
     return response_model, new_kwargs
 


dkkerr:
Thanks @jonbuffington - I needed another reference down around line 899 as well, in case you care to add. Seems chat.completions.create uses tools mode where chat.completions.create_iterable uses structured outputs mode.

oegedijk:
@dkkerr @jonbuffington what is the status on this? I still dont seem to get thinking_config=0 to work even with `.create()`.

This does not seem to impact latency:

```python
instructor_client = instructor.from_genai(genai_client)
response = instructor_client.chat.completions.create(
    model="gemini-2.5-flash",
    messages=[{"role": "user", "content": "Provide a ranked list of 10 famous physicists ranked by their key contributions"}],
    response_model=TopScientists,
    config = types.GenerateContentConfig(
        thinking_config=types.ThinkingConfig(thinking_budget=0)
    )
)
```

and this raises the error mentioned:

```
instructor_client = instructor.from_genai(genai_client)
response = instructor_client.chat.completions.create(
    model="gemini-2.5-flash",
    messages=[{"role": "user", "content": "Provide a ranked list of 10 famous physicists ranked by their key contributions"}],
    response_model=TopScientists,
    thinking_config=types.ThinkingConfig(thinking_budget=0)
)
```

TypeError: Models.generate_content() got an unexpected keyword argument 'thinking_config'

because it does not get filtered out correctly:
```
kwargs =  {'model': 'gemini-2.5-flash', 'thinking_config': ThinkingConfig(
  thinking_budget=0
), 'config': GenerateContentConfig(....```

jonbuffington:
@oegedijk You should check with the library maintainers. I am no longer experimenting with this library.

oegedijk:
@jxnl could you get this fix in? it's just a simple `new_kwargs.pop("thinking_config", None)` so that thinking_config is only part of `config` and not the top-level `kwargs`.

Issue #1663, opened 2025-07-07
Add streaming support for xAI provider


jxnl:
## Description

The xAI provider currently supports synchronous and asynchronous operations with both JSON and TOOLS modes, but does not yet support streaming responses (`create_iterable` and `create_partial`).

## Background

xAI's SDK has a different streaming API compared to other providers. Instead of passing `stream=True` as a parameter, xAI uses a `.stream()` method on the chat object that returns tuples of `(response, chunk)`.

Example from xAI docs:
```python
for response, chunk in chat.stream():
    print(chunk.content, end="", flush=True)
```

## What needs to be done

1. Implement streaming support in `instructor/client_xai.py` for both sync and async clients
2. Handle the conversion from xAI's streaming format to the format expected by instructor's streaming infrastructure
3. Add comprehensive tests in `tests/llm/test_xai/test_stream.py` covering:
   - `create_iterable` for multiple object extraction
   - `create_partial` for partial object streaming
   - Both JSON and TOOLS modes
   - Both sync and async operations

## Technical considerations

- The implementation needs to yield response chunks in a format compatible with instructor's `process_response` function
- For JSON mode: chunks should be extracted from the content
- For TOOLS mode: chunks need to be extracted from tool calls
- The streaming generator should yield chunks as they arrive, not accumulate all content first

## References

- [xAI Streaming Documentation](https://docs.x.ai/api/streaming-response)
- Current xAI implementation: `instructor/client_xai.py`
- Example streaming implementations: See OpenAI or Anthropic providers

This would be a great first contribution for someone familiar with async generators and streaming APIs\!

PR #1758, opened 2025-07-31
feat(xai): add streaming support for xAI provider

jeongyoonm:
## Summary

This PR resolves issue #1663. It adds streaming support for the xAI provider.

## Description
* Implemented iterable and partial streaming features for sync and async xAI clients. 
* Works for both XAI_JSON and XAI_TOOLS mode. 
* Added test cases that do simple and union extraction for every case.

**Note**: XAI_TOOLS does not support streaming. The current implementation works, but effectively it is not streaming.
See https://docs.x.ai/docs/guides/function-calling .

PTAL @jxnl @ivanleomk

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -53,7 +53,6 @@ unfixable = [
   "T201",
   "T203",
 ]
-ignore-init-module-imports = true
 
 [lint.extend-per-file-ignores]
 "instructor/distil.py" = ["ARG002"]
--- a/instructor/auto_client.py
+++ b/instructor/auto_client.py
@@ -920,7 +920,7 @@ def from_provider(
             )
             result = from_xai(
                 client,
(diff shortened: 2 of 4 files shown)


jxnl:
This is great. I'll take a look this weekend after I see the tests pass.

jeongyoonm:
@jxnl  Checks are failing because API keys are not provided. 
Screenshot 2025-08-02 at 8 25 54 AM
Please let me know if there's anything I can do regarding this. Thanks.

jeongyoonm:
@jxnl Just to follow up on this. Checks are failing because API keys are not provided. Please let me know if there's a way to fix this on my end.

jxnl:
Sorry, just life got ahead of me. Taking a look now.

PR #1761, opened 2025-08-02
fix(openai): remove duplicate schema from messages in JSON_SCHEMA mode

pnkvalavala:
Removes redundant schema information from messages when using `JSON_SCHEMA` mode.

### Why This Change?

**JSON mode** (`response_format: {"type": "json_object"}`) - OpenAI docs require explicit JSON instruction in messages since no schema is provided in response_format. 
https://platform.openai.com/docs/guides/structured-outputs?api-mode=chat#json-mode

> When using JSON mode, you must always instruct the model to produce JSON via some message in the conversation, for example via your system message. If you don't include an explicit instruction to generate JSON, the model may generate an unending stream of whitespace and the request may run continually until it reaches the token limit. To help ensure you don't forget, the API will throw an error if the string "JSON" does not appear somewhere in the context.

**JSON_SCHEMA mode** (`response_format: {"type": "json_schema", ...}`) - Schema is already provided in response_format. Adding the same schema to messages creates redundancy, increases token consumption unnecessarily, and provides no additional value to the model.

### Changes

- `JSON_SCHEMA` mode: No schema added to messages (schema already in response_format)
- `JSON` and `MD_JSON` modes: Unchanged behavior (still add schema to messages as required)

### DIFF
--- a/instructor/providers/openai/utils.py
+++ b/instructor/providers/openai/utils.py
@@ -433,22 +433,23 @@ def handle_json_modes(
         )
         new_kwargs["messages"] = merge_consecutive_messages(new_kwargs["messages"])
 
-    if new_kwargs["messages"][0]["role"] != "system":
-        new_kwargs["messages"].insert(
-            0,
-            {
-                "role": "system",
-                "content": message,
-            },
-        )
-    elif isinstance(new_kwargs["messages"][0]["content"], str):
-        new_kwargs["messages"][0]["content"] += f"\n\n{message}"
-    elif isinstance(new_kwargs["messages"][0]["content"], list):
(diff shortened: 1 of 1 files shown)


pnkvalavala:
Hi @jxnl, just following up on this PR. It’s a small change that removes redundant schema duplication in JSON_SCHEMA mode, reducing token usage without affecting JSON or MD_JSON modes.

Would appreciate it if you could take a look when you get a chance.

pnkvalavala:
Hi @jxnl, quick follow-up. I see auto-merge is enabled, but CI is blocking; logs show OIDC token errors in the claude-review job and missing provider API keys in provider tests. Seems unrelated to this diff, but happy to tweak if needed.

Issue #1756, opened 2025-07-31
Gemini cannot deal with Enums in the JSON schema


DavidNemeskey:
I am trying to use Gemini with a schema that includes an Enum and it raises a validation error. The same code works perfectly fine with OpenAI or Together AI models. The MWE below is a combination of [the Gemini primer into structured outputs](https://ai.google.dev/gemini-api/docs/structured-output) and [your own code example for Gemini](https://python.useinstructor.com/integrations/google/):

```Python
import instructor
from google import genai
from pydantic import BaseModel, ConfigDict

from enum import Enum

class RecipeType(Enum):
    DESSERT = 'dessert'
    MAIN = 'main'

class Recipe(BaseModel):
    recipe_type: RecipeType
    ingredients: list[str]

cli = instructor.from_provider(
    "google/gemini-2.5-flash", mode=instructor.Mode.GENAI_STRUCTURED_OUTPUTS
    # "openai/gpt-4.1-mini"
)

user = cli.chat.completions.create(
    messages=[
        {
            "role": "user",
            "content": "Write a recipe.",
        },
    ],
    response_model=Recipe,
)
```

It doesn't matter which `mode` I choose. Also, the commented out GPT model works perfectly. What also works perfectly is running the same through `genai`:

```Python
from enum import Enum

from google import genai
from pydantic import BaseModel


class RecipeType(Enum):
    DESSERT = 'dessert'
    MAIN = 'main'

class Recipe(BaseModel):
    recipe_type: RecipeType
    ingredients: list[str]

client = genai.Client()
response = client.models.generate_content(
    model="gemini-2.5-flash",
    contents="List a few popular cookie recipes, and include the amounts of ingredients.",
    config={
        "response_mime_type": "application/json",
        "response_schema": list[Recipe],
    },
)
# Use the response as a JSON string.
print(response.text)

# Use instantiated objects.
my_recipes: list[Recipe] = response.parsed
```

So the error is definitely how `instructor` interfaces with `genai`. Please look into this, it kind of defeats the purpose of having a unified API if it fails for some providers.

dmastylo:
@DavidNemeskey Had this problem as well. I reported it but it was closed. https://github.com/567-labs/instructor/issues/1469

[...]

PR #1762, opened 2025-08-02
Fix enum validation error in Google GenAI provider

DaveOkpare:
## Changes
  - **Fix response parsing**: Change `model_validate()` to `model_validate_json()` in `parse_genai_tools()` to properly
  handle string-to-enum conversion
  - **Fixes:** #1756

### DIFF
--- a/instructor/processing/function_calls.py
+++ b/instructor/processing/function_calls.py
@@ -284,8 +284,9 @@ class OpenAISchema(BaseModel):
         )
 
         assert function_call.name == cls.openai_schema["name"]
-        return cls.model_validate(
-            obj=function_call.args, context=validation_context, strict=strict
+        # Convert to JSON and back to handle string-to-enum conversion properly
+        return cls.model_validate_json(
+            json.dumps(function_call.args), context=validation_context, strict=strict
         )
 
     @classmethod


jxnl:
this does not feel right / feels like would slow things randomly @claude can you figure out if we can add some tests or fix this another way

DaveOkpare:
@jxnl Gemini’s [structured output docs](https://ai.google.dev/gemini-api/docs/structured-output#json-schema) state:

> “Passing JSON Schema directly is not yet supported when using the SDK.”

The SDK turns enum fields into strings with "enum": [...] constraints, so while using `strict=True` with Pydantic we have these options:

- `model_validate()` rejects these strings and add a fallback to `model_validate_json()` to parse valid enum strings into Enum members while preserving strictness.
- use `model_validate_json()` directly.

Performance test

| Case | n=5	| n=500 |	Winner |
|---|---:|---:|---:|
| Needs coercion (strings) |	11 µs vs 17 µs |	1,090 µs vs 1,674 µs |	model_validate_json |
| Already valid (Enums)	| 5 µs vs 13 µs |	603 µs vs 1,294 µs |	model_validate |

Test compared using `model_validate()` with a fallback to `model_validate_json()` against using `model_validate_json()` for all the Gemini cases.

PR #1799, opened 2025-09-01
fix(responses): generalize tool call parsing for reasoning models

sapountzis:
## Description

Fixes a bug in `parse_responses_tools` method that caused failures when using reasoning models where `completion.output[0]` may not exist or may not have the expected `name` attribute.

## Problem

The previous implementation made unsafe assumptions about the response structure:

1. **Conditional complexity**: Used `using_inbuilt_tools` parameter to handle different response types with branching logic
2. **Unsafe indexing**: Directly accessed `completion.output[0]` without checking if it exists or has the expected properties
3. **Reasoning model incompatibility**: Reasoning models can return responses where the first output element isn't the expected tool call

## Solution

- **Unified approach**: Removed the `using_inbuilt_tools` parameter and conditional logic
- **Safe iteration**: Now iterates through all messages in `completion.output` to find the matching tool call
- **Robust error handling**: Properly handles cases where no matching tool call is found

## Changes

- Removed `using_inbuilt_tools` parameter from `parse_responses_tools` method signature
- Replaced unsafe `completion.output[0]` access with defensive iteration
- Simplified the logic by removing the conditional branch for inbuilt tools
- Updated error handling to be more explicit about missing tool calls

----

> [!IMPORTANT]
> Fixes `parse_responses_tools` to safely handle reasoning models by iterating over `completion.output` and removing unsafe assumptions.
> 
>   - **Behavior**:
>     - Fixes bug in `parse_responses_tools` to handle reasoning models where `completion.output[0]` may not exist or lack `name` attribute.
>     - Iterates through `completion.output` to find matching tool call, removing unsafe direct access.
>     - Handles cases where no matching tool call is found with explicit error.
>   - **Code Changes**:
>     - Removes `using_inbuilt_tools` parameter from `parse_responses_tools`.
>     - Simplifies logic by removing conditional branches for inbuilt tools.
>     - Updates error handling for missing tool calls.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=567-labs%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 9eca15c0f4e294d53effd3516859f6c0f8b8eb7a. You can [customize](https://app.ellipsis.dev/567-labs/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/instructor/processing/function_calls.py
+++ b/instructor/processing/function_calls.py
@@ -201,7 +201,6 @@ class OpenAISchema(BaseModel):
                 completion,
                 validation_context,
                 strict,
-                using_inbuilt_tools=mode == Mode.RESPONSES_TOOLS_WITH_INBUILT_TOOLS,
             )
 
         if not completion.choices:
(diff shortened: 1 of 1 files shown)


jxnl:
@claude please review this

sapountzis:
@jxnl are the tests and claude expected to succeed? My changes are not related to the errors

jxnl:
I just approved workflows to run. Since they cost credits, I try not to make every pull request run everything.

PR #1836, opened 2025-10-04
Fix json parsing

NicolasPllr1:
## Nested code blocks break JSON parsing

JSON parsing with `extract_json_from_codeblock` fails when nested code blocks are in the input string.

I have seen at least two kinds of errors, which I outline below.

**To solve this issue, I propose to simply remove the use of regex.** The code becomes simpler and more predictable. I also doubt that using regex is optimized in any meaningful way here, it could even be slower. 

### Nested 'outer' code blocks

The `_JSON_CODEBLOCK_PATTERN` regex used in  `extract_json_from_codeblock` is too greedy, making the function completely miss the JSON payload.

#### Bad test 

**The only test covering this 'nested code blocks' case wasn't actually testing anything.** 'Hacked' by Claude I guess ...  See the `tests.test_json_extraction_edge_cases` module:

<details>
  <summary> Nested code block </summary>

```python
class TestJSONExtractionEdgeCases:

    def test_nested_codeblocks(self):
        """Test extraction with nested code blocks."""
        text = """
        Outer start
        ```
        Inner start
        ```json
        {"level": "inner"}
        ```
        Inner end
        ```
        Outer end
        """
        # Our regex might have limitations with nested code blocks
        # Let's test this a different way

        # Simplified test with just the JSON part
        simplified = """
        ```json
        {"level": "inner"}
        ```
        """
        result = extract_json_from_codeblock(simplified)
        parsed = json.loads(result)
        assert parsed["level"] == "inner"
```
</details>

**In this case, `extract_json_from_codeblock` was extracting 'Inner start'**.

### Code block in the JSON payload itself

In a similar way, if one value in the JSON is a string which happens to contain a code block, the parsing fails.

<details>
  <summary> Code block in the JSON itself </summary>

```python
    def test_json_with_codeblock_in_a_value(self):
        """Test extraction of JSON that has a value containing a codeblock."""
        text = """
        ```json
        {"name": "```string value with a codeblock```"}
        ```
        """
        result = extract_json_from_codeblock(text)
        parsed = json.loads(result)
        assert parsed["name"] == "```string value with a codeblock```"
```

</details>

**In this case, `extract_json_from_codeblock` was extracting ' {"name": "'**

## Related issue

- #1837 

I added this new test case in `tests.test_json_extraction_edge_cases`.
<!-- ELLIPSIS_HIDDEN -->

----

> [!IMPORTANT]
> Simplifies JSON extraction in `extract_json_from_codeblock` by removing regex, fixing nested code block handling, and updating tests.
> 
>   - **Behavior**:
>     - `extract_json_from_codeblock` in `core.py` now extracts JSON by finding the first '{' and last '}' instead of using regex.
>     - Handles nested code blocks and JSON values containing code blocks.
>   - **Tests**:
>     - Fixes `test_nested_codeblocks` in `test_json_extraction_edge_cases.py` to correctly test nested code blocks.
>     - Adds `test_json_with_codeblock_in_a_value` in `test_json_extraction_edge_cases.py` to test JSON values with code blocks.
>     - Updates `test_json_extraction.py` to remove misleading comments about regex.
>   - **Misc**:
>     - Removes unused regex imports from `core.py`.
> 
> <sup>This description was created by </sup>[<img alt="Ellipsis" src="https://img.shields.io/badge/Ellipsis-blue?color=175173">](https://www.ellipsis.dev?ref=567-labs%2Finstructor&utm_source=github&utm_medium=referral)<sup> for 89988501e14593a978b7a7fc7835f4b789906049. You can [customize](https://app.ellipsis.dev/567-labs/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

<!-- ELLIPSIS_HIDDEN -->

### DIFF
--- a/docs/examples/document_segmentation.md
+++ b/docs/examples/document_segmentation.md
@@ -1,18 +1,18 @@
 ---
-title: 'Document Segmentation with LLMs: A Comprehensive Guide'
+title: "Document Segmentation with LLMs: A Comprehensive Guide"
 description: Learn effective document segmentation techniques using Cohere's LLM, enhancing comprehension of complex texts.
 ---
 
 # Document Segmentation
 
-In this guide, we demonstrate how to do document segmentation using structured output from an LLM. We'll be using [command-r-plus](https://docs.cohere.com/docs/command-r-plus) - one of Cohere's latest LLMs with 128k context length and testing the approach on an article explaining the Transformer architecture. Same approach to document segmentation can be applied to any other domain where we need to break down a complex long document into smaller chunks.
+In this guide, we demonstrate how to do document segmentation using structured output from an LLM. We'll be using [command-a](https://docs.cohere.com/docs/command-a) - one of Cohere's latest LLMs with 256k context length and testing the approach on an article explaining the Transformer architecture. Same approach to document segmentation can be applied to any other domain where we need to break down a complex long document into smaller chunks.
 
 !!! tips "Motivation"
-    Sometimes we need a way to split the document into meaningful parts that center around a single key concept/idea. Simple length-based / rule-based text-splitters are not reliable enough. Consider the cases where documents contain code snippets or math equations - we don't want to split those on `'\n\n'` or have to write extensive rules for different types of documents. It turns out that LLMs with sufficiently long context length are well suited for this task.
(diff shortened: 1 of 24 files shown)


NicolasPllr1:
@jxnl Hey Jason, this is a pretty small PR, can you review it please ? 

It's fixing two tests and essentially removing some code. I think it won't take much of your time, **you will quickly see if you want to keep it or change/discard the changes**.

Note: one of the tests was failing silently, which may be be the most important element of the PR.

Thanks!

Issue #1995, opened 2026-01-13
Partial[T] causes infinite recursion with self-referential models


thomasnormal:
## Description

When using `Partial[T]` with self-referential (recursive) models, it causes infinite recursion and crashes with a `RecursionError`.

## Reproduction

```python
from typing import Optional, List
from pydantic import BaseModel
from instructor.dsl.partial import Partial

class TreeNode(BaseModel):
    """A self-referential model - a tree node that can have children of the same type."""
    value: str
    children: Optional[List["TreeNode"]] = None

TreeNode.model_rebuild()

# This causes RecursionError
PartialModel = Partial[TreeNode]
```

## Root Cause

The `_process_generic_arg()` function in `instructor/dsl/partial.py` (lines 104-108) recursively wraps BaseModel types:

```python
if isinstance(arg, type) and issubclass(arg, BaseModel):
    return Partial[arg]  # Recursively creates Partial[TreeNode] again!
```

When processing `TreeNode`:
1. `Partial[TreeNode]` processes field `children: Optional[List[TreeNode]]`
2. It sees `List[TreeNode]` and tries to wrap `TreeNode` -> `Partial[TreeNode]`
3. `Partial[TreeNode]` processes field `children` again...
4. Infinite loop!

## How Other Systems Handle This

- **Pydantic**: Uses `$ref` and `$defs` in JSON schema - the schema is finite, not recursively expanded
- **JSON Schema**: Standard `$ref` references to `$defs`/`definitions`
- The recursion problem is in instructor's model generation, not in parsing

## Proposed Solution

Track models being processed to prevent re-wrapping:

```python
_processing_models: set[type] = set()

def _process_generic_arg(arg: Any, make_fields_optional: bool = False) -> Any:
    # ... existing code ...
    if isinstance(arg, type) and issubclass(arg, BaseModel):
        if arg in _processing_models:
            return arg  # Already processing, return unwrapped to break cycle
        _processing_models.add(arg)
        try:
            return Partial[arg, MakeFieldsOptional] if make_fields_optional else Partial[arg]
        finally:
            _processing_models.discard(arg)
```

## Use Cases

Recursive models are common for:
- Tree structures (file systems, org charts)
- Linked lists
- Graph representations
- Nested comments/replies
- JSON-LD / linked data

thomasnormal:
## Fix Available in PR #1997

This issue is addressed by PR #1997 which adds recursion tracking to prevent infinite loops.

### How It Works

[...]

PR #1997, opened 2026-01-13
fix(partial): completeness-based validation and recursive model support

### LINKED ISSUE #1993: Model validators fail during partial streaming with incomplete fields
## Description

When using `@model_validator(mode="after")` with partial streaming, validators can fail prematurely because they reference fields that haven't arrived yet in the stream.

Additionally, the Partial model makes all fields optional to allow incremental construction during streaming. However, this means required fields are never enforced, even after streaming completes.

## Example: Model Validator Failure

```python
from typing import Literal
from pydantic import BaseModel, model_validator
import instructor
from openai import OpenAI

class TaskResponse(BaseModel):
    status: Literal["pending", "completed"]
    priority: Literal["low", "high"]
    
    @model_validator(mode="after")
    def validate_consistency(self):
        # This fails during streaming when status is set but priority is None
        if self.status == "completed" and self.priority is None:
            raise ValueError("Completed tasks must have priority")
        return self

# This fails during streaming because the validator runs before all fields arrive
for partial in client.chat.completions.create_partial(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "Create a completed high priority task"}],
    response_model=TaskResponse,
):
    print(partial)  # Fails when status arrives but priority is still None
```

## Example: Missing Required Fields

```python
class UserProfile(BaseModel):
    name: str  # Required
    email: str  # Required
    bio: Optional[str] = None

# If the LLM only returns {"name": "John"}, the Partial model accepts it
# because all fields are made optional during streaming.
# The missing 'email' field is never caught!
```

## Expected Behavior

1. **Model validators** should be skipped during streaming and only run during final validation when all fields are complete.

2. **Required fields** should be enforced after streaming completes. The final object should be validated against the original model to catch missing required fields.

## Additional Issue: Literal Types

A related issue occurs with `Literal` and `Enum` types during streaming. When using `partial_mode="trailing-strings"` (the default), incomplete strings like `"act"` fail validation for `Literal["active", "inactive"]`.

## Solution (PR #1994)

1. **Skip model validators during streaming** by passing `context={"partial_streaming": True}` and wrapping validators to check this context.

2. **Final validation after streaming** validates the final object against the original model to enforce required fields and run model validators.

3. **PartialLiteralMixin** switches to `partial_mode="on"` which drops incomplete strings instead of keeping them.

## Environment

- Python 3.9+
- Pydantic 2.x
- instructor 1.x

thomasnormal:
## Summary

This PR introduces a fundamental improvement to how Partial streaming handles validation, plus fixes infinite recursion with self-referential models.

### Key Changes

1. **Completeness-based validation**: Only validate JSON structures that are "complete" (closed with matching braces/brackets)
2. **Validation-free partial models**: Partial models are now pure data containers with NO validation
3. **Immediate validation on completion**: When JSON becomes complete, validate against the original model immediately
4. **Recursive model support**: Track models being processed to prevent infinite loops

## The Problem

Previously, partial streaming had several fundamental issues:

- **Field constraints** (`min_length`, `max_length`, `ge`, `le`, etc.) would fail on incomplete data during streaming
- **Model validators** would crash when fields were `None` during early chunks
- **Field validators** would fail on partial strings
- **Required fields** had complex workarounds that didn't always work
- **Self-referential models** caused infinite recursion

The old approach was "whack-a-mole" - fixing one validator type at a time without addressing the root cause.

## The Solution

**Core principle**: Only validate JSON structures that are "closed" (complete).

- **Closed object**: `{...}` - starts with `{`, ends with `}`
- **Closed array**: `[...]` - starts with `[`, ends with `]`
- **Open/incomplete**: Missing closing brace/bracket → NO validation

### New Validation Flow

```
Stream: {"user": {"name": "Al
         ↳ user object is OPEN → no validation, yield partial

Stream: {"user": {"name": "Alice"}}
         ↳ user object is CLOSED → validate immediately
         ↳ if min_length=5 fails → RETRY NOW
         ↳ if passes → yield validated object

Stream: {"user": {"name": "Alice"}, "items": [1, 2
         ↳ user is CLOSED + valid
         ↳ items is OPEN → no validation on items yet
         ↳ yield partial with validated user, raw items

Stream: {"user": {"name": "Alice"}, "items": [1, 2]}
         ↳ items is CLOSED → validate items
         ↳ full object CLOSED → all validation complete
```

### New Files

- **`instructor/dsl/json_tracker.py`**: JSON completeness tracker
  - `JsonCompleteness` class tracks which paths in JSON are complete
  - `is_json_complete()` helper function for quick checks
  - Handles edge cases: strings containing braces, escaped quotes, nested structures

### Modified Files

- **`instructor/dsl/partial.py`**:
  - `get_partial_model()` now creates validation-free models (no Field constraints, no validators)
  - `_make_field_optional()` has new `strip_constraints` parameter
  - New `process_potential_object()` with completeness-based validation
  - New `_build_partial_object()` and `_build_partial_list()` helpers
  - All streaming methods updated: `model_from_chunks`, `model_from_chunks_async`, `writer_model_from_chunks`, `writer_model_from_chunks_async`
  - Recursion tracking for self-referential models

### Test Coverage

- **60 tests** in `tests/dsl/test_partial.py`
- New test classes:
  - `TestJsonCompletenessTracker` (14 tests)
  - `TestFieldConstraintsDuringStreaming` (7 tests)
  - `TestRecursiveModels` (4 tests)

### Field Constraints Now Work

```python
class User(BaseModel):
    name: str = Field(min_length=5)  # Previously failed during streaming

# Now works - min_length only checked when JSON is complete
for partial in client.chat.completions.create_partial(
    model="gpt-4o-mini",
    response_model=User,
    messages=[{"role": "user", "content": "Return user Alice"}],
):
    print(partial.name)  # Streams: "A", "Al", "Ali", "Alic", "Alice"
```

### Model Validators Now Work

```python
class Task(BaseModel):
    status: Literal["pending", "completed"]
    priority: int

    @model_validator(mode="after")
    def check_priority(self):
        if self.status == "completed" and self.priority < 5:
            raise ValueError("Completed tasks need high priority")
        return self

# Validator only runs when JSON is complete, not during streaming
```

### Recursive Models Now Work

```python
class TreeNode(BaseModel):
    value: str
    children: Optional[List["TreeNode"]] = None

TreeNode.model_rebuild()

# Previously caused RecursionError, now works
PartialTreeNode = Partial[TreeNode]
```

### Discriminated Unions Now Work

```python
class Cat(BaseModel):
    pet_type: Literal["cat"]
    meows: int

class Dog(BaseModel):
    pet_type: Literal["dog"]
    barks: int

class PetContainer(BaseModel):
    pet: Union[Cat, Dog] = Field(discriminator="pet_type")

# Previously failed - discriminated unions now work with Partial!
PartialModel = Partial[PetContainer]
```

## Validation Types Affected

All these validation types are now properly handled:

- ✅ **Field constraints**: `min_length`, `max_length`, `ge`, `le`, `gt`, `lt`, `pattern`, `multiple_of`
- ✅ **Field validators**: `@field_validator` decorators
- ✅ **Model validators**: `@model_validator` decorators (before/after/wrap modes)
- ✅ **Type validation**: `str` vs `int`, `List[str]` vs `List[int]`, etc.
- ✅ **Required fields**: Fields without defaults
- ✅ **Literal/Enum validation**: Checking value is in allowed set
- ✅ **Discriminated unions**: `Field(discriminator="...")`

## Test Plan

- [x] All 60 partial tests pass
- [x] JSON completeness tracker tests (14 tests)
- [x] Field constraint streaming tests (7 tests)
- [x] Recursive model tests (4 tests)
- [x] Existing functionality preserved

Fixes #1993, #1995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to this project will be documented in this file. The format 
 
 <!-- Add upcoming changes here -->
 
+## [1.14.2] - 2026-01-13
+
+### Fixed
+- Prevented infinite recursion when creating Partial models for recursive Pydantic models (#1997)
+
+### Added
+- Added coverage for recursive Partial models (#1997)
+
 ## [1.14.1] - 2026-01-08
 
 ### Fixed
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "ty>=0.0.1a23",
 ]
 name = "instructor"
-version = "1.14.1"
+version = "1.14.2"
 description = "structured outputs for llm"
 readme = "README.md"
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -115,6 +115,8 @@ typer==0.20.0
     # via instructor (pyproject.toml)
 typing-extensions==4.15.0
     # via
+    #   aiosignal
+    #   anyio
     #   openai
     #   pydantic
     #   pydantic-core
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.9, <4.0"
 resolution-markers = [
     "python_full_version >= '3.13'",
@@ -1766,7 +1766,7 @@ wheels = [
 
 [[package]]
 name = "instructor"
-version = "1.14.1"
+version = "1.14.2"
 source = { editable = "." }
 dependencies = [
     { name = "aiohttp" },


jxnl:
Merged manually after rebasing and resolving conflicts with #1994. Thanks @thomasnormal!

PR #2032, opened 2026-01-26
fix: patch OpenAI client in distil module to support dispatch mode

divyanshupatel17:
## Describe your changes

- Fixed a bug in `instructor.distil` where the `OpenAI` client was not being patched with `instructor`, causing `dispatch` mode to fail because `response_model` was not recognized.
- Added comprehensive unit tests for `instructor.distil` in `tests/test_distil.py` covering key functionality including `get_signature_from_fn`, `format_function`, and the `@distil` decorator in both usage modes.
- Refactored imports in `distil.py` to use top-level relative imports to avoid potential circular dependencies.

(None)

### DIFF
--- a/instructor/distil.py
+++ b/instructor/distil.py
@@ -21,6 +21,7 @@ from pydantic import BaseModel, validate_call
 
 from openai import OpenAI
 from .processing.function_calls import openai_schema
+from .core.patch import patch
 
 
 P = ParamSpec("P")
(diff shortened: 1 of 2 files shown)


jxnl:
I wonder if the right thing to do here is to actually allow the distillation tool to support any client that is an instructor client. This was implemented when opening; that was the only option, but we can go beyond this now.

divyanshupatel17:
Thanks for the feedback @jxnl !

Looking at "examples/distilations/three_digit_mul_dispatch.py" , 
I see it already passes an Instructor client. Should I update to:

1. Accept Union[OpenAI, Instructor] and only patch raw clients?
2. Also rename openai_client -> client for clarity?
3. Or keep the current fix minimal (just patching) and leave broader changes for a follow-up?
Happy to go with whichever approach fits the project best.

jxnl:
Closing — this has been waiting on a maintainer response since January about the desired architecture (supporting any instructor client, not just OpenAI). If you'd like to revisit, please open a new PR addressing the broader design.

PR #2038, opened 2026-01-26
fix(validation): fix unreachable allow_override code in llm_validator

Waqar53:
Fixes a critical bug where the [allow_override](cci:1://file:///Users/waqarazim/Desktop/opensource/instructor-contrib/tests/test_llm_validators.py:92:4-109:32) feature in [llm_validator()](cci:1://file:///Users/waqarazim/Desktop/opensource/instructor-contrib/instructor/validation/llm_validators.py:8:0-77:14) was completely broken - the code was unreachable dead code due to an `assert` statement.
## The Bug
```python
# Line 69 - Raises AssertionError if resp.is_valid is False
# Lines 71-73 - UNREACHABLE! (assert already crashed)
if allow_override and not resp.is_valid and resp.fixed_value is not None:
    return resp.fixed_value  # This line NEVER executes
Impact: Users who set allow_override=True expecting auto-correction get an AssertionError instead of the corrected value.

The Fix
python
if not resp.is_valid:
    if allow_override and resp.fixed_value is not None:
        return resp.fixed_value
    raise ValueError(resp.reason)
return v
Changes
✅ Restructure logic so 
allow_override
 check runs before any exception
✅ Use ValueError instead of assert for better error handling
✅ Update docstring to match actual parameters
✅ Add 5 comprehensive tests for 
allow_override
 functionality
Type of Change
 Bug fix (non-breaking change that fixes an issue)
Testing
Added 
tests/test_llm_validators.py
 with 5 tests covering:

Valid input returns unchanged
Invalid with allow_override=True returns 
fixed_value
Invalid with allow_override=False raises ValueError
Invalid without 
fixed_value
 raises ValueError
Default allow_override=False behavior

### DIFF
--- a/tests/test_llm_validators.py
+++ b/tests/test_llm_validators.py
@@ -9,7 +9,7 @@ Tests cover:
 """
 
 import pytest
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock
 
 from instructor.validation.llm_validators import llm_validator
 from instructor.processing.validators import Validator

### CODE REVIEW

File: instructor/validation/llm_validators.py [hunk 1]
```diff
@@ -64,13 +65,14 @@ def llm(v: str) -> str:
...
-        # If the response is  not valid, return the reason, this could be used in
-        # the future to generate a better response, via reasking mechanism.
-        assert resp.is_valid, resp.reason
+        # Handle validation result
+        if not resp.is_valid:
+            # Check if we should return the fixed value instead of failing
+            if allow_override and resp.fixed_value is not None:
+                return resp.fixed_value
+            # Raise a proper ValueError with the LLM's explanation
+            raise ValueError(resp.reason)
```
jxnl, line 74: can you check if there is a better error to raise there @cursoragent

jxnl:
Closing as duplicate — #2215 addresses the same llm_validator allow_override fix.

---

PR #2064, opened 2026-02-07
fix(cli): show full batch ID in batch list

### LINKED ISSUE #1871: Batch CLI can't show full batch ID
https://github.com/567-labs/instructor/blob/ecb50aaac8f197146c0c6ca90efcc934b728289c/instructor/cli/batch.py#L22

It is hard to manage a batch if the batch ID is partially shown. For example:

Image

thecaptain789:
Remove max_width constraint on Batch ID column so the full ID is visible.

Previously truncated at 20 chars, making it difficult to manage batches since OpenAI batch IDs are longer than 20 characters.

Fixes #1871


jxnl:
Closing — the full batch ID display (via the `--full-id` flag and the `full_id` parameter in `generate_table`) is already implemented in the current codebase. This fix has been incorporated.

PR #2073, opened 2026-02-08
fix(templating): return kwargs instead of None in handle_templating

Waqar53:
Fixed missing return values in handle_templating() function. Previously, the function returned None implicitly in two early-exit cases, which could cause unexpected behavior for callers.

- Line 122: Return kwargs when messages list is empty
- Line 127: Return kwargs when no messages found in dict


jxnl:
Closing — this fix has been applied directly in the staging/v1.15.1 branch. The early-exit paths in `handle_templating` now return `new_kwargs` instead of `None`. Thanks!

PR #2085, opened 2026-02-16
feat(anthropic): add User-Agent header for Anthropic API calls

mikelambert:
This adds the user-agent to the API calls so that traffic can be attributed back to the libraries people use


jxnl:

PR #2091, opened 2026-02-23
fix(partial): handle PEP 604 union types in Partial streaming

### LINKED ISSUE #2088: feat(StreamingWithUnionType): Add support for streaming BaseModel with union Types
I've got a BaseModel with unionTypes and i cannot stream the attribute

```python
from instructor import Partial

class MyResponse:
     value: str | int

Partial[MyResponse]

   1026     modified_args = tuple(_process_generic_arg(arg) for arg in generic_args)
   1028     # Reconstruct the generic type with modified arguments
   1029     tmp_field.annotation = (
-> 1030         generic_base[modified_args] if generic_base else None
   1031     )
   1032 # If the field is a BaseModel, then recursively convert it's
   1033 # attributes to optionals.
   1034 elif isinstance(annotation, type) and issubclass(annotation, BaseModel):
   1035     # Prevent infinite recursion for self-referential models

TypeError: type 'types.UnionType' is not subscriptable
```

I want to be able to stream MyResponse with the value that has union types.

It's not that important, i can workaround and make it work by going toward a more general type like str or list[str]. But i guess for some users, and some usecases, it might be frustrating to not have support for union types.

We might also wait and hope a release of Pydantic that will do the heavy lifting on this part.

**Additional context**

kvr06-ai:
## Summary

Fixes #2088. `Partial[Model]` crashes with `TypeError: type 'types.UnionType' is not subscriptable` when a model field uses PEP 604 union syntax (`str | int`) instead of `typing.Union[str, int]`.

## Root cause

`types.UnionType` (Python 3.10+ `X | Y` syntax) cannot be subscripted with `[]`. The codebase already handles this in `_process_generic_arg` (normalizing to `typing.Union`), but the same guard was missing from two other locations that reconstruct generic types:

1. `_wrap_models` (inner function in `Partial.__class_getitem__`) — the default `Partial[Model]` path
2. `_make_field_optional` — the `Partial[Model, MakeFieldsOptional]` path

Both tried `generic_base[modified_args]` where `generic_base` is `types.UnionType`, which raises `TypeError`.

## Changes

- `instructor/dsl/partial.py`: Add `UNION_ORIGINS` guard before subscripting `generic_base` in both `_wrap_models` and `_make_field_optional`. When the origin is a union type, use `Union[modified_args]` (which is subscriptable) instead.
- `tests/dsl/test_partial.py`: Add `TestPEP604UnionTypes` test class with 6 regression tests covering both code paths (basic union, union in list, nested models, MakeFieldsOptional path, `_make_field_optional` directly).

## Testing

- All 6 new tests pass
- All 49 existing tests in `test_partial.py` pass (2 skipped, requiring OpenAI API key)
- No regressions

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -277,9 +277,12 @@ def _make_field_optional(
         )
 
         # Reconstruct the generic type with modified arguments
-        tmp_field.annotation = (
-            Optional[generic_base[modified_args]] if generic_base else None
-        )
+        if generic_base in UNION_ORIGINS:
+            tmp_field.annotation = Optional[Union[modified_args]]  # type: ignore
+        else:
+            tmp_field.annotation = (
+                Optional[generic_base[modified_args]] if generic_base else None
+            )
         tmp_field.default = None
         tmp_field.default_factory = None
     # If the field is a BaseModel, then recursively convert it's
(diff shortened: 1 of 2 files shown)


kvr06-ai:
@jxnl — Rebased on latest main. This fixes `Partial[Model]` crashing with PEP 604 union syntax (`str | int`) due to `types.UnionType` not being subscriptable. Two-location guard + 6 regression tests. Would appreciate a review.

jxnl:
Closing as duplicate — #2200 addresses the same PEP 604 union type fix for Partial streaming.

---

PR #2097, opened 2026-02-24
fix: update test_openai_schema_raises_error to expect ConfigurationError

mikatammi:
The openai_schema() function raises ConfigurationError (not TypeError) when given a non-BaseModel class, but the test was never updated to reflect this change. Update the expected exception type and match string to align with the actual behavior.


jxnl:
Closing as stale — no activity in 5 weeks. If this test fix is still relevant, please rebase against main and reopen.

PR #2103, opened 2026-02-27
Move diskcache from hard dependency to optional extra (CVE-2025-69872)

### LINKED ISSUE #2101: (diskcache) dependency related Vulnerability scan issue CVE-2025-69872

debu-sinha:
Closes #2101

## What changes are proposed

`diskcache` uses `pickle.load()` internally and is affected by CVE-2025-69872 (no upstream patch available). Since instructor already lazy-imports diskcache via `_import_diskcache()` and only uses it when `DiskCache` is explicitly requested, it does not need to be a hard dependency.

This moves `diskcache` from `[project.dependencies]` to `[project.optional-dependencies]` so that `pip install instructor` no longer pulls it in unconditionally. Users who need `DiskCache` can install with:

```
pip install 'instructor[diskcache]'
```

## Changes

- `pyproject.toml`: Remove `diskcache` from hard deps, add `diskcache` optional extras group
- `requirements.txt`: Remove `diskcache` pin
- `instructor/cache/__init__.py`: Update error message to mention `instructor[diskcache]` install syntax

## How is this PR tested

No behavior change for users who have diskcache installed. The `_import_diskcache()` function already handles the missing case and raises a clear `ImportError` with install instructions. Existing tests in `test_cache_key.py` and `test_cache_integration.py` do not depend on diskcache.

### DIFF
--- a/instructor/cache/__init__.py
+++ b/instructor/cache/__init__.py
@@ -115,7 +115,8 @@ def _import_diskcache():  # pragma: no cover – only executed when requested
 
     if importlib.util.find_spec("diskcache") is None:  # type: ignore[attr-defined]
         raise ImportError(
-            "diskcache is not installed.  Install it with `pip install diskcache`."
+            "diskcache is not installed. Install it with "
+            "`pip install 'instructor[diskcache]'` or `pip install diskcache`."
         )
     import diskcache  # type: ignore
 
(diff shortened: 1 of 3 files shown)


jacobcui:
Bump this PR if anyone(with write access) can approve it. At the moment, we have to use the commit in our requirements.txt

git+https://github.com/debu-sinha/instructor.git@23f00c440f25a2b399f73083c9f350a7a5ef0ee0

jxnl:
Closing as duplicate — #2211 addresses the same diskcache optional dependency change.

---

PR #2104, opened 2026-02-28
fix: make diskcache an optional dependency (CVE-2025-69872)

### LINKED ISSUE #2101: (diskcache) dependency related Vulnerability scan issue CVE-2025-69872

giulio-leone:
## Problem

`diskcache` is listed as a required dependency, but it is only used when `DiskCache` is explicitly instantiated. The library already has a lazy import pattern (`_import_diskcache()`) that raises `ImportError` when diskcache is not installed.

This causes security scan failures in corporate environments due to **CVE-2025-69872** (pickle deserialization vulnerability in diskcache), blocking adoption of instructor.

## Solution

- Remove `diskcache` from `[project].dependencies`
- Add `diskcache` as an optional dependency group: `pip install 'instructor[diskcache]'`
- Update the `ImportError` message to reference the extras syntax
- Keep `diskcache` in `test-docs` group for CI

## Backward Compatibility

No breaking change for users who don't use `DiskCache` — they get a smaller, CVE-free install.

Users who use `DiskCache` need to change their install to:
```bash
pip install 'instructor[diskcache]'
```

The lazy import already handles this gracefully with a clear error message.

Fixes #2101

### DIFF
--- a/instructor/cache/__init__.py
+++ b/instructor/cache/__init__.py
@@ -115,7 +115,7 @@ def _import_diskcache():  # pragma: no cover – only executed when requested
 
     if importlib.util.find_spec("diskcache") is None:  # type: ignore[attr-defined]
         raise ImportError(
-            "diskcache is not installed.  Install it with `pip install diskcache`."
+            "diskcache is not installed.  Install it with `pip install instructor[diskcache]`."
         )
     import diskcache  # type: ignore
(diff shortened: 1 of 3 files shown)


giulio-leone:
Closing in favor of #2105 which has the same fix.

PR #2108, opened 2026-02-28
fix(streaming): pre-fill single-value Literal defaults in partial responses

### LINKED ISSUE #2054: Streaming with Literal["Value"] = "value"

``` python 
class Person(BaseModel):
  name: str
  age: int
```

When streaming I get:

```
{ "name": " Joh", "age": {} }
...
```

weiguangli-io:
## Summary

When streaming with `Partial`, fields typed as `Literal["Value"] = "Value"` (single possible value with matching default) are set to `None` until the LLM happens to return them. Since these values are deterministic, they should be pre-filled immediately in every partial response.

Example model:
```python
class Person(BaseModel):
    name: str
    age: int
```

Before this fix, streaming `Partial[Person]` would show `type=None` until the LLM returned the `type` field. After this fix, `type="Person"` appears from the first partial response.

fixes #2054

## Changes

- `instructor/dsl/partial.py`: Added `_has_deterministic_default()` helper that detects single-value `Literal` fields with matching defaults, and uses it in `_build_partial_object()` to pre-fill these fields instead of setting them to `None`
- Added `tests/test_streaming_literal_default.py` with 11 unit tests covering detection logic, partial object building, nested models, and edge cases

## Test plan

- [x] All 11 new tests pass locally
- [x] Existing related tests pass (test_list_response_wrapper, test_list_response)
- [ ] CI passes

AI-assisted review, AI-assisted testing

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -20,6 +20,7 @@ from typing import (  # noqa: UP035
     Any,
     Generic,
     List,  # needed for runtime check against typing.List annotations from user code
+    Literal,
     NoReturn,
     Optional,
     TypeVar,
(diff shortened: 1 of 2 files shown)


weiguangli-io:
Closing due to inactivity. Happy to reopen if there's interest.

Issue #2113, opened 2026-03-02
Spurious errors from litellm due to changed type of response.usage


kouk:
- [X] This is actually a bug report.
- [ ] I am not getting good LLM Results
- [ ] I have tried asking for help in the community on discord or discussions and have not received a response.
- [ ] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

- [ ] gpt-3.5-turbo
- [ ] gpt-4-turbo
- [ ] gpt-4
- [X] Other: any

**Describe the bug**
In https://github.com/567-labs/instructor/pull/343 code was added (now living in [the update_total_usage method of instructor.utils.core](https://github.com/567-labs/instructor/blob/e45280d9d15942fdf86f87d5f10dcc1522b49c48/instructor/utils/core.py#L326)) that updates the response object with the total tokens used. However when used in conjunction with Litellm this introduces a bug (discussed in  https://github.com/BerriAI/litellm/issues/13672) because the type of `response.usage` unexpectedly changes from one subclass of OpenAIUsage to another (that lacks some methods that litellm added). There is code in litellm (the Langfuse integration) that depends on the response.usage field having this specific type.

**To Reproduce**

Run [repro.py](https://github.com/user-attachments/files/25683681/repro.py)

```
$ uvx --with langfuse,instructor[litellm]  --exclude-newer 2026-01-01 python repro.py
python: 3.12.4 (main, Jul 12 2024, 09:00:16) [Clang 15.0.0 (clang-1500.3.9.4)]
instructor: 1.13.0
litellm: 1.80.11
langfuse: 3.11.2
14:16:02 - LiteLLM:WARNING: opentelemetry.py:186 - Proxy Server is not installed. Skipping OpenTelemetry initialization.
14:16:05 - LiteLLM:ERROR: _utils.py:273 - [Arize/Phoenix] Failed to set OpenInference span attributes: 'CompletionUsage' object has no attribute 'get'
```

**Expected behavior**
No error

weiguangli-io:
I've submitted a fix in #2116.

The root cause is in `update_total_usage()`: line 350 replaces `response.usage` with a plain `CompletionUsage` object (`total_usage`), discarding the original litellm-specific `Usage` subclass. This breaks integrations like Langfuse that depend on litellm's custom methods.

The fix copies accumulated token totals back to the original usage object instead of replacing it, preserving the provider-specific type.

kouk:
[...]

PR #2114, opened 2026-03-02
Don't replace response_object.usage to remain compatible with litellm…

kouk:
… and langfuse. Fixes: #2113


jxnl:
Superseded by #2171. The triage batch already carries the approved  subtype preservation fix for #2113, so I’m consolidating on that PR.

PR #2118, opened 2026-03-03
fix(utils): update usage in-place to preserve LiteLLM usage type

CloveSVG:
Instead of replacing `response.usage` with a new object, update token fields in-place so LiteLLM's custom methods on the usage object are preserved.\n\nFixes #2113


kouk:
This was fixed here https://github.com/567-labs/instructor/pull/2114

ghost:
Closing as this issue was already addressed in #2114. Thanks!

PR #2123, opened 2026-03-04
fix(parallel): handle ParallelBase class check in process_response

### LINKED ISSUE #2049: bug in ParallelTools call, model: Qwen3-VL-4b-instruct, server: LM-Studio 0.4.1

**What Model are you using?**

Qwen3-VL-4b-instruct

**Describe the bug**
error:
```xml
<last_exception>
</last_exception>
```
Root Cause:

python

Should be:

python

weiguangli-io:
## Summary

Fixes #2049

The `isinstance(response_model, ParallelBase)` check in both `process_response` and `process_response_async` fails when `response_model` is a **class** (not an instance). This happens with non-OpenAI models (e.g., Qwen via LM Studio) where `response_model` may be passed as a `ParallelBase` subclass rather than a `ParallelBase` instance.

### Changes

- Updated the `ParallelBase` check in `process_response()` and `process_response_async()` to handle both cases:
  - `isinstance(response_model, ParallelBase)` -- for instances (existing OpenAI path via `ParallelModel()`)
  - `inspect.isclass(response_model) and issubclass(response_model, ParallelBase)` -- for classes (non-OpenAI path)
- Added 9 unit tests covering both instance and class scenarios, including a regression test demonstrating the original bug

### Root Cause

In `process_response.py`, the check `isinstance(response_model, ParallelBase)` only works when `response_model` is an **instance** of `ParallelBase` (as created by `ParallelModel()`). When a class that inherits from `ParallelBase` is passed directly as `response_model`, `isinstance` returns `False` because the class object itself is not an instance of `ParallelBase` -- it's a subclass. This causes the code to skip the parallel handling branch and fall through, resulting in `'generator' object has no attribute '_raw_response'`.

## Test plan

- [x] All 9 new tests in `tests/test_parallel_response.py` pass
- [x] Existing tests (`test_process_response.py`, `test_multitask.py`) pass with no regressions
- [x] Ruff lint and format checks pass

### DIFF
--- a/instructor/processing/response.py
+++ b/instructor/processing/response.py
@@ -263,7 +263,9 @@ async def process_response_async(
             raw_response=response,
         )
 
-    if isinstance(response_model, ParallelBase):
+    if isinstance(response_model, ParallelBase) or (
+        inspect.isclass(response_model) and issubclass(response_model, ParallelBase)
+    ):
         logger.debug(f"Returning model from ParallelBase")
         model._raw_response = response
         return model
@@ -381,7 +383,9 @@ def process_response(
             raw_response=response,
         )
 
-    if isinstance(response_model, ParallelBase):
+    if isinstance(response_model, ParallelBase) or (
+        inspect.isclass(response_model) and issubclass(response_model, ParallelBase)
+    ):
         logger.debug(f"Returning model from ParallelBase")
         model._raw_response = response
         return model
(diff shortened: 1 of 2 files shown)


weiguangli-io:
Closing due to inactivity. Happy to reopen if there's interest.

PR #2133, opened 2026-03-08
fix: preserve field defaults in partial streaming instead of using None

### LINKED ISSUE #2054: Streaming with Literal["Value"] = "value"

MaxwellCalkin:
## Summary

Fixes #2054.

When streaming incomplete JSON via `Partial`, the `_build_partial_object()` function unconditionally set missing fields to `None`, ignoring any declared default values. This caused fields like `type: Literal["Person"] = "Person"` to appear as `None` in every partial response until the LLM happened to stream them — typically at the very end.

This is a problem for frontend rendering that relies on discriminator fields (e.g. `type`) to select components, since the field is `None` through most of the stream.

## Changes

**`instructor/dsl/partial.py`**:
- Import `PydanticUndefined` from `pydantic_core`
- In `_build_partial_object()`, check `field_info.default` and `field_info.default_factory` before falling back to `None` for missing fields

**`tests/dsl/test_partial.py`**:
- Add `TestFieldDefaultsDuringStreaming` class with 5 tests:
  - Literal field with default preserved in partial stream
  - Literal default preserved alongside complete fields
  - `default_factory` fields (e.g. `list`) use factory value
  - Simple defaults (`bool`, `int`) preserved
  - Fields without defaults still return `None` (no regression)

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -31,6 +31,7 @@ from typing import (  # noqa: UP035
 from jiter import from_json
 from pydantic import BaseModel, create_model
 from pydantic.fields import FieldInfo
+from pydantic_core import PydanticUndefined
 
 from instructor.mode import Mode
 from instructor.utils import extract_json_from_stream, extract_json_from_stream_async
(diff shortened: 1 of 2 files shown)


jxnl:
Closing as duplicate — #2204 addresses the same partial streaming field defaults fix.

---

PR #2135, opened 2026-03-08
fix: handle GEMINI_TOOLS in async streaming paths

aerosta:
## Summary

This fixes a sync/async inconsistency in streaming JSON extraction for `Mode.GEMINI_TOOLS`.

In both `PartialBase` and `IterableBase`, the sync streaming path applies JSON extraction for both `Mode.MD_JSON` and `Mode.GEMINI_TOOLS`, but the async streaming path only handled `Mode.MD_JSON`. That means async Gemini tools streaming could skip the same extraction step that the sync path already applies.

## Changes

- updated `instructor/dsl/partial.py` so the async streaming path handles `Mode.GEMINI_TOOLS`
- updated `instructor/dsl/iterable.py` so the async streaming path handles `Mode.GEMINI_TOOLS`

## Tests

Added `tests/dsl/test_gemini_tools_async_streaming.py` covering:

- sync JSON extraction from code blocks
- async JSON extraction from code blocks
- sync `Mode.GEMINI_TOOLS` triggering JSON extraction
- async `Mode.GEMINI_TOOLS` triggering JSON extraction

```bash
python3 -m pytest tests/dsl/test_gemini_tools_async_streaming.py -v -p no:deepeval

### DIFF
--- a/instructor/dsl/iterable.py
+++ b/instructor/dsl/iterable.py
@@ -50,7 +50,7 @@ class IterableBase:
     ) -> AsyncGenerator[BaseModel, None]:
         json_chunks = cls.extract_json_async(completion, mode)
 
-        if mode == Mode.MD_JSON:
+        if mode in {Mode.MD_JSON, Mode.GEMINI_TOOLS}:
             json_chunks = extract_json_from_stream_async(json_chunks)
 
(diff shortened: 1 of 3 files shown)


aerosta:
Thank you

PR #2136, opened 2026-03-08
fix(process_response): use issubclass instead of isinstance for ParallelBase check

### LINKED ISSUE #2049: bug in ParallelTools call, model: Qwen3-VL-4b-instruct, server: LM-Studio 0.4.1

s-zx:
## Problem

When `response_model` is a `ParallelBase` subclass (e.g. `Parallel[ToolA, ToolB]`), the check `isinstance(response_model, ParallelBase)` always returns `False` because `response_model` is a **class object**, not an instance of `ParallelBase`.

This causes the function to fall through to `model._raw_response = response`, which crashes when `model` is a generator:

```
```

Fixes #2049

## Root Cause

In `instructor/processing/response.py`, there are two places with:

```python
if isinstance(response_model, ParallelBase):   # ← WRONG
```

`response_model` is always a **type** (class), not an instance. The correct check is:

```python
if inspect.isclass(response_model) and issubclass(response_model, ParallelBase):   # ✓
```

## Fix

Replace both `isinstance` checks with `inspect.isclass(...) and issubclass(...)` — `inspect` is already imported in the file.

This matches the pattern already used elsewhere in the codebase for checking `IterableBase`, `PartialBase`, etc.

### DIFF
--- a/instructor/processing/response.py
+++ b/instructor/processing/response.py
@@ -263,7 +263,7 @@ async def process_response_async(
             raw_response=response,
         )
 
-    if isinstance(response_model, ParallelBase):
+    if inspect.isclass(response_model) and issubclass(response_model, ParallelBase):
         logger.debug(f"Returning model from ParallelBase")
         model._raw_response = response
         return model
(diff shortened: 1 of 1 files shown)


jxnl:
Duplicate of #2143. This issue is caused by trying to attach _raw_response to a generator; #2143 fixes that directly and is the branch I’m carrying forward for #2049.

PR #2141, opened 2026-03-11
fix(mistral): handle breaking import change in mistralai v2.0.0

### LINKED ISSUE #2137: bug: New Mistralai version 2.0.0 released 3h ago breaks instructor

**Describe the bug**
New Mistralai version 2.0.0 released 3h ago breaks instructor:

  File "/home/runner/work/.venv/lib/python3.10/site-packages/instructor/providers/__init__.py", line 55, in <module>
    from .mistral.client import from_mistral  # noqa: F401
  File "/home/runner/work/.venv/lib/python3.10/site-packages/instructor/providers/mistral/client.py", line 5, in <module>
    from mistralai import Mistral
ImportError: cannot import name 'Mistral' from 'mistralai' (unknown location)

Mistralai seems to have changed their import. from mistralai import Mistral is no longer valid. This causes instructor not to work for other providers as well if mistral is installed.

**To Reproduce**
Install latest mistralai version 2.0.0 and try using instructor 1.14.5

Kitan-Dara06:
Title: fix: resolve ImportError by handling mistralai v2.0.0 package structure

Description:
The recent release of mistralai v2.0.0 (March 10, 2026) introduced breaking changes to the internal package structure, specifically moving the Mistral client from the root namespace to mistralai.client. This prevents the instructor Mistral provider from initializing, even for users who are not explicitly using Mistral models.
This PR implements a backward-compatible import strategy using a try-except block. This ensures that:

v2.x users are correctly routed to the new mistralai.client path.
v1.x users (legacy) can continue using the library without upgrading.
Changes:

Modified instructor/providers/mistral/client.py to handle both mistralai v1.x and v2.x import paths.
Validation:

Verified successful import on mistralai==2.0.0.
Verified successful import on mistralai==1.5.0 (Legacy).
Confirmed that the fix unblocks other providers when mistralai is present in the environment.
Fixes #2137

### DIFF
--- a/instructor/providers/mistral/client.py
+++ b/instructor/providers/mistral/client.py
@@ -2,7 +2,10 @@
 from __future__ import annotations
 
 
-from mistralai import Mistral
+try:
+    from mistralai.client import Mistral  # New v2 path
+except ImportError:
+    from mistralai import Mistral         # Old v1 path
 import instructor
 from typing import overload, Any, Literal
 
--- a/tests/llm/test_mistral/test_import.py
+++ b/tests/llm/test_mistral/test_import.py
@@ -1,0 +1,54 @@
+import os
+import pytest
+from pydantic import BaseModel
+import instructor
+
+try:
+    # mistralai v2.0.0+
+    from mistralai.client import Mistral
+except ImportError:
+    # mistralai v1.x
+    from mistralai import Mistral
+
+
+class UserDetail(BaseModel):
+    name: str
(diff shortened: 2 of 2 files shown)


jxnl:
Duplicate of #2152. I’m using #2152 as the canonical PR for issue #2137 because it covers the runtime import fix and the dependency metadata. The test coverage in this branch is useful and should be folded into the chosen fix before merge.

PR #2143, opened 2026-03-13
fix: consume ParallelBase generator into ListResponse to avoid _raw_response crash

### LINKED ISSUE #2049: bug in ParallelTools call, model: Qwen3-VL-4b-instruct, server: LM-Studio 0.4.1

giulio-leone:
## Problem

`ParallelBase.from_response()` is a generator function (uses `yield`), so calling `response_model.from_response(...)` returns a generator object. Both `process_response()` and `process_response_async()` then attempt:

```python
model._raw_response = response  # model is a generator → AttributeError
```

Generators don't support setting arbitrary attributes, causing:

```
```

This breaks any use of `ParallelBase` (parallel tool calls) with LM-Studio, Qwen, and other providers.

## Root Cause

In `instructor/processing/response.py`, the `isinstance(response_model, ParallelBase)` check correctly identifies parallel responses, but the next line tries to set `_raw_response` on a generator — an unsupported operation.

## Fix

Consume the generator into a `ListResponse` (the same wrapper already used for `IterableBase` results) which properly supports `_raw_response`:

```python
# Before (crashes)
return model

# After (works)
return ListResponse.from_list(list(model), raw_response=response)
```

Applied to both sync (`process_response`) and async (`process_response_async`) paths.

## Testing

- All existing DSL tests pass (12/12)
- `ListResponse` already handles `_raw_response` correctly for `create_with_completion()`

Fixes #2049

### DIFF
--- a/tests/test_list_response_wrapper.py
+++ b/tests/test_list_response_wrapper.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 from collections.abc import AsyncGenerator, Generator
+from types import SimpleNamespace
 
 import pytest
 from pydantic import BaseModel
 
+from instructor.dsl.parallel import ParallelBase
 from instructor.dsl.iterable import IterableBase
 from instructor.dsl.response_list import ListResponse
 from instructor.mode import Mode
@@ -44,6 +46,26 @@ class DummyCompletion(BaseModel):
     """Minimal stand-in for a provider completion object."""
 
 
+class DummyParallelModel(BaseModel):
+    name: str
+
+
+def _make_parallel_completion(*, name: str = "DummyParallelModel", arguments: str = '{"name":"Alice"}'):
+    return SimpleNamespace(
+        choices=[
+            SimpleNamespace(
+                message=SimpleNamespace(
+                    tool_calls=[
+                        SimpleNamespace(
+                            function=SimpleNamespace(name=name, arguments=arguments)
+                        )
+                    ]
+                )
+            )
+        ]
+    )
+
+
 def test_process_response_returns_list_response_for_iterable_model():
     raw = DummyCompletion()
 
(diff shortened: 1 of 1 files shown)


giulio-leone:
Hi @jxnl @ivanleomk — friendly ping. This fixes a crash in `ParallelBase` where the generator wasn't consumed before accessing `_raw_response`, causing an `AttributeError`. The fix wraps the generator output in a `ListResponse` to ensure proper materialization. Would appreciate a review. Thanks!

giulio-leone:
Rebased this branch onto the latest `main` and re-pushed it. It is now `0 behind / 1 ahead` locally.

Local validation I ran:
- `.venv/bin/python -m pytest tests/test_list_response.py -q` -> `3 passed`
- direct branch-vs-main reproduction of the exact `ParallelBase` failure path from #2049

Real proof for the reported bug using a synthetic `PARALLEL_TOOLS` response and `process_response(...)`:
- on `origin/main`: `AttributeError: 'generator' object has no attribute '_raw_response'`\n- on this branch: returns `ListResponse`, `isinstance(result, ListResponse) == True`, first parsed item is `Alice`, and `result.get_raw_response() is raw == True`\n\nSo the current patch fixes the real failure mode without changing the approved behavior: parallel tool responses are materialized into a `ListResponse` wrapper that preserves the provider raw response instead of trying to attach `_raw_response` to a generator.

giulio-leone:
Added a regression-only follow-up commit that exercises the real `ParallelBase` failure path in both sync and async flows.

Local verification on commit `98b8e68`:
- `.venv/bin/ruff check instructor/processing/response.py tests/test_list_response_wrapper.py` — passed twice
- `.venv/bin/pytest tests/test_list_response_wrapper.py tests/test_list_response.py -q` — `9 passed` twice
- direct branch proof: `process_response(...)` and `process_response_async(...)` now both return `ListResponse`, preserve parsed content (`Alice`), and keep `get_raw_response() is raw`
- direct `origin/main` proof in a separate worktree still reproduces the reported `AttributeError: 'generator' object has no attribute '_raw_response'` in both sync and async paths\n\nSo the branch now has both the fix and regression coverage for the exact crash reported in #2049.

PR #2145, opened 2026-03-13
Fix: AttributeError: 'Stream' object has no attribute 'choices' and Real-Time Streaming

ramikhafagi96:
# Fix streaming behavior in `create()` and `create_partial()`

## Summary

This PR fixes two issues in the streaming path:

1.  `create(stream=True)` crashes with\
    `AttributeError: 'Stream' object has no attribute 'choices'`.
2.  `create_partial()` buffers the entire stream into a list, preventing
    real-time streaming of partial models.

The changes enable **true streaming for partial models** and ensure
`create(stream=True)` works correctly.

------------------------------------------------------------------------

### 1. `create(stream=True)` crash

When `stream=True` is used without `Partial`, the OpenAI `Stream` object
is passed directly to `process_response()`.

Execution falls through to `from_response()`, which expects a
`ChatCompletion` object and accesses `completion.choices`. Since
`Stream` is an iterator, this results in:

    AttributeError: 'Stream' object has no attribute 'choices'

------------------------------------------------------------------------

### 2. `create_partial()` disables streaming

`create_partial()` wraps the model with `Partial` and sets
`stream=True`, but `process_response()` consumes the generator:

``` python
return list(
    response_model.from_streaming_response(response, mode=mode)
)
```

Because `list()` eagerly consumes the generator, the entire stream is
buffered before returning. As a result, partial models are not yielded
in real time.

------------------------------------------------------------------------

### Generator passthrough for partial streaming

Return the generator directly instead of wrapping it in `list()`.

Before:

``` python
return list(
    response_model.from_streaming_response(response, mode=mode)
)
```

After:

``` python
return response_model.from_streaming_response(response, mode=mode)
```

This allows callers to iterate over partial models as they arrive.

------------------------------------------------------------------------

### Fallback for `create(stream=True)`

Added `_accumulate_stream()` to consume a raw `Stream` and construct a
synthetic `ChatCompletion` before passing it to `from_response()`.

This prevents the `Stream` → `choices` crash.

------------------------------------------------------------------------

### Retry behavior with streaming

Retry logic is fundamentally incompatible with streaming partial
responses.

When `create_partial()` streams results, partial models may already be
yielded to the caller. If validation fails later in the stream, retrying
the request would require retracting previously yielded results, which
is not possible.

For this reason, when `process_response()` returns a generator
(streaming `Partial` case), `retry_sync` now returns it directly instead
of attempting retry logic.

This preserves existing retry behavior for **non-streaming calls**,
while allowing streaming generators to pass through unchanged.

------------------------------------------------------------------------

## Changes

**response.py** - Return generator for `PartialBase + stream` - Add
`_accumulate_stream()` helper - Add fallback handling for
`create(stream=True)`

**retry.py** - Add `Generator` import - Return generators directly in
`retry_sync`

------------------------------------------------------------------------

## Impact

-   `create_partial()` now streams partial models in real time
-   `create(stream=True)` no longer crashes
-   Async and non-streaming paths remain unchanged

------------------------------------------------------------------------

Tested with **`instructor==1.14.1`**.

### DIFF
--- a/instructor/core/retry.py
+++ b/instructor/core/retry.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Generator
 from json import JSONDecodeError
 from typing import Any, Callable, TypeVar
 
@@ -196,7 +197,7 @@ def retry_sync(
                         response=response, total_usage=total_usage
                     )
 
-                    return process_response(  # type: ignore
+                    result = process_response(  # type: ignore
                         response=response,
                         response_model=response_model,
                         validation_context=context,
@@ -204,6 +205,15 @@ def retry_sync(
                         mode=mode,
                         stream=stream,
                     )
+                    
+                    # When process_response returns a generator (streaming Partial),
+                    # return it directly, as retry is incompatible with streaming
+                    # because we can't un-yield partial models already sent to the caller
+                    if isinstance(result, Generator):
+                        return result # type: ignore
+                    
+                    return result # type: ignore
+                    
                 except (
                     ValidationError,
                     JSONDecodeError,
--- a/instructor/processing/response.py
+++ b/instructor/processing/response.py
@@ -170,6 +170,77 @@ T_ParamSpec = ParamSpec("T_ParamSpec")
 T = TypeVar("T")
 
 
+def _accumulate_stream(stream: Any, mode: Mode) -> ChatCompletion:
+    import json as _json
+    
+    content_parts: list[str] = []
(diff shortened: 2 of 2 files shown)


jxnl:
I don’t have time to take this one over right now. I’m not going to merge it as-is, but I’ll revisit the streaming design later when I can review the retry/reask semantics and add the missing coverage.

jxnl-oai:
Thanks for the streaming fix. #2263 is the stronger current version of this work and adds broader retry handling plus focused regression tests, so I am closing this one as superseded by #2263.

Issue #2146, opened 2026-03-13
Google GenAI SafeSettings change in 1.14.4 breaks Gemini API usage


andersio:
- [x] This is actually a bug report.
- [ ] I am not getting good LLM Results
- [ ] I have tried asking for help in the community on discord or discussions and have not received a response.
- [x] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

gemini-3-flash

**Describe the bug**
The SafetySettings changes in https://github.com/567-labs/instructor/pull/2007/changes are incompatible with the Gemini API.

The Gemini API does not support these `IMAGE_` harm categories as per https://ai.google.dev/api/generate-content#harmcategory.

The `IMAGE_` family of harm categories are only available through Vertex AI API.
https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/HarmCategory

**To Reproduce**
Run Instructor 1.14.4-1.14.5 with google/gemini-3-flash-preview, and attach any kind of image.

Observe Instructor failing with:

```
[...]

PR #2007, opened 2026-01-16
GenAI SafetySettings image content

jxnl:
fix: resolve Google GenAI SafetySettings breaking issue with image content

## Describe your changes

This change resolves a critical `400 INVALID_ARGUMENT` error in Google GenAI (Gemini) integration when `safety_settings` are used with image content. Previously, the system incorrectly sent text-only harm categories, causing the API to reject the request.

The fix involves:
1.  **Dynamic Category Selection**: `update_genai_kwargs` now detects the presence of image content in the request and dynamically switches to using `HARM_CATEGORY_IMAGE_*` categories when appropriate.
2.  **Threshold Mapping**: User-provided text-based safety thresholds (e.g., `HARM_CATEGORY_HATE_SPEECH`) are now correctly mapped to their corresponding image categories (e.g., `HARM_CATEGORY_IMAGE_HATE`).
3.  **Execution Order Correction**: The `handle_genai_tools` function was reordered to ensure that message conversion and multimodal content extraction occur *before* `update_genai_kwargs` is called, allowing for accurate detection of image content.

New unit tests have been added to validate that image content correctly triggers the use of image-specific safety categories and that text-based thresholds are properly applied.

## Issue ticket number and link

567-265, https://github.com/567-labs/instructor/issues/1773

## Checklist before requesting a review

- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.
- [ ] If it is a core feature, I have added documentation.

---
Linear Issue: [567-265](https://linear.app/fivesixseven/issue/567-265/fix-google-genai-safetysettings-breaking-issue-1773)

<a href="https://cursor.com/background-agent?bcId=bc-5fdac5bf-82e0-4456-ac30-4cbe8d4a8f96"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-5fdac5bf-82e0-4456-ac30-4cbe8d4a8f96"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

----

> [!IMPORTANT]
[...]

PR #2147, opened 2026-03-14
fix(gemini): exclude IMAGE_ harm categories from Gemini API safety settings

gambletan:
## Summary

Fixes #2146

`HARM_CATEGORY_IMAGE_*` harm categories (e.g., `HARM_CATEGORY_IMAGE_HATE`, `HARM_CATEGORY_IMAGE_DANGEROUS_CONTENT`) are only valid for the **Vertex AI** API, not the **Gemini API** (`google.genai`). The current code sends these image categories when image content is detected, causing a `400 INVALID_ARGUMENT` error for all Gemini API users processing images since v1.14.4.

## Changes

- **`instructor/providers/gemini/utils.py`**: Exclude all `HARM_CATEGORY_IMAGE_*` categories from the safety settings sent to the Gemini API, alongside the existing exclusions for `HARM_CATEGORY_UNSPECIFIED` and `HARM_CATEGORY_JAILBREAK` (both also Vertex-AI-only). Always use text harm categories regardless of whether image content is present.
- Removed the now-unnecessary `_genai_kwargs_has_image_content` image-detection logic and `_map_text_to_image_category_name` helper from the safety settings path, since image categories are never sent.

## Root cause

PR #2007 added `IMAGE_*` harm categories to the safety settings. These categories exist in the `google.genai` SDK's `HarmCategory` enum but are **rejected by the Gemini API endpoint** — they are only accepted by the Vertex AI endpoint. The previous code selected image categories whenever image content was detected, breaking all Gemini API image workflows.

## Test plan

- Verified that safety settings no longer include any `HARM_CATEGORY_IMAGE_*` entries
- Gemini API calls with image content should no longer return `400 INVALID_ARGUMENT`
- Users who pass custom `safety_settings` as a list (SDK format) are unaffected (passthrough logic is preserved)

### DIFF
--- a/instructor/providers/gemini/utils.py
+++ b/instructor/providers/gemini/utils.py
@@ -373,52 +373,32 @@ def update_genai_kwargs(
         base_config["safety_settings"] = safety_settings
         safety_settings = None
 
-    # Filter out image related harm categories which are not
-    # supported for text based models
-    # Exclude JAILBREAK category as it's only for Vertex AI, not google.genai
+    # Exclude categories not supported by the google.genai Gemini API:
+    # - HARM_CATEGORY_UNSPECIFIED is a sentinel value
+    # - HARM_CATEGORY_JAILBREAK is only for Vertex AI
+    # - HARM_CATEGORY_IMAGE_* categories are only for Vertex AI, NOT the
+    #   Gemini API (google.genai). Sending them causes a 400 INVALID_ARGUMENT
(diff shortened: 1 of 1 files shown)


jxnl:
Duplicate of #2151. I’m keeping one canonical PR open for issue #2146 to avoid split review. Before merge, the surviving branch still needs the safety-settings tests updated to reflect the google.genai API behavior.

PR #2148, opened 2026-03-14
fix(mistral): support mistralai v2.0.0 breaking import change

### LINKED ISSUE #2137: bug: New Mistralai version 2.0.0 released 3h ago breaks instructor

gambletan:
## Summary

Fixes #2137

`mistralai` 2.0.0 (released March 10, 2026) moved the `Mistral` client class from `mistralai.Mistral` to `mistralai.client.Mistral`. This causes an `ImportError` at `import instructor` time that breaks **all** providers — not just Mistral — whenever `mistralai` is installed.

## Changes

| File | Change |
|---|---|
| `instructor/providers/mistral/client.py` | Moved the top-level `from mistralai import Mistral` into a `_import_mistral_class()` helper that tries the v1.x path first, then falls back to `mistralai.client.Mistral` (v2.x). Uses `TYPE_CHECKING` for type annotations. |
| `instructor/providers/__init__.py` | Wrapped the mistral provider import in `try/except ImportError` (matching the existing pattern for `vertexai`) so a broken mistralai install does not prevent other providers from loading. |
| `instructor/auto_client.py` | Added the same v1/v2 try/except fallback for the `Mistral` import in the auto-client provider selection. |

## Compatibility

- **mistralai v1.x**: Uses `from mistralai import Mistral` (existing behavior)
- **mistralai v2.0.0+**: Falls back to `from mistralai.client import Mistral`
- **No mistralai installed**: No change — the provider is simply not registered

## Test plan

- Verified import works with both v1.x and v2.x import paths
- Verified that `import instructor` succeeds even when mistralai is installed but has the new package structure
- Other providers remain unaffected

### DIFF
--- a/instructor/auto_client.py
+++ b/instructor/auto_client.py
@@ -519,7 +519,10 @@ def from_provider(
 
     elif provider == "mistral":
         try:
-            from mistralai import Mistral
+            try:
+                from mistralai import Mistral  # v1.x
+            except ImportError:
+                from mistralai.client import Mistral  # v2.x  # type: ignore[no-redef]
             from instructor import from_mistral  # type: ignore[attr-defined]
             import os
 


jxnl:
Duplicate of #2152. This branch is close, but I’m keeping #2152 as the canonical PR for issue #2137 because it also updates the dependency constraint for declared mistralai v2 compatibility.

conjon42:
Hello @jxnl, as far as I can tell, PR #2152 is no longer part of this repository. If it was scrapped, could we re-visit this or other PRs that would make this compatible with Mistral's v2 API? Thank you!

PR #2163, opened 2026-03-17
feat: add Model Armor support for GenAI content blocking

edge7:
- Add ContentBlockedError exception for blocked responses                                                                                                                                                                         
- Check empty candidates in parse_genai_tools, parse_genai_structured_outputs, parse_gemini_json                                                                                                                                  
- Skip retry on ContentBlockedError (safety blocks are not retriable)                                                                                                                                                             
- Propagate model_armor_config through GenerateContentConfig

### DIFF
--- a/instructor/core/exceptions.py
+++ b/instructor/core/exceptions.py
@@ -599,3 +599,15 @@ class MultimodalError(ValueError, InstructorError):
             context_parts.append(f"file: {file_path}")
         context = f" ({', '.join(context_parts)})" if context_parts else ""
         super().__init__(f"{message}{context}", *args, **kwargs)
+
+
+class ContentBlockedError(InstructorError):
+    """Raised when the response is blocked by safety filters or Model Armor."""
+
+    def __init__(self, block_reason=None, block_message=None):
+        self.block_reason = block_reason
+        self.block_message = block_message
+        msg = f"Response blocked: {block_reason or 'unknown reason'}"
+        if block_message:
+            msg += f" - {block_message}"
+        super().__init__(msg)
--- a/instructor/core/retry.py
+++ b/instructor/core/retry.py
@@ -249,6 +249,10 @@ def retry_sync(
                     )
                     raise e
                 except Exception as e:
+                    from .exceptions import ContentBlockedError
+
+                    if isinstance(e, ContentBlockedError):
+                        raise  # No retry in case of Content Blocked
                     # Emit completion:error for non-validation errors (API errors, network errors, etc.)
                     logger.debug(f"Completion error: {e}")
                     hooks.emit_completion_error(e)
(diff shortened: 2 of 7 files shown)


edge7:
**Summary**                                                                                                                                                                                                                           
                                                                                                                                                                                                                                    
  _What is Model Armor?_                                          
                                                                                                                                                                                                                                    
  https://cloud.google.com/security/products/model-armor is a Google Cloud service that integrates with Vertex AI to screen both prompts and responses for safety violations (harassment, dangerous content, prompt injection,      
  jailbreak attempts, etc.). It can be configured via floor settings (project-wide minimum thresholds) or templates (per-request fine-grained control).                                                                             
                                                                                                                                                                                                                                    
  When Model Armor blocks a request, the Vertex AI generateContent API returns a response with empty candidates and a promptFeedback containing the block reason:                                                                   
   
```
  {                                                                                                                                                                                                                                 
    "promptFeedback": {                                         
      "blockReason": "MODEL_ARMOR",
      "blockReasonMessage": "Blocked by Floor Setting. The prompt violated..."                                                                                                                                                      
    },                                                                                                                                                                                                                              
    "candidates": []                                                                                                                                                                                                                
  }      
```                                                                                                                                                                                                                           
                                                                
  The problem

  Instructor currently does not handle this case. When candidates are empty:                                                                                                                                                        
   
  - parse_genai_tools() hits assert len(candidates) == 1 → cryptic AssertionError                                                                                                                                                   
  - parse_genai_structured_outputs() accesses completion.text → ValueError
  - parse_gemini_json() crashes twice: first on completion.text, then the error handler itself tries to access candidates[0]                                                                                                        
                                                                                                                                                                                                                                    
  In all cases, the block reason is lost, and the retry loop retries the same prompt N times — which will always be blocked again.                                                                                                  
                                                                                                                                                                                                                                    
  Additionally, users cannot pass model_armor_config (template names) through Instructor because it is silently dropped during config merging.                                                                                      
                                                                
  What this PR does                                                                                                                                                                                                                 
                                                                
  1. New ContentBlockedError exception — extends InstructorError, carries block_reason and block_message from the API response                                                                                                      
  2. Early detection in all 3 GenAI parsers — checks for empty candidates before accessing response content, raising ContentBlockedError with full context
  3. No retry on content blocks — safety blocks are permanent, not transient validation errors. The retry loop now immediately re-raises ContentBlockedError instead of wasting retries                                             
  4. model_armor_config propagation — added to config_fields_to_merge so users can pass template names through GenerateContentConfig

jxnl:
Closing this for scope. Provider/vendor additions like Model Armor should be handled externally rather than carried in this backlog.

edge7:
@jxnl can you clarify? what do you mean?
Lots of companies will integrate Model Armor in their Workflow. You should see it as a Vertex Safety Setting extension

PR #2169, opened 2026-03-18
fix: handle ANTHROPIC_REASONING_TOOLS in parse dispatch

frankgoldfish:
## Summary

- **Bug**: In `OpenAISchema.from_response()`, there were two consecutive `if` checks for `ANTHROPIC_TOOLS`. The first check (on its own) matched and returned immediately, making the second check (which also covered `ANTHROPIC_REASONING_TOOLS`) unreachable dead code. As a result, calls with `mode=Mode.ANTHROPIC_REASONING_TOOLS` would never be dispatched to `parse_anthropic_tools` and would fall all the way through to a `ConfigurationError: Invalid or unsupported mode`.
- **Fix**: Combine both conditions into a single `if mode in {Mode.ANTHROPIC_TOOLS, Mode.ANTHROPIC_REASONING_TOOLS}:` branch, ensuring both modes correctly dispatch to `parse_anthropic_tools`.

## Affected code

**Before** (`instructor/processing/function_calls.py`):
```python
if mode == Mode.ANTHROPIC_TOOLS:
    return cls.parse_anthropic_tools(completion, validation_context, strict)

# Dead code — ANTHROPIC_TOOLS already returned above; ANTHROPIC_REASONING_TOOLS never reached
if mode == Mode.ANTHROPIC_TOOLS or mode == Mode.ANTHROPIC_REASONING_TOOLS:
    return cls.parse_anthropic_tools(completion, validation_context, strict)
```

**After**:
```python
if mode in {Mode.ANTHROPIC_TOOLS, Mode.ANTHROPIC_REASONING_TOOLS}:
    return cls.parse_anthropic_tools(completion, validation_context, strict)
```

## Test plan

- [ ] Call `from_response` with `mode=Mode.ANTHROPIC_TOOLS` — should still work as before.
- [ ] Call `from_response` with `mode=Mode.ANTHROPIC_REASONING_TOOLS` — previously raised `ConfigurationError`, now correctly dispatches to `parse_anthropic_tools`.

### DIFF
--- a/instructor/processing/function_calls.py
+++ b/instructor/processing/function_calls.py
@@ -156,10 +156,7 @@ class OpenAISchema(BaseModel):
             cls (OpenAISchema): An instance of the class
         """
 
-        if mode == Mode.ANTHROPIC_TOOLS:
-            return cls.parse_anthropic_tools(completion, validation_context, strict)
-
-        if mode == Mode.ANTHROPIC_TOOLS or mode == Mode.ANTHROPIC_REASONING_TOOLS:
+        if mode in {Mode.ANTHROPIC_TOOLS, Mode.ANTHROPIC_REASONING_TOOLS}:
             return cls.parse_anthropic_tools(completion, validation_context, strict)
 
         if mode == Mode.ANTHROPIC_JSON:


jxnl:
Duplicate of #2140. This is the same ANTHROPIC_REASONING_TOOLS dispatch fix, and #2140 is already reviewed and approved.

PR #2173, opened 2026-03-18
Fix GenAI SafeSettings regression: always use text harm categories

joaquinhuigomez:
Fixes a regression where GenAI (non-Vertex) requests with image content crash with InvalidArgument because the safety settings code was sending Vertex-only HARM_CATEGORY_IMAGE_* categories to the GenAI API. The fix ensures text harm categories are always used for GenAI calls regardless of content type.

### DIFF
--- a/instructor/providers/gemini/utils.py
+++ b/instructor/providers/gemini/utils.py
@@ -304,66 +304,6 @@ def update_genai_kwargs(
             if val is not None:  # Only set if value is not None
                 base_config[gemini_key] = val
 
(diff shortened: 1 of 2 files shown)


joaquinhuigomez:
Ping — this is still relevant on the latest release. Any chance of a review?

jxnl:
Closing in favor of #2174, which was included in the v1.15.0 staging branch (#2231). Both fix the same Gemini IMAGE_ harm category issue, but #2174 is cleaner (fully removes the dead variable, updates more test files).

PR #2200, opened 2026-03-20
Support PEP 604 unions in Partial models

### LINKED ISSUE #2088: feat(StreamingWithUnionType): Add support for streaming BaseModel with union Types

LittleChenLiya:
## Summary

Support PEP 604 union annotations like `str | int` when building `Partial[...]` models.

`Partial[MyResponse]` currently fails for fields annotated with Python 3.10+ union syntax because the partial model builder tries to subscript `types.UnionType` directly.

## Changes

- add a shared helper to rebuild generic annotations
- special-case PEP 604 unions through `typing.Union` instead of subscripting `types.UnionType`
- add a regression test covering `Partial[...]` on a model with `str | int`

## Testing

- `python -m pytest tests/dsl/test_partial.py -k "pep604_union_annotations or union_literal_accepts_valid_values"`

Closes #2088

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -259,6 +259,18 @@ def _process_generic_arg(
             return arg
 
 
+def _rebuild_generic_annotation(
+    generic_base: Any,
+    generic_args: tuple[Any, ...],
+) -> Any:
+    # `types.UnionType` (the runtime origin of `str | int`) cannot be
+    # subscripted directly, so reconstruct unions through `typing.Union`.
+    if generic_base in UNION_ORIGINS:
+        return Union[generic_args]  # type: ignore[arg-type]
+
+    return generic_base[generic_args]
+
+
 def _make_field_optional(
     field: FieldInfo,
 ) -> tuple[Any, FieldInfo]:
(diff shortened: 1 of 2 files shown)


jxnl:
Already applied — PEP 604 union handling via `UNION_ORIGINS` and `_rebuild_generic_annotation` is already in `dsl/partial.py`. Closing.

PR #2202, opened 2026-03-21
Allow MD_JSON in from_bedrock

### LINKED ISSUE #2076: Allow MD_JSON for bedrock client

**Additional context**
Using version = "1.14.5"

LittleChenLiya:
## Summary

Allow `Mode.MD_JSON` when using the Bedrock client helpers.

The core response processing layer already supports `MD_JSON`, but `from_bedrock` rejected it at mode validation time. This prevented Bedrock clients from using the same markdown-JSON handling path as other providers.

## Changes

- add `Mode.MD_JSON` to the allowed modes in `from_bedrock`
- add a regression test for direct `from_bedrock(..., mode=Mode.MD_JSON)`
- add a regression test for `from_provider("bedrock/...", mode=Mode.MD_JSON)`

## Testing

- `python -m pytest tests/llm/test_bedrock/test_client.py`

Closes #2076

### DIFF
--- a/instructor/providers/bedrock/client.py
+++ b/instructor/providers/bedrock/client.py
@@ -51,6 +51,7 @@ def from_bedrock(
     valid_modes = {
         instructor.Mode.BEDROCK_TOOLS,
         instructor.Mode.BEDROCK_JSON,
+        instructor.Mode.MD_JSON,
     }
 
     if mode not in valid_modes:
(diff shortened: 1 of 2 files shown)


jxnl:
Leaving open with requested changes: the happy-path works, but reask_md_json crashes on Bedrock dict responses (AttributeError on response.choices[0]) when retries fire. Needs reask_md_json to handle the Bedrock dict response shape before this can merge.

jxnl:

PR #2204, opened 2026-03-21
Preserve default literal fields during partial streaming

### LINKED ISSUE #2054: Streaming with Literal["Value"] = "value"

LittleChenLiya:
## Summary

Preserve explicit default literal fields in partial streaming responses.

When building partial objects during streaming, missing fields were always filled with `None`, which caused known defaults like `type: Literal["Person"] = "Person"` to disappear until the model emitted them explicitly at the end of the stream.

## Changes

- keep explicit field defaults when constructing incomplete partial objects
- preserve streaming behavior for required fields and nested models
- add a regression test covering a default `Literal[...]` field during partial streaming

## Testing

- `python -m pytest tests/dsl/test_partial.py -k "literal_default_is_available_during_streaming or partial_with_default_factory or partial_validates_incremental_streaming_data"`

Closes #2054

### DIFF
--- a/instructor/dsl/partial.py
+++ b/instructor/dsl/partial.py
@@ -30,7 +30,7 @@ from typing import (  # noqa: UP035
 
 from jiter import from_json
 from pydantic import BaseModel, create_model
-from pydantic.fields import FieldInfo
+from pydantic.fields import FieldInfo, PydanticUndefined
 
 from instructor.mode import Mode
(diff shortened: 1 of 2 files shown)


jxnl:
Already applied — partial streaming already uses `PydanticUndefined` to preserve field defaults including `Literal` fields. Closing.

PR #2206, opened 2026-03-23
init: Use lazy loading when importing instructor

AbdealiLoKo:
Use lazy module imports to reduce the time and RAM usage of instructor when just instructor was imported.
Fixes https://github.com/567-labs/instructor/issues/2205


jxnl:
Closing — lazy loading is a meaningful improvement but carries regression risk for a patch release. Deferring to 1.16.0.

PR #2208, opened 2026-03-23
feat(client): add two_step and two_step_async pipeline helpers

### LINKED ISSUE #2201: Add model agnostic support for reasoning followed by structured output
Instructor can produce structured output in one pass, but some models and providers do not support native thinking **followed** by structured response in the same call to Instructor. In those cases, users have to hand-roll a two-call workflow: one call for reasoning/planning and a second call for the final structured answer. That is easy to get wrong and duplicates unnecessary code. 

this would currently have to be implemented like:
```python
# --- Models ---
class ReasoningStep(BaseModel):
    reasoning: str

class Answer(BaseModel):
    final_answer: str

# --- LLM setup ---
llm = Llama( )

client = instructor.patch(
    create=llm.create_chat_completion_openai_v1,
    mode=instructor.Mode.JSON_SCHEMA,
)

# --- Step 1: reasoning / planning ---
reasoning_result = client(
    messages=[
        {"role": "system", "content": "Think step-by-step and explain your reasoning."},
        {"role": "user", "content": "Why is the sky blue?"},
    ],
    response_model=ReasoningStep,
    max_tokens=256,  # separate budget for thinking
)

# --- Step 2: final structured answer ---
final_result = client(
    messages=[
        {"role": "system", "content": "Use the previous reasoning as context. Return only the final answer."},
        {"role": "user", "content": "Question: Why is the sky blue?"},
        {"role": "assistant", "content": reasoning_result.reasoning},
        {"role": "user", "content": "Now produce the final structured answer."},
    ],
    response_model=Answer,
    max_tokens=128, # separate budget for final answer
)

print(final_result.final_answer)
```

Add a first-class two-step pipeline helper that lets users separate a “thinking” pass from the final structured output pass. In addition this can help separate reasoning token count from response token count.

A possible API could look like this:
```python
class ReasoningPlan(BaseModel):
    reasoning: str

class Answer(BaseModel):
    final_answer: str

result = instructor.two_step(
    client=client,
    messages=[{"role": "user", "content": "Explain why the sky is blue."}],
    think_model=ReasoningPlan,
    final_model=Answer,
    think_max_tokens=400,
    final_max_tokens=50,
)
```

1. [**NOT A REAL SOLUTION**] Using a single schema with a reasoning or chain_of_thought field. This might seem like a solution at first **BUT** JSON key value pairs by definition are not ordered and there is no guarantee the model would think before giving the final structured answer. 
```python
llama = llama_cpp.Llama()

create = instructor.patch(
    create=llama.create_chat_completion_openai_v1,
    mode=instructor.Mode.JSON_SCHEMA,
)

class Answer(BaseModel):
    reasoning: str
    final_answer: str

result = create(
    messages=[
        {"role": "system", "content": "Think carefully, then fill the schema."},
        {"role": "user", "content": "Answer the question: ..."},
    ],
    response_model=Answer,
)
```

3. Prompting the model to “think first, then answer.” This is useful, but it is only a soft instruction and does not provide a reusable abstraction.

5. Manually writing a two-call pipeline in application code. This is the current workaround, but it is repetitive and inconsistent across implementations.

7. Using provider-specific thinking features where available (e.g. Anthropic seems to support this natively). This is not usable with other models.

Alex-Resch:
## Summary
Adds two_step and two_step_async as standalone functions for a 
model-agnostic two-step reasoning pipeline for models without 
native thinking support.

## Changes
- Add two_step and two_step_async functions in client.py
- Add tests in tests/test_two_step.py

Closes #2201

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -29,6 +29,14 @@ Functions to create Instructor clients from various providers.
 
 ::: instructor.from_litellm
 
+## Two-Step Pipeline
+
+Helper functions for separating reasoning and structured output into two LLM calls.
+
+::: instructor.two_step
+
+::: instructor.two_step_async
+
 ## DSL Components
 
 Domain-specific language components for advanced patterns and data handling.
--- a/instructor/__init__.py
+++ b/instructor/__init__.py
@@ -25,6 +25,8 @@ from .core.client import (
     AsyncInstructor,
     from_openai,
     from_litellm,
+    two_step,
+    two_step_async,
 )
 from .core import hooks
 from .utils.providers import Provider
@@ -43,6 +45,8 @@ __all__ = [
     "from_openai",
     "from_litellm",
     "from_provider",
+    "two_step",
+    "two_step_async",
     "AsyncInstructor",
     "Provider",
     "OpenAISchema",
--- a/instructor/core/client.py
+++ b/instructor/core/client.py
@@ -28,6 +28,8 @@ from .hooks import Hooks, HookName
 
 
 T = TypeVar("T", bound=Union[BaseModel, "Iterable[Any]", "Partial[Any]"])
+T_think = TypeVar("T_think", bound=BaseModel)
+T_final = TypeVar("T_final", bound=BaseModel)
 
 
 class Response:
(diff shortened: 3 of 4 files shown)


jxnl:
Closing — two_step/two_step_async pipeline helpers are a nice addition but belong in 1.16.0 rather than a patch release.

Alex-Resch:
Got it, thanks. Happy to pick this back up for 1.16.0.

PR #2211, opened 2026-03-24
fix(deps): make diskcache an optional dependency

### LINKED ISSUE #2101: (diskcache) dependency related Vulnerability scan issue CVE-2025-69872

RahulModugula:
Closes #2101

Moves `diskcache` from hard dependency to optional in `pyproject.toml`. The import is already lazy via `_import_diskcache()`, so this just removes the unnecessary hard requirement.

Users who need disk caching can install it with:
```
pip install "instructor[diskcache]"
```

This addresses the CVE concern mentioned in the issue by allowing users to avoid installing diskcache if they don't need caching.


richardwadsworth:
hi @jxnl.  This is confusing.  I see this PR is referenced as a duplicate of [PR #2103](https://github.com/567-labs/instructor/pull/2103).  But both are now closed and [Issue 2101](https://github.com/567-labs/instructor/issues/2101) is still open.

Issue #2056, opened 2026-02-04
Security: Retry amplification and LLM validator injection findings


swarmbrr:
Hi,

I conducted a security audit of instructor focusing on attack surfaces when processing untrusted LLM outputs. Found two issues worth sharing:

### 1. Retry Amplification (Medium severity)

**Location**: `instructor/core/retry.py`, `instructor/core/patch.py`

The retry mechanism lacks rate limiting. An adversarial LLM (or prompt-injected response) can craft outputs that always fail validation, causing:
- Each retry appends 2 messages (assistant + tool error)
- Context grows exponentially
- 10 retries = 506x context growth in my PoC

**Impact**: Token budget exhaustion, cost amplification.

**Suggested mitigation**: Optional token budget limit in `max_retries` config, alongside existing `timeout`.

### 2. LLM Validator Injection (Medium severity)

**Location**: `instructor/validation/llm_validators.py:50-66`

User values are interpolated directly into validation prompts without escaping:

```python
{"role": "user", "content": f"Does `{v}` follow the rules: {statement}"}
```

If `v` contains prompt injection payloads, the validator LLM may be manipulated to return `is_valid: true` for invalid data.

**Suggested mitigation**: Explicit delimiters or structured input format for user values.

---

Happy to share the full audit document and PoC if helpful. No rush on response—I'm following a 30-day disclosure window before publishing findings publicly.

Let me know if you'd like more details on either finding.

xXMrNidaXx:
Thanks for the detailed security audit - these are both valid concerns that affect anyone using LLM validation in production.

**On Retry Amplification:**
In addition to token budget limits, consider:
- Exponential backoff between retries (not just linear)
- Context truncation strategy (keep only last N messages on retry)
- Early termination if error messages are repetitive (same error 3x = give up)

**On LLM Validator Injection:**
For production systems using LLM validation:
```python
# Instead of string interpolation:
{"role": "user", "content": f"Does '{v}' follow the rules: {statement}"}

# Consider structured format with clear delimiters:
{"role": "user", "content": f"""
<VALIDATION_REQUEST>
<VALUE>{html.escape(str(v))}</VALUE>
<RULES>{statement}</RULES>
</VALIDATION_REQUEST>
Does the VALUE follow the RULES? Answer only 'valid' or 'invalid'.
"""}
```

Also worth considering:
- Separate, isolated model for validation (lower capability = lower injection risk)
- Regex pre-validation before LLM validation for obvious bad inputs
[...]

PR #2215, opened 2026-03-25
fix(validators): make allow_override work in llm_validator

Krishnachaitanyakc:
## Summary

- Fix dead code in `llm_validator` where the `allow_override` parameter could never take effect
- The `assert resp.is_valid` on line 69 always fired before the `allow_override` check on line 71, making the override path unreachable
- Restructure the logic so that when `allow_override=True` and the LLM provides a `fixed_value`, the corrected value is returned instead of raising `AssertionError`

## Problem

```python
# Before (broken): assert always fires first, override is dead code
assert resp.is_valid, resp.reason                          # <-- raises here
if allow_override and not resp.is_valid and resp.fixed_value is not None:
    return resp.fixed_value  # <-- never reached
```

```python
# After (fixed): check override before asserting
if not resp.is_valid:
    if allow_override and resp.fixed_value is not None:
        return resp.fixed_value
```

## Test plan

- [x] `test_valid_value_returns_original` -- valid input returns original value
- [x] `test_invalid_without_override_raises` -- invalid input without override raises AssertionError
- [x] `test_invalid_with_override_returns_fixed_value` -- invalid input with override returns LLM-suggested fix
- [x] `test_invalid_with_override_but_no_fixed_value_raises` -- invalid input with override but no fix still raises
- [x] `test_valid_value_with_override_returns_original` -- valid input with override enabled returns original

```bash
python -m pytest tests/test_llm_validator_allow_override.py -v
```

Relates to #2056

### DIFF
--- a/instructor/validation/llm_validators.py
+++ b/instructor/validation/llm_validators.py
@@ -64,13 +64,14 @@ def llm_validator(
             temperature=temperature,
         )
 
-        # If the response is  not valid, return the reason, this could be used in
-        # the future to generate a better response, via reasking mechanism.
-        assert resp.is_valid, resp.reason
+        # If the value is not valid but we allow overrides and the LLM
+        # suggested a corrected value, return the fixed value instead of
+        # raising an assertion error.
+        if not resp.is_valid:
+            if allow_override and resp.fixed_value is not None:
+                return resp.fixed_value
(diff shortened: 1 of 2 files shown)


jxnl:
Already applied — `llm_validator` already checks `allow_override` before asserting. Closing.

PR #2216, opened 2026-03-25
fix: consume ParallelBase generator into ListResponse to avoid _raw_response crash

### LINKED ISSUE #2049: bug in ParallelTools call, model: Qwen3-VL-4b-instruct, server: LM-Studio 0.4.1

giulio-leone:
## Problem

`ParallelBase.from_response()` is a generator function (uses `yield`), so calling `response_model.from_response(...)` returns a generator object. Both `process_response()` and `process_response_async()` then attempt:

```python
model._raw_response = response  # model is a generator → AttributeError
```

Generators don't support setting arbitrary attributes, causing:

```
```

This breaks any use of `ParallelBase` (parallel tool calls) with LM-Studio, Qwen, and other providers.

## Root Cause

In `instructor/processing/response.py`, the `isinstance(response_model, ParallelBase)` check correctly identifies parallel responses, but the next line tries to set `_raw_response` on a generator — an unsupported operation.

## Fix

Consume the generator into a `ListResponse` (the same wrapper already used for `IterableBase` results) which properly supports `_raw_response`:

```python
# Before (crashes)
return model

# After (works)
return ListResponse.from_list(list(model), raw_response=response)
```

Applied to both sync (`process_response`) and async (`process_response_async`) paths.

## Testing

- All existing DSL tests pass (12/12)
- `ListResponse` already handles `_raw_response` correctly for `create_with_completion()`

Fixes #2049

> ⚠️ This reopens #2143 which was accidentally closed due to fork deletion.

### DIFF
--- a/tests/test_list_response_wrapper.py
+++ b/tests/test_list_response_wrapper.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 from collections.abc import AsyncGenerator, Generator
+from types import SimpleNamespace
 
 import pytest
 from pydantic import BaseModel
 
+from instructor.dsl.parallel import ParallelBase
 from instructor.dsl.iterable import IterableBase
 from instructor.dsl.response_list import ListResponse
 from instructor.mode import Mode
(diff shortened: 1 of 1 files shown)


jxnl:
Applied directly in staging/v1.15.1 — ParallelBase generator is now consumed into a ListResponse (same as IterableBase), avoiding the AttributeError when setting _raw_response on a generator. Closing.

PR #2223, opened 2026-03-28
feat(hooks): expose retry metadata on completion:error and completion:last_attempt hooks

### LINKED ISSUE #2222: Expose attempt metadata on completion:error and completion:last_attempt hooks
## Summary

- the final failed attempt

## Current behavior

Hook signatures appear to be:

## Desired behavior

## Use case

## Notes

- `failed_attempts`

majiayu000:
Fixes #2222

`emit_completion_error` and `emit_completion_last_attempt` now forward `attempt_number`, `max_attempts`, and `is_last_attempt` as keyword arguments to registered handlers.

Handlers that accept `**kwargs` (or the specific keyword params) get the metadata. Old-style handlers like `def on_error(error): ...` keep working as before, no warnings. This is done via `inspect.signature().bind()` in `emit()` to check compatibility at call time.

Changes in `retry.py`: the four emission sites (sync/async x validation/non-validation) now compute `is_last_attempt` before emitting and pass all three kwargs. `max_attempts` comes from `getattr(max_retries.stop, 'max_attempt_number', None)` so it's `None` for custom stop conditions.

Tests in `tests/test_hook_metadata.py` cover metadata forwarding, backward compat with old handlers, and handlers that accept specific kwargs.

### DIFF
--- a/instructor/cli/batch.py
+++ b/instructor/cli/batch.py
@@ -15,9 +15,11 @@ app = typer.Typer()
 console = Console()
 
 
-def generate_table(batch_jobs: list[BatchJobInfo], provider: str, full_id: bool = False):
+def generate_table(
+    batch_jobs: list[BatchJobInfo], provider: str, full_id: bool = False
+):
     """Generate enhanced table for batch jobs using unified BatchJobInfo objects
-    
+
     Args:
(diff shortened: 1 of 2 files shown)


jxnl:

PR #2226, opened 2026-03-28
fix: change Validator.is_valid default to False (fail-closed)

### LINKED ISSUE #2225: llm_validator can fail open when validator response omits is_valid
## Summary

`llm_validator` can fail open when the validator model returns partial JSON that omits `is_valid`.

The internal `Validator` currently defaults `is_valid=True` in [`instructor/processing/validators.py`](https://github.com/567-labs/instructor/blob/main/instructor/processing/validators.py#L10-L27). Because of that, a response like:

```json
{"reason": "value violates the rule"}
```

can parse successfully as valid, so the semantic check may silently pass even though the model indicated failure in `reason`.

## Why this matters

With weaker or local models, especially via Ollama / OpenAI-compatible endpoints, partial structured outputs are common. In that setup, `llm_validator` is not reliably fail-closed.

Observed validator outputs included:

```json
{"reason": "..."}
{"error": "..."}
{"prompt": "..."}
{"subject": "...", "action": "...", "object": "..."}
```

The malformed outputs may parse-fail and retry, but the partial `{"reason": "..."}` case is more concerning because it can parse and default to `is_valid=True`.

## Current behavior

[`llm_validator`](https://github.com/567-labs/instructor/blob/main/instructor/validation/llm_validators.py#L9-L76) requests `response_model=Validator` and then does:

```python
```

Relevant lines:
- validator model: [`validators.py#L10-L27`](https://github.com/567-labs/instructor/blob/main/instructor/processing/validators.py#L10-L27)
- request + assert: [`llm_validators.py#L50-L73`](https://github.com/567-labs/instructor/blob/main/instructor/validation/llm_validators.py#L50-L73)

If `is_valid` is omitted and defaulted, invalid values can be accepted.

## Expected behavior

If the validator response omits `is_valid`, validation should fail closed rather than defaulting to valid.

## Suggested fix

One of these would solve it:

1. Make `is_valid` required with no default
2. Treat missing `is_valid` as a parsing / validation failure
3. Use a stricter internal validator model for `llm_validator`

## Reproduction

I do not have a minimal standalone repro yet, but this is easy to hit with weaker local models used as semantic validators.

A typical problematic response is:

```json
{"reason": "contains multiple events and abstract descriptions"}
```

That should not be interpreted as valid.

## Environment

- Instructor: 1.14.4
- Provider: local OpenAI-compatible endpoint at `http://REDACTED/v1`
- Validator model: `qwen3.5`
- Runtime: Ollama-backed local model

tranhoangtu-it:
## Summary
Change `Validator.is_valid` default from `True` to `False` so that when the LLM response omits `is_valid`, validation fails safely instead of silently passing.

## Problem
```python
# Before: default=True → fail-open
# LLM returns {"reason": "violates rule"} without is_valid
# → is_valid defaults to True → validation passes silently
```

## Fix
```diff
- is_valid: bool = Field(default=True, ...)
+ is_valid: bool = Field(default=False, ...)
```

## Test plan
- [ ] LLM response with `is_valid=True` → passes (no change)
- [ ] LLM response with `is_valid=False` → fails (no change)
- [ ] LLM response omitting `is_valid` → now fails (was silently passing)

Closes #2225

### DIFF
--- a/instructor/processing/validators.py
+++ b/instructor/processing/validators.py
@@ -14,7 +14,7 @@ class Validator(OpenAISchema):
     """
 
     is_valid: bool = Field(
-        default=True,
+        default=False,
         description="Whether the attribute is valid based on the requirements",
     )
     reason: Optional[str] = Field(


jxnl:
Closing as duplicate of #2230, which addresses the same Validator.is_valid default change.

PR #2227, opened 2026-03-28
fix: make Validator.is_valid default to False (fail-closed)

### LINKED ISSUE #2225: llm_validator can fail open when validator response omits is_valid

nightcityblade:
Fixes #2225

## Problem

`Validator.is_valid` defaults to `True`, so if the LLM omits `is_valid` from its response (e.g. returning only `{"reason": "value violates the rule"}`), the validation silently passes — a fail-open behavior.

## Fix

Change `is_valid` default from `True` to `False`. Now if the model omits the field, validation fails-closed, which is the safe default for a security/validation check.

This is a one-line change in `instructor/processing/validators.py`.

### DIFF
--- a/instructor/processing/validators.py
+++ b/instructor/processing/validators.py
@@ -14,7 +14,7 @@ class Validator(OpenAISchema):
     """
 
     is_valid: bool = Field(
-        default=True,
+        default=False,
         description="Whether the attribute is valid based on the requirements",
     )
     reason: Optional[str] = Field(


aress31:
I don't think that just switching the default returned value from `True` to `False` will be a real fix for this issue. 

It would be interesting to understand why the validator don't return the expected shape and fix it too for a reliable fix...

jxnl:
Closing as duplicate of #2230, which addresses the same Validator.is_valid default change.

PR #2174, opened 2026-03-19
fix: exclude IMAGE_ harm categories from standard Gemini API

gambletan:
## Summary

Fixes #2146

- `IMAGE_` family harm categories (e.g. `HARM_CATEGORY_IMAGE_HATE`) added in v1.14.4 (PR #2007) are only supported by the **Vertex AI** API, not the standard **Gemini API**
- Sending them to the standard Gemini API causes `400 INVALID_ARGUMENT` errors: `Invalid value at safety_settings[0].category, "HARM_CATEGORY_IMAGE_HATE"`
- Since `update_genai_kwargs()` is only called from GenAI (non-Vertex) code paths (`handle_genai_tools` and `handle_genai_structured_outputs`), always filter out `IMAGE_` categories
- Removed the now-unused `_genai_kwargs_has_image_content()` helper and image-category mapping logic
- Updated tests to verify `IMAGE_` categories are never included in standard Gemini API safety settings

## Test plan

- [x] `tests/genai/test_safety_settings.py` - 3 tests pass (updated to verify IMAGE_ exclusion)
- [x] `tests/test_genai_config_merging.py` - 21 tests pass (no regressions)
- [ ] Manual test with standard Gemini API to confirm no 400 errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2233, opened 2026-03-28
fix: exclude Vertex AI-only IMAGE_ harm categories from Gemini API

IgnazioDS:
## Summary
- Filter out `IMAGE_` harm categories (e.g., `HARM_CATEGORY_IMAGE_HATE`) when using the standard Gemini API
- These categories are Vertex AI-only and cause 400 errors on the Gemini API
- Remove the now-unused `_genai_kwargs_has_image_content` helper function
- Update tests to verify IMAGE_ categories are excluded from Gemini API calls

## Problem
Since PR #2007 (v1.14.4), all Gemini API calls that trigger safety settings fail with a 400 error because `update_genai_kwargs` includes Vertex AI-only `IMAGE_` harm categories. The `update_genai_kwargs` function is only called from the GenAI (non-Vertex) code path, so these categories should never be included. See #2146.

## Root Cause
PR #2007 added logic to switch between text and image harm categories based on whether the request contained image content. However, `IMAGE_` harm categories (`HARM_CATEGORY_IMAGE_HATE`, `HARM_CATEGORY_IMAGE_HARASSMENT`, etc.) are only supported by the Vertex AI API. The standard Gemini API rejects them with a 400 `INVALID_ARGUMENT` error.

## Fix
- Always use text-only categories in `update_genai_kwargs` since it is exclusively called from the Gemini API (non-Vertex) code path
- Vertex AI handlers (`handle_vertexai_tools`, `handle_vertexai_json`, etc.) have their own separate processing and are unaffected

## Test plan
- [x] Updated `tests/genai/test_safety_settings.py` to verify IMAGE_ categories are excluded
- [x] Updated `tests/llm/test_genai/test_utils.py` to verify IMAGE_ categories are excluded even with image content
- [ ] Gemini API calls no longer fail with 400 due to unsupported harm categories
- [ ] Vertex AI calls are unaffected (separate code path)

Fixes #2146

### DIFF
--- a/instructor/providers/gemini/utils.py
+++ b/instructor/providers/gemini/utils.py
@@ -304,66 +304,6 @@ def update_genai_kwargs(
             if val is not None:  # Only set if value is not None
                 base_config[gemini_key] = val
 
-    def _genai_kwargs_has_image_content(genai_kwargs: dict[str, Any]) -> bool:
-        """
-        Best-effort check for image content in a GenAI request.
-
-        We use this to decide whether to send text vs image harm categories in
-        `safety_settings`. The google-genai SDK has separate image categories
-        (e.g., `HARM_CATEGORY_IMAGE_HATE`) which are required for image content.
-        """
-        # Prefer typed GenAI contents if present (works with autodetect_images)
-        contents = genai_kwargs.get("contents")
-        if isinstance(contents, list):
-            for content in contents:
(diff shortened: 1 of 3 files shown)


jxnl:
Closing as duplicate — #2174 (already merged into staging/v1.15.0) fixes the same Gemini IMAGE_ harm category issue with a cleaner approach.

PR #2251, opened 2026-04-09
test(hooks): add unit tests for completion:error attempt metadata for…

### LINKED ISSUE #2222: Expose attempt metadata on completion:error and completion:last_attempt hooks

RudrenduPaul:
Closes #2222

The `completion:error` and `completion:last_attempt` hooks were updated in v1.15.1 to forward `attempt_number`, `max_attempts`, and `is_last_attempt` as keyword arguments to handlers. However, no tests existed to verify this behavior or the backward compatibility guarantee for old-style handlers.

This PR adds 8 unit tests covering:

- `attempt_number`, `max_attempts`, `is_last_attempt` are forwarded correctly
- `is_last_attempt=True` on the final attempt
- `max_attempts=None` for unlimited retries
- Old-style `handler(error)` still works without receiving kwargs
- Multiple handlers all receive metadata
- Mixed old-style and new-style handlers coexist on the same hook

## Motivation and Context

Issue #2222 requested that `completion:error` and `completion:last_attempt` hooks expose attempt metadata to handlers. This was implemented in v1.15.1 but had no test coverage. These tests lock in the behavior and the backward compatibility contract.

## How Has This Been Tested?

All 8 tests pass locally: `pytest tests/test_hooks.py -v`

## Breaking Changes

None.

## Types of changes
- [x] Documentation update

## Checklist
- [x] I have read the contributing guide
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added or updated documentation as needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ### Fixed
 - **Templating (GenAI/VertexAI)**: `process_message` no longer crashes with `TypeError: Can't compile non template nodes` when multimodal messages contain image/URI/bytes Parts alongside `validation_context`. Non-text Parts (where `part.text` is `None`) now pass through unchanged. ([#2253](https://github.com/567-labs/instructor/issues/2253))
 
+### Tests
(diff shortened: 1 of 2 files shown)


jxnl:
Thanks for the focused test coverage. We are consolidating hook retry-metadata fixes and tests into a maintainer PR for the next minor release, so I am closing this PR rather than merging the test-only PR separately.

PR #2255, opened 2026-04-10
fix(genai): skip templating for non-text Parts in multimodal messages

### LINKED ISSUE #2253: TypeError: Can't compile non template nodes when using image parts with GENAI mode and validation_context
## Description

When passing multimodal messages (containing image parts) to a Gemini model via `GENAI_STRUCTURED_OUTPUTS` or `GENAI_TOOLS` mode with a non-None `validation_context`, `handle_templating` crashes with:

```
TypeError: Can't compile non template nodes
```

## Root cause

In `instructor/templating.py`, `process_message` checks `hasattr(part, "text")` before applying Jinja2 templating. For image/binary `google.genai.types.Part` objects, the `text` attribute exists but is `None`. This causes `SandboxedEnvironment().from_string(None)` to fail because Jinja2's `compile()` skips `_parse` for non-strings and passes `None` directly to the code generator.

https://github.com/jxnl/instructor/blob/main/instructor/templating.py#L25

```python
# Current (broken):
types.Part.from_text(text=apply_template(part.text, context))
if hasattr(part, "text")
else part

# Fix:
types.Part.from_text(text=apply_template(part.text, context))
if hasattr(part, "text") and part.text is not None
else part
```

The same issue exists in the VertexAI branch (line 46).

## Reproduction

No credentials needed — the bug is in message preprocessing, before any API call.

```python
import asyncio

import instructor
from google import genai
from google.genai import types


class Result(BaseModel):
    description: str


async def main() -> None:
    client = genai.Client(api_key="fake-key-not-needed")
    llm = instructor.from_genai(
        client,
        mode=instructor.Mode.GENAI_STRUCTURED_OUTPUTS,
        use_async=True,
    )

    # Crashes in handle_templating before the API call is made
    result = await llm.create(
        model="gemini-2.5-flash",
        response_model=Result,
        validation_context={"key": "value"},
        messages=[
            types.Content(
                role="user",
                parts=[
                    types.Part.from_text(text="What is in this image?"),
                    types.Part.from_uri(
                        file_uri="gs://bucket/image.jpg",
                        mime_type="image/jpeg",
                    ),
                ],
            )
        ],
    )


asyncio.run(main())
```

## Traceback

```
File "instructor/templating.py", line 25, in process_message
    types.Part.from_text(text=apply_template(part.text, context))
File "instructor/templating.py", line 11, in apply_template
    return dedent(SandboxedEnvironment().from_string(text).render(**context))
File "jinja2/environment.py", line 1111, in from_string
    return cls.from_code(self, self.compile(source), gs, None)
File "jinja2/compiler.py", line 112, in generate
    raise TypeError("Can't compile non template nodes")
```

## Version

- instructor 1.12.0 (also verified unfixed in 1.15.1)
- google-genai 1.x
- jinja2 3.x

lawrence3699:
## Summary

Fixes #2253

When using GENAI_TOOLS or GENAI_STRUCTURED_OUTPUTS mode with multimodal messages (image/URI/bytes parts) and a non-None `validation_context`, `process_message` crashes:

```
TypeError: Can't compile non template nodes
```

The crash happens because genai `Part` objects for images/URIs/bytes have a `text` attribute that exists but is `None`. The existing `hasattr(part, "text")` check passes, so `None` gets fed to `SandboxedEnvironment().from_string()`, which Jinja2 rejects.

## Fix

Add `and part.text is not None` to the condition on both the GENAI path (line 26) and the VertexAI path (line 47), so non-text parts pass through unchanged.

**Before:** any Part with a `text` attribute (including `text=None`) is templated → crash
**After:** only Parts with actual text content are templated; image/URI/bytes parts pass through

## Tests

Added 4 regression tests in `tests/genai/test_templating_multimodal.py`:
- URI parts pass through untemplated
- Raw bytes parts pass through untemplated
- End-to-end `handle_templating` with mixed text+image content
- No-context passthrough (no crash, no transformation)

All existing tests pass (`pytest tests/genai/ tests/test_message_processing.py tests/test_formatting.py` — 46 passed).

### DIFF
--- a/instructor/templating.py
+++ b/instructor/templating.py
@@ -23,7 +23,7 @@ def process_message(
             parts=[
                 (
                     types.Part.from_text(text=apply_template(part.text, context))
-                    if hasattr(part, "text")
+                    if hasattr(part, "text") and part.text is not None
                     else part
                 )
                 for part in message.parts
@@ -44,7 +44,7 @@ def process_message(
             parts=[
                 (
                     gm.Part.from_text(apply_template(part.text, context))
-                    if hasattr(part, "text")
+                    if hasattr(part, "text") and part.text is not None
                     else part
                 )
                 for part in message.parts
--- a/tests/genai/test_templating_multimodal.py
+++ b/tests/genai/test_templating_multimodal.py
@@ -1,0 +1,80 @@
+"""Regression test for #2253: templating crashes on genai image parts.
+
+When a genai Content message contains non-text Parts (images, URIs, bytes),
+the `text` attribute exists but is None. process_message must skip
+templating for these parts instead of passing None to Jinja2.
+"""
+
+from google.genai import types
+
+from instructor.mode import Mode
+from instructor.templating import handle_templating, process_message
+
+
+def test_process_message_skips_non_text_genai_parts():
+    """Image/URI parts with text=None must pass through untemplated."""
+    msg = types.Content(
+        role="user",
+        parts=[
+            types.Part.from_text(text="Describe {{ subject }}"),
+            types.Part.from_uri(
+                file_uri="gs://bucket/image.jpg", mime_type="image/jpeg"
+            ),
+        ],
+    )
+
+    result = process_message(msg, {"subject": "this image"}, Mode.GENAI_TOOLS)
+
+    assert result.parts[0].text == "Describe this image"
+    # The URI part should pass through unchanged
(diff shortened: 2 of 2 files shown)


jxnl:
Thanks for the PR. Closing in favor of #2259 which we merged as the canonical fix for #2253.

PR #2287, opened 2026-04-25
fix(bedrock): strip think tags and fix reask crash for reasoning models

### LINKED ISSUE #2076: Allow MD_JSON for bedrock client

octo-patch:
Fixes #2076

## Problem

Reasoning models on AWS Bedrock (e.g. Kimi K2 Thinking) prefix their text response with a `<think>...</think>` block before emitting the actual JSON. The JSON extractor (`extract_json_from_codeblock`) finds the first `{` and last `}` in the whole response, which may land inside the think block and produce invalid JSON. This causes repeated validation failures and costly retries.

A second related bug: when `Mode.MD_JSON` is used with `from_bedrock()` and a validation error fires, `reask_md_json` tries to access `response.choices[0]`. Bedrock responses are plain dicts (no `.choices`), so this raises:
```
AttributeError: 'dict' object has no attribute 'choices'
```
Because `dict` also lacks `.choices`, Bedrock responses were incorrectly hitting the stream-response branch and the assistant turn was silently dropped from the retry message list.

## Solution

**1. Think-tag stripping** (`instructor/processing/function_calls.py`)
- `parse_bedrock_json` (used by `BEDROCK_JSON` mode): strip `<think>…</think>` before code-block extraction and JSON parse.
- `_extract_text_content` (used by `parse_json` / `MD_JSON` mode): strip think tags when the completion is a Bedrock dict.

**2. Bedrock-aware reask** (`instructor/providers/openai/utils.py`)
- `reask_md_json`: check for Bedrock dict responses (`isinstance(response, dict) and "output" in response`) *before* the stream-response guard so they are handled with the correct Bedrock message format instead of crashing.

## Testing

9 new unit tests in `tests/llm/test_bedrock/test_think_tags.py` cover:
- Plain JSON with no think tags (regression guard)
- Think tags before JSON
- Think tags that themselves contain braces (the core failure case)
- Think tags combined with markdown code-fence responses
- Multi-line think blocks
- `_extract_text_content` stripping for the MD_JSON path
- `reask_md_json` with a Bedrock dict response (asserts correct Bedrock-format correction message)
- `reask_md_json` with a normal OpenAI response (unchanged behavior)

```
9 passed in 26.82s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## [Unreleased]
 
 ### Fixed
+- **Bedrock**: Strip `<think>...</think>` blocks from Bedrock responses in both `BEDROCK_JSON` and `MD_JSON` modes before JSON extraction, fixing parse failures with reasoning models such as Kimi K2 Thinking ([#2076](https://github.com/567-labs/instructor/issues/2076))
+- **Bedrock/MD_JSON**: `reask_md_json` no longer raises `AttributeError: 'dict' object has no attribute 'choices'` when retrying after a validation failure with a Bedrock converse dict response ([#2076](https://github.com/567-labs/instructor/issues/2076))
 - **Templating (GenAI/VertexAI)**: `process_message` no longer crashes with `TypeError: Can't compile non template nodes` when multimodal messages contain image/URI/bytes Parts alongside `validation_context`. Non-text Parts (where `part.text` is `None`) now pass through unchanged. ([#2253](https://github.com/567-labs/instructor/issues/2253))
 
 ---
--- a/instructor/processing/function_calls.py
+++ b/instructor/processing/function_calls.py
@@ -71,7 +71,9 @@ def _extract_text_content(completion: Any) -> str:
     # Bedrock format
     if isinstance(completion, dict) and "output" in completion:
         try:
-            return completion.get("output").get("message").get("content")[0].get("text")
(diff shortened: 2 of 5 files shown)


jordandunmire97-ai (approved):

jxnl:
Reviewed for merge readiness. This is the right scoped PR for #2076 and should stay open: it handles both the `<think>...</think>` parsing failure and the Bedrock dict retry crash. It is currently conflicting, mostly around changelog/current parser code, so it needs a rebase/refresh and CI before merge. One cleanup I would make while rebasing: avoid duplicating think-tag stripping helpers across modules unless both are actually used.

jordandunmire97-ai (approved):

jxnl:
Rebuilt this on the current v2 Bedrock handler in #2514, with credit to @octo-patch. Current `main` already contains the Bedrock-dict reask path; the canonical PR replaces the remaining provider-local JSON regex with the shared last-complete-JSON extractor, preserves escaped content, and adds caller-input immutability regressions. Keeping this PR open until #2514 reaches `main`, then it can close as superseded.

jxnl:
Closing this PR as superseded by #2514. Current `main` already contains the Bedrock dict-reask correction from this patch; #2514 carries the remaining reasoning-prefixed JSON fix on the current v2 handler by reusing the shared JSON extractor, with deterministic regression coverage and a fully green matrix. Credit is retained in #2514. Issue #2076 remains open until the fix reaches `main`.

PR #2275, opened 2026-04-15
fix(security): redact api_key from debug log output

CrepuscularIRIS:
## Summary

- Fixes #2265 — `logger.debug` calls in `handle_response_model()` logged the full `new_kwargs` dict verbatim, leaking `api_key` (and similar secrets) to any downstream project running at `DEBUG` level.
- Adds a `_redact_kwargs()` helper that replaces `api_key`, `api_secret`, `api_token`, and `authorization` values with `'[REDACTED]'` before they reach the f-string or the `extra` dict. The original dict is untouched, so the actual API call still receives the real key.
- Adds 5 regression tests (unit + integration via `caplog`) to verify redaction works and the original kwargs are never mutated.
- Adds a `### Security` entry to `CHANGELOG.md` under `[Unreleased]`.

## Reproduction

```python
import logging, sys
logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)

from instructor.processing.response import handle_response_model
from instructor.mode import Mode
from pydantic import BaseModel

class User(BaseModel):
    name: str
    age: int

# Before fix: api_key appears in plaintext in log output
# After fix: api_key shows as '[REDACTED]'
_, kw = handle_response_model(
    response_model=User, mode=Mode.TOOLS,
    model="gpt-4", messages=[{"role": "user", "content": "test"}],
    api_key="sk-SECRET",
)
assert kw["api_key"] == "sk-SECRET"  # returned kwargs still carry real key
```

## Test plan

- [x] `pytest tests/test_redact_kwargs.py` — 5/5 pass
- [x] `ruff check` and `ruff format` clean
- [x] Existing tests unaffected (298 pass; 9 pre-existing failures in gemini/genai unrelated)
- [x] Manual reproduction confirms `[REDACTED]` in log output

PR #2297, opened 2026-05-03
fix(logging): redact api_key and sensitive keys in handle_response_model debug log

### LINKED ISSUE #2265: logger.debug in response.py leaks api_key verbatim via new_kwargs
In `instructor/processing/response.py` (around line 405), there's a debug log call:

```python
logger.debug(f"Instructor Request: ..., new_kwargs={new_kwargs}")
```

`new_kwargs` contains the full request dict including `api_key` in plaintext. When a downstream project (e.g. cognee) sets its root logger to DEBUG, the API key gets written to stdout/log files.

**Impact:** Any instructor user running at DEBUG log level with a cloud LLM provider has their API key written to logs. Docker stdout logs are often shipped to centralized logging (Loki, CloudWatch, etc.) and retained long-term.

**Suggested fix:** Redact known-sensitive keys (`api_key`, `api_secret`) from `new_kwargs` before logging, or log only safe fields.

**Discovered in:** cognee v0.5.8 + instructor 1.12.0 + Mammouth/Mistral provider. Confirmed: key appeared ~120 times in container stdout during a single cognify run.

Found in cooperation with Claude (Anthropic).

xodn348:
## Summary

`handle_response_model()` in `instructor/processing/response.py` logs the full request kwargs at DEBUG level using `f"... {new_kwargs=}"`. Because `new_kwargs` is a shallow copy of the caller's kwargs dict, any `api_key` (or `api_secret`, `authorization`, `token`) passed directly to the provider client ends up written verbatim to every DEBUG-level log sink — stdout, CloudWatch, Loki, Datadog, etc. — where it may be retained for months. This was confirmed by a user running at DEBUG level with Mammouth/Mistral where the key appeared ~120 times in container stdout during a single `cognify` run (see #2265).

The fix adds a private `_SENSITIVE_KEYS` frozenset and a `_redact_kwargs()` helper that returns a shallow copy with known-sensitive keys replaced by `"[redacted]"`. Both `logger.debug` call-sites in `handle_response_model()` (the parallel-modes early-return path and the standard path) now log `_redact_kwargs(new_kwargs)` instead of `new_kwargs` directly. The actual `new_kwargs` dict returned to callers is **not** modified.

## Issue

Fixes #2265

## Local verification

```
$ cd /tmp/instructor
$ ruff check instructor/processing/response.py
All checks passed!

$ python3 -c "
from instructor.processing.response import _redact_kwargs
result = _redact_kwargs({'api_key': 'sk-secret123', 'model': 'gpt-4', 'messages': []})
assert result['api_key'] == '[redacted]'
assert result['model'] == 'gpt-4'
result2 = _redact_kwargs({'api_secret': 'tok', 'temperature': 0.7})
assert result2['api_secret'] == '[redacted]'
result3 = _redact_kwargs({'model': 'claude-3', 'max_tokens': 100})
assert result3 == {'model': 'claude-3', 'max_tokens': 100}
print('All assertions passed')
"
All assertions passed

$ python -m pytest tests/test_process_response.py tests/test_response_model_conversion.py -q --no-header \
    -k "not gemini and not genai and not vertexai"
7 passed, 3 deselected in 1.52s

$ python -m pytest tests/ -q --no-header \
    -k "not llm and not openai and not anthropic and not gemini and not genai and not vertexai \
        and not cohere and not bedrock and not fireworks and not mistral and not cerebras \
        and not litellm and not writer and not xai and not batch_in_memory and not test_patch" \
    --ignore=tests/test_auto_client.py --ignore=tests/docs
270 passed, 23 skipped, 126 deselected, 59 warnings in 1.67s
=== LOCAL_TEST_PASSED ===
```

## Risk

Only the logged representation of `new_kwargs` changes; the dict itself is unmodified, so all downstream request logic is unaffected. The redaction is a strict allowlist of known credential key names (`api_key`, `api_secret`, `authorization`, `token`) — any future keys not in this set will still be logged in full, which is the safe default. The only observable behaviour change is that DEBUG logs will show `'api_key': '[redacted]'` rather than the real value.

### DIFF
--- a/instructor/processing/response.py
+++ b/instructor/processing/response.py
@@ -58,6 +58,13 @@ from ..mode import Mode
 from .multimodal import convert_messages
 from ..utils.core import prepare_response_model
 
+_SENSITIVE_KEYS: frozenset[str] = frozenset({"api_key", "api_secret", "authorization", "token"})
+
+
+def _redact_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
+    """Return a shallow copy of kwargs with sensitive keys replaced by '[redacted]'."""
+    return {k: "[redacted]" if k in _SENSITIVE_KEYS else v for k, v in kwargs.items()}
+
 # Anthropic utils
 from ..providers.anthropic.utils import (
     handle_anthropic_json,
@@ -439,8 +446,9 @@ def handle_response_model(
 
     if mode in PARALLEL_MODES:
         response_model, new_kwargs = PARALLEL_MODES[mode](response_model, new_kwargs)  # type: ignore
+        _safe_kwargs = _redact_kwargs(new_kwargs)
         logger.debug(
-            f"Instructor Request: {mode.value=}, {response_model=}, {new_kwargs=}",
+            f"Instructor Request: {mode.value=}, {response_model=}, new_kwargs={_safe_kwargs!r}",
             extra={
                 "mode": mode.value,
                 "response_model": (
@@ -449,7 +457,7 @@ def handle_response_model(
                     and hasattr(response_model, "__name__")
                     else str(response_model)
                 ),
-                "new_kwargs": new_kwargs,
+                "new_kwargs": _safe_kwargs,
             },
         )
         return response_model, new_kwargs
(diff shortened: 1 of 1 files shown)


jxnl-oai:
Thanks for the fix. This overlaps with #2275, which covers the same issue with broader regression coverage and a changelog entry, so I am closing this one as a duplicate in favor of #2275.

jxnl-oai (approved):
Reviewed locally. The core fix is sound; I found one additional nested-redaction edge case to carry as a follow-up cleanup on main because this environment cannot push to the source fork.

PR #2308, opened 2026-05-17
fix: stream array-root JSON extraction

pragnyanramtha:
## Describe your changes

Fixes the streaming JSON extractor so markdown/plain streaming responses whose root JSON value is an array are preserved as arrays instead of starting at the first object inside the array. The sync and async extractors now track both object and array delimiters while preserving the existing string/escape handling.

Added regression coverage for chunked markdown JSON streams with array roots in both sync and async paths.

N/A

## Validation

- `uv run --extra dev pytest tests/test_utils.py tests/test_json_extraction.py tests/test_json_extraction_edge_cases.py tests/dsl/test_gemini_tools_async_streaming.py -q` -> 63 passed, 2 skipped
- `uv run ruff format --check instructor/utils/core.py tests/test_utils.py` -> passed
- `uv run ruff check instructor/utils/core.py tests/test_utils.py` -> passed
- `git diff --check` -> passed


jxnl:
Thanks for the focused fix. We are consolidating small streaming/JSON parser fixes into a maintainer PR for the next minor release, so I am closing this PR rather than merging individual narrow PRs. The array-root streaming extraction case will be covered there.

PR #2317, opened 2026-05-22
fix(retry): emit hook attempt metadata

### LINKED ISSUE #2222: Expose attempt metadata on completion:error and completion:last_attempt hooks

2830500285:
## Describe your changes

Emit retry metadata from the documented `completion:error` and `completion:last_attempt` hooks in v2 retry paths.

Fixes #2222

## Changes

- Adds shared retry metadata helpers for attempt number, max attempts, and final-attempt detection.
- Emits `completion:error` metadata for retryable completion call failures in sync and async retry paths.
- Emits `completion:last_attempt` once when retry handling reaches its final failure, including validation exhaustion.
- Adds sync and async regression tests, including legacy one-argument hook compatibility.
- Updates the changelog.

## Testing

- `uv run pytest tests/v2/test_retry_runtime.py`
- `uv run ruff check instructor/v2/core/retry.py tests/v2/test_retry_runtime.py`
- `uv run ty check instructor/v2/core/retry.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Hooks**: Emit documented retry metadata from `completion:error` and `completion:last_attempt` hooks in v2 retry paths.
+
 ---
 
 ## [1.15.2] - 2026-05-10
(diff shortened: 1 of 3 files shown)


jxnl:

PR #2319, opened 2026-05-24
fix(v2/retry): max_retries off-by-one — treat as retry count, not max-attempts

devteamaegis:
## Summary

`retry_sync_v2` and `retry_async_v2` use `stop_after_attempt(max(max_retries, 1))` to build the tenacity stop condition, which interprets `max_retries` as the **total number of attempts** rather than the number of **retries**.

### What breaks

| `max_retries` | Before (wrong) | After (correct) |
|---|---|---|
| 0 | `stop_after_attempt(1)` → 1 attempt, 0 retries ✓ | same |
| 1 (default in `patch.py`) | `stop_after_attempt(1)` → 1 attempt, **0 retries** ✗ | `stop_after_attempt(2)` → 2 attempts, 1 retry ✓ |
| 3 (default in `client.py`) | `stop_after_attempt(3)` → 3 attempts, **2 retries** ✗ | `stop_after_attempt(4)` → 4 attempts, 3 retries ✓ |

With `max_retries=1` — the default in `patch.py` — the function silently never retries even after a validation error, making the `max_retries` parameter a no-op for the most common call-sites.

## Fix

```python
# Before
stop_condition = stop_after_attempt(max(max_retries, 1))

# After
stop_condition = stop_after_attempt(max_retries + 1)
```

Applied to both `retry_sync_v2` (line 129) and `retry_async_v2` (line 345).

## Tests

Added three regression tests in `tests/v2/test_retry_runtime.py` that exercise the **integer** `max_retries` code-path directly. The existing tests all construct a `Retrying`/`AsyncRetrying` object explicitly, which bypasses the affected `if isinstance(max_retries, int):` branch entirely.

- `test_max_retries_0_makes_single_attempt` — no retries, raises after 1 attempt
- `test_max_retries_1_allows_one_retry` — succeeds on the 2nd attempt
- `test_max_retries_3_allows_three_retries` — succeeds on the 4th attempt

### DIFF
--- a/instructor/v2/core/retry.py
+++ b/instructor/v2/core/retry.py
@@ -126,7 +126,11 @@ def retry_sync_v2(
 
     # Setup retrying
     if isinstance(max_retries, int):
-        stop_condition = stop_after_attempt(max(max_retries, 1))
+        # max_retries is the number of *retries*, so the total number of attempts
+        # is max_retries + 1.  Using max(max_retries, 1) (i.e. treating the value
+        # as max-attempts) would silently drop a retry: max_retries=1 would only
+        # make one attempt (zero retries) instead of two.
+        stop_condition = stop_after_attempt(max_retries + 1)
         timeout = kwargs.get("timeout")
         if isinstance(timeout, (int, float)):
             stop_condition = stop_condition | stop_after_delay(timeout)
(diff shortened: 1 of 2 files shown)


biswajeetdev (commented):
The bug is clearly named and the fix is minimal — `stop_after_attempt(max(max_retries, 1))` was treating `max_retries` as a max-attempts count when the documented semantics (and the parameter name) mean retry count. So `max_retries=1` gave `stop_after_attempt(1)` — one attempt, zero retries — the opposite of what callers expect.

`stop_after_attempt(max_retries + 1)` is exactly right.

One edge case worth confirming: `max_retries=0` now gives `stop_after_attempt(1)` — one attempt, no retries. That is semantically correct for "no retries." The old code gave the same result via `max(0, 1) = 1`, so behavior is unchanged for the zero case.

The regression tests are thorough — harness-driven approach with parameterized fail counts makes it easy to add new cases. One small addition that would round out coverage: a test where `max_retries=0` confirms exactly 1 total attempt (the "no retry at all" contract), since that boundary is what old users who wrote `max_retries=0` would rely on.

PR #2323, opened 2026-05-27
fix: use actual attempt count for InstructorRetryException.n_attempts

rmotgi1227:
## Summary

`InstructorRetryException.n_attempts` is documented as *"The total number of attempts made"* but was set to `len(failed_attempts)` — a list that only grows when `response_parser` raises a `ValidationError`.  If the underlying API call itself throws a non-`ValidationError` (network error, rate-limit, `RuntimeError`, …), that attempt is never recorded in `failed_attempts`, so `n_attempts` is silently under-counted.

## Concrete example

```
max_retries=3
attempt 1 → ValidationError   (tenacity retries)
attempt 2 → RuntimeError       (tenacity re-raises immediately)

Before fix: n_attempts = 1    ← only the ValidationError counted
After  fix: n_attempts = 2    ← both attempts counted
```

The edge case that hurts users the most is a single API error on the very first call:

```python
except InstructorRetryException as e:
    print(f"Failed after {e.n_attempts} attempts")
    # Before: prints "Failed after 0 attempts" — obviously wrong
    # After:  prints "Failed after 1 attempts"
```

## Root cause

Both `retry_sync_v2` and `retry_async_v2` raised `InstructorRetryException` with `n_attempts=len(failed_attempts)`.  `failed_attempts` is only appended inside the `except ValidationError` block, so any other failure path left it empty.

## Fix

Introduce a `total_attempts` counter (one line) at the top of each retry loop body:

```python
total_attempts = attempt.retry_state.attempt_number
```

This is always current regardless of how the iteration ends, and is used in place of `len(failed_attempts)` when constructing the exception.  Applied to both the sync (`retry_sync_v2`) and async (`retry_async_v2`) paths.

## Tests

`tests/core/test_retry_n_attempts.py` — three cases, no real API calls:

| Test | Scenario | Expected `n_attempts` |
|------|----------|----------------------|
| `test_api_error_on_first_attempt_reports_one` (sync) | 1 API error, 0 retries | 1 |
| `test_validation_then_api_error_reports_two` (sync) | 1 ValidationError + 1 API error | ≥ 1 |
| `test_api_error_on_first_attempt_reports_one` (async) | 1 API error, 0 retries | 1 |

### DIFF
--- a/instructor/v2/core/retry.py
+++ b/instructor/v2/core/retry.py
@@ -141,10 +141,12 @@ def retry_sync_v2(
     failed_attempts: list[FailedAttempt] = []
     last_exception: Exception | None = None
     total_usage = _initialize_usage(provider)
+    total_attempts = 0
 
     try:
         for attempt in max_retries_instance:
             with attempt:
+                total_attempts = attempt.retry_state.attempt_number
                 # Call API
                 if hooks:
                     hooks.emit_completion_arguments(**kwargs)
@@ -217,14 +219,14 @@ def retry_sync_v2(
             last_exception = e
 
         logger.error(
-            f"Max retries exceeded. Total attempts: {len(failed_attempts)}, "
+            f"Max retries exceeded. Total attempts: {total_attempts}, "
             f"Last error: {last_exception}"
         )
 
         raise InstructorRetryException(
             str(last_exception),
             last_completion=failed_attempts[-1].completion if failed_attempts else None,
-            n_attempts=len(failed_attempts),
+            n_attempts=total_attempts,
             total_usage=total_usage,
             messages=extract_messages(kwargs),
             create_kwargs=kwargs,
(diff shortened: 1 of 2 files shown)


biswajeetdev (commented):
The root bug is subtle: `len(failed_attempts)` only counts attempts that appended to `failed_attempts` inside the retry loop, which only happens on `InstructorValidationError`. A non-validation exception on a later attempt (e.g., a network error after the first validation failure) would increment the real attempt counter but never append to `failed_attempts`, leaving `n_attempts` understated.

Using `attempt.retry_state.attempt_number` from tenacity directly is the right fix — it tracks all attempts regardless of exception type.

Two small notes:

1. `total_attempts` is initialised to `0` before the loop. If the loop body never executes (edge case: `stop_condition` stops before the first attempt — unlikely but theoretically possible with a custom stop condition), both `raise` paths at the bottom would report `n_attempts=0` and `failed_attempts[-1]` would raise `IndexError`. The existing code had the same latent issue with `len(failed_attempts)`, so this is not a regression, just worth a comment.

2. The fix is applied to both `retry_sync_v2` and `retry_async_v2` — good. Worth checking that `retry_sync` (v1 path, if it exists) has the same semantics or is documented as intentionally different.

PR #2324, opened 2026-05-28
feat(hooks): fire completion:error and completion:last_attempt with attempt metadata

### LINKED ISSUE #2222: Expose attempt metadata on completion:error and completion:last_attempt hooks

NishchayMahor:
## Describe your changes

`completion:error` and `completion:last_attempt` were already declared on the `Hooks` API (`emit_completion_error` and `emit_completion_last_attempt` exist on `Hooks` and accept `**kwargs`), but neither hook is fired from `retry_sync_v2` or `retry_async_v2`. Any user-registered handler for these two events sits dormant today.

This wires both up and threads attempt metadata through as keyword arguments:

- `completion:error` fires from both the API-call failure path and the validation-error path inside the retry loop, with `error`, `attempt_number`, `max_attempts`, `is_last_attempt`.
- `completion:last_attempt` fires once from the outer `except` right before raising `InstructorRetryException`, with the final exception and the same metadata (`is_last_attempt=True`).

A handler can now distinguish an intermediate retryable error from the final failure without keeping its own shadow counter, which was the original ask in the issue. Both sync and async paths get the same treatment so the behaviour is symmetric.

### Backward compatibility

`Hooks.emit` already does `inspect.signature(handler).bind(...)` and falls back to calling the handler without the new kwargs when it can't bind them, so an existing `def on_error(error): ...` registration keeps working unchanged. The tests cover this — see `test_retry_sync_v2_attempt_metadata_keeps_legacy_handler_signature_compatible`.

### \`max_attempts\` semantics

\`max_attempts\` is \`None\` when the caller passes a \`Retrying\`/\`AsyncRetrying\` instance directly, since extracting the attempt limit from an arbitrary \`stop\` callable isn't reliable. When \`max_retries\` is an \`int\`, \`max_attempts\` is set to \`max(max_retries, 1)\`. In the \`None\` case \`is_last_attempt\` stays \`False\` on intermediate \`completion:error\` events, but \`completion:last_attempt\` still fires with \`is_last_attempt=True\` on exhaustion, which is the more important signal.

### Test note

While writing the new tests I had to wire the mock \`reask_handler\` parameters as \`kwargs, response, exception\` (positional names matching the call site) rather than \`_response\`/\`_exception\`, otherwise tenacity exhausts after a single attempt because the reask call raises \`TypeError\` on keyword binding. I didn't touch the two pre-existing tests that still use that pattern.

Closes #2222

## Local verification

\`\`\`
\$ pytest tests/v2/test_retry_runtime.py -v
========== 11 passed in 0.03s ==========

\$ pytest tests/v2/ --deselect tests/v2/test_provider_modes.py -q
========== 1418 passed, 109 skipped, 111 deselected ==========

\$ pre-commit run --files instructor/v2/core/retry.py tests/v2/test_retry_runtime.py
Run Linter Check (Ruff)..................................................Passed
Run Formatter (Ruff).....................................................Passed
\`\`\`

Ruff lint, Ruff format, and the 11 new + 6 existing retry-runtime tests all pass. \`ty check instructor/v2/core/retry.py\` reports the same 2 pre-existing \`unused-type-ignore-comment\` diagnostics with and without this branch — happy to clean those up in a separate PR if it'd help.

### DIFF
--- a/instructor/v2/core/retry.py
+++ b/instructor/v2/core/retry.py
@@ -55,6 +55,16 @@ def _finalize_parsed_response(parsed: Any, response: Any) -> Any:
     return parsed
 
 
+def _attempt_metadata(attempt_number: int, max_attempts: int | None) -> dict[str, Any]:
+    return {
+        "attempt_number": attempt_number,
+        "max_attempts": max_attempts,
+        "is_last_attempt": (
+            max_attempts is not None and attempt_number >= max_attempts
+        ),
+    }
+
+
 def _initialize_usage(provider: Provider | Mode) -> Any:
     from openai.types.completion_usage import (
         CompletionTokensDetails,
@@ -135,8 +145,10 @@ def retry_sync_v2(
             retry=retry_if_exception_type(ValidationError),
             reraise=True,
         )
+        max_attempts: int | None = max(max_retries, 1)
     else:
         max_retries_instance = max_retries
+        max_attempts = None
 
     failed_attempts: list[FailedAttempt] = []
     last_exception: Exception | None = None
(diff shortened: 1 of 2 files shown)


jxnl:

PR #2328, opened 2026-05-30
feat(hooks): pass attempt_number to parse:error handlers

sebastianbreguel:
## Problem

`parse:error` is the only error hook that actually fires in the v2 retry loop (`completion:error` / `completion:last_attempt` are defined but never emitted). Yet it received only the exception:

```python
def handler(error): ...
```

So a handler can read *what* failed validation (via `error.errors()`), but not *which retry attempt* it failed on. With `max_retries=7`, you can't tell whether a field fails on attempt 1 or only gives up at 7 — exactly the question you need answered when calibrating retries or instrumenting flaky schema validation. The `attempt_number` is already available at the emission site; it just wasn't being passed through.

## Change

Pass `attempt_number` to `parse:error` handlers from both retry sites (sync `retry_sync_v2` and async `retry_async_v2`):

```python
def handler(error, *, attempt_number): ...
```

## Backward compatible by design

No new mechanism — `Hooks.emit` already tries to bind kwargs to the handler signature and falls back to positional-only (`handler(*args)`) when it can't. Existing handlers that only accept `error` keep working unchanged; the metadata is silently dropped for them.

## Scope

Intentionally limited to `attempt_number`. `max_attempts` / `is_last_attempt` are **not** included because they can't be reliably derived at parse-fail time when `max_retries` is a custom `Retrying` / `AsyncRetrying` instance (the `stop` condition is a black box evaluated *after* the attempt). `attempt_number` is the one value that's always available and correct.

## Tests

- Updated `test_retry_sync_v2_reasks_after_validation_error` to assert the emitted `attempt_number`.
- Added `test_emit_parse_error_supports_old_and_new_style_handlers` exercising the real `Hooks` class with both an old-style `(error)` handler and a new-style `(error, *, attempt_number)` handler.

`tests/v2/test_retry_runtime.py` — 7 passed. `ruff check` / `ruff format` clean.

## Docs / CHANGELOG

Updated the hooks table + prose in `docs/concepts/hooks.md` and added a `[Unreleased]` CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+- **Hooks**: `parse:error` now passes `attempt_number` to handlers, so you can tell which retry a Pydantic validation failed on. Handlers that only accept `error` keep working unchanged.
+
 ---
 
 ## [1.15.2] - 2026-05-10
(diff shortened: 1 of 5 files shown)


jxnl:
Thanks for the focused hook-metadata fix. This is now absorbed into the maintainer consolidation PR #2340, which covers `parse:error` attempt metadata along with the adjacent retry hook fixes for the next minor release.

PR #2330, opened 2026-05-31
feat: emit attempt metadata in completion:error and completion:last_attempt hooks

### LINKED ISSUE #2222: Expose attempt metadata on completion:error and completion:last_attempt hooks

sridhar-3009:
## Summary

Resolves #2222.

`completion:error` and `completion:last_attempt` were defined in the hooks API but **never called** from the retry loop. This PR wires both into `retry_sync_v2` and `retry_async_v2` and passes attempt metadata as keyword arguments.

### `retry.py`
- `emit_completion_error` now fires for each API-level exception with:
  - `attempt_number`: current tenacity attempt counter
  - `max_attempts`: int if `max_retries` was an int, else `None`
  - `is_last_attempt=False`
- `emit_completion_last_attempt` fires once on retry exhaustion with:
  - `attempt_number`: total failed attempts
  - `max_attempts`: same as above
  - `is_last_attempt=True`

### `hooks.py`
- `CompletionErrorHandler` protocol updated to document the new optional kwargs.

## Backwards compatibility

Existing handlers that accept only `error: Exception` continue to work — `emit()` already uses `inspect.signature` to detect whether kwargs should be forwarded.

## Example

```python
def on_error(error, *, attempt_number, max_attempts, is_last_attempt):
    level = logging.ERROR if is_last_attempt else logging.WARNING
    logger.log(level, "API error on attempt %d/%s: %s", attempt_number, max_attempts, error)

client.on("completion:error", on_error)
client.on("completion:last_attempt", on_error)
```

### DIFF
--- a/instructor/v2/core/hooks.py
+++ b/instructor/v2/core/hooks.py
@@ -34,7 +34,14 @@ class CompletionResponseHandler(Protocol):
 class CompletionErrorHandler(Protocol):
     """Protocol for completion error and last attempt handlers."""
 
-    def __call__(self, error: Exception) -> None: ...
+    def __call__(
+        self,
+        error: Exception,
+        *,
+        attempt_number: int = ...,
+        max_attempts: int | None = ...,
+        is_last_attempt: bool = ...,
+    ) -> None: ...
 
 
 class ParseErrorHandler(Protocol):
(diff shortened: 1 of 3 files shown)


jxnl:

PR #2336, opened 2026-06-01
fix(utils): guard empty/absent messages in handle_gemini_json

### LINKED ISSUE #2335: fix: handle_gemini_json crashes with empty or absent messages list
## Bug

`handle_gemini_json` in `instructor/v2/providers/gemini/utils.py` line 477 unconditionally accesses `new_kwargs["messages"][0]["role"]` without guarding for a missing `messages` key or an empty list.

- If `messages` is absent: `KeyError: 'messages'`
- If `messages` is `[]`: `IndexError: list index out of range`

Reproduction:

```python
from instructor.v2.providers.gemini.utils import handle_gemini_json

class User(BaseModel):
    name: str

handle_gemini_json(User, {"messages": []})
# IndexError: list index out of range
```

## Root cause

The condition `if new_kwargs["messages"][0]["role"] != "system":` assumes messages is a non-empty list without any guard.

## Fix

Replace the unguarded access with `messages = new_kwargs.get("messages") or []` and check `not messages or messages[0]["role"] != "system"` before inserting the system prompt.

devteamaegis:
## What's broken

`handle_gemini_json` crashes when called with an empty or absent `messages` list. Line 477 unconditionally accesses `new_kwargs["messages"][0]["role"]`, raising `IndexError: list index out of range` for `{"messages": []}` and `KeyError: 'messages'` when the key is missing entirely. Both are reachable from normal caller code.

## Why it happens

The condition had no guard before indexing into `messages`, assuming the list was always non-empty.

## Fix

Replaced the bare index with `messages = new_kwargs.get("messages") or []` and changed the condition to `not messages or messages[0]["role"] != "system"` so an empty or missing list safely inserts the system prompt rather than crashing.

## Test

Added `test_handle_gemini_json_empty_or_absent_messages_does_not_raise` to `tests/v2/test_gemini_utils_deterministic.py`. It calls `handle_gemini_json` with both `{"messages": []}` and `{}` and asserts no exception is raised and the response mime type is set correctly.

Fixes #2335

### DIFF
--- a/instructor/v2/providers/gemini/utils.py
+++ b/instructor/v2/providers/gemini/utils.py
@@ -474,10 +474,11 @@ def handle_gemini_json(
         """
     )
 
-    if new_kwargs["messages"][0]["role"] != "system":
-        new_kwargs["messages"].insert(0, {"role": "system", "content": message})
+    messages = new_kwargs.get("messages") or []
+    if not messages or messages[0]["role"] != "system":
+        new_kwargs.setdefault("messages", []).insert(0, {"role": "system", "content": message})
     else:
-        new_kwargs["messages"][0]["content"] += f"\n\n{message}"
+        messages[0]["content"] += f"\n\n{message}"
 
     new_kwargs["generation_config"] = new_kwargs.get("generation_config", {}) | {
         "response_mime_type": "application/json"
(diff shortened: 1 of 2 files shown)


jxnl:
Thanks for the focused fix. We are consolidating this small bug-fix cluster into a maintainer PR that covers multiple duplicate reports for the next minor release, so I am closing this PR rather than merging individual one-line PRs. The Gemini empty/missing `messages` case will be covered there.

PR #2337, opened 2026-06-02
fix(gemini): guard handle_gemini_json against empty or missing messages

### LINKED ISSUE #2335: fix: handle_gemini_json crashes with empty or absent messages list

Oxygen56:
## Summary

`handle_gemini_json` in `instructor/v2/providers/gemini/utils.py` crashed with `KeyError`/`IndexError` when `messages` was missing from kwargs or was an empty list.

## Root cause

The condition `if new_kwargs["messages"][0]["role"] != "system"` assumed messages is a non-empty list without any guard.

## Fix

Replace the unguarded access with defensive `.get()` and `.setdefault()` calls:

```python
messages = new_kwargs.get("messages") or []
if not messages or messages[0].get("role") != "system":
    new_kwargs.setdefault("messages", []).insert(0, {"role": "system", "content": message})
```

## Test

Added `test_handle_gemini_json_guards_empty_or_missing_messages` covering both the missing-key and empty-list cases.

Closes #2335

### DIFF
--- a/instructor/v2/providers/gemini/utils.py
+++ b/instructor/v2/providers/gemini/utils.py
@@ -474,8 +474,9 @@ def handle_gemini_json(
         """
     )
 
-    if new_kwargs["messages"][0]["role"] != "system":
-        new_kwargs["messages"].insert(0, {"role": "system", "content": message})
+    messages = new_kwargs.get("messages") or []
+    if not messages or messages[0].get("role") != "system":
+        new_kwargs.setdefault("messages", []).insert(0, {"role": "system", "content": message})
     else:
         new_kwargs["messages"][0]["content"] += f"\n\n{message}"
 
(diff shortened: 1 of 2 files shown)


jxnl:
Thanks for the focused fix. We are consolidating this small bug-fix cluster into a maintainer PR that covers multiple duplicate reports for the next minor release, so I am closing this PR rather than merging individual one-line PRs. The Gemini empty/missing `messages` case will be covered there.

PR #2339, opened 2026-06-02
fix(streaming): ignore braces inside strings when splitting Iterable objects

vidigoat:
## Problem

`IterableBase.get_object` (used by `create_iterable()` / `Iterable[T]` streaming) splits the streamed buffer into JSON objects by counting `{` and `}` to track nesting depth — but it never tracks whether it is **inside a JSON string**:

```python
for i, c in enumerate(s):
    if c == "{": stack += 1
    if c == "}":
        stack -= 1
        if stack == 0:
            return s[start_index : i + 1], s[i + 2 :]
```

So any `{` or `}` that appears **inside a string value** changes the depth. A closing brace inside a string (e.g. a `":}"` emoticon, a code snippet, or any stray brace in model-generated text) drops `stack` to `0` early, slicing the object mid-string and returning **invalid JSON**. That chunk then fails `model_validate_json`, raising a `ValidationError`/`JSONDecodeError` that **kills the entire stream**.

### Reproduction (no API key)

```python
from instructor.dsl.iterable import IterableBase
obj, rest = IterableBase.get_object('{"bio": "a } b"}, {"bio": "next"}', 0)
# main:  obj == '{"bio": "a }'   -> json.loads(obj) raises "Unterminated string"
```

End-to-end via the streaming API:

```python
chunks = ['{"tasks": [', '{"name": "Alice", "bio": "happy :}"}', ', {"name": "Bob", "bio": "plain"}', ']}']
list(IterableModel(User).tasks_from_chunks(chunks))   # main: ValidationError
```

## Fix

Make `get_object` **string-aware** — track `in_string` (toggled on an unescaped `"`) and an escape flag, and only count `{`/`}` when outside a string. This mirrors the pattern already used in `instructor/v2/core/json.py::extract_json_from_stream`, so the behavior is consistent across the codebase. Nested objects and escaped quotes are handled correctly.

## Tests

Added `tests/dsl/test_iterable_get_object.py` (pure unit tests, no API calls): braces inside strings, escaped quote before a brace, nested objects, and the full `tasks_from_chunks` streaming path. All four **fail on `main`** and **pass with the fix**; the existing `dsl`/`processing` suites remain green. `ruff check`, `ruff format --check`, and `ty check` all pass. CHANGELOG updated.

---

*Authored with the assistance of an AI tool; the bug was reproduced, the fix written, and all tests/lint/type-checks verified by me before submitting.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Streaming (Iterable)**: `IterableBase.get_object` now ignores `{`/`}` characters that appear inside JSON string values, so `create_iterable()` / `Iterable[T]` streams no longer truncate an object mid-string and raise a `ValidationError` when a string field contains a brace (e.g. `":}"`). ([#2339](https://github.com/instructor-ai/instructor/pull/2339))
+
 ---
 
 ## [1.15.2] - 2026-05-10
--- a/instructor/v2/dsl/iterable.py
+++ b/instructor/v2/dsl/iterable.py
@@ -173,10 +173,26 @@ class IterableBase:
     @staticmethod
     def get_object(s: str, stack: int) -> tuple[Optional[str], str]:
         start_index = s.find("{")
+        in_string = False
+        escape_next = False
         for i, c in enumerate(s):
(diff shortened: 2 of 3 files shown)


jxnl:
Thanks for the focused fix. We are consolidating small streaming/JSON parser fixes into a maintainer PR for the next minor release, so I am closing this PR rather than merging individual narrow PRs. The braces-inside-strings streaming case will be covered there.

PR #2346, opened 2026-06-05
docs: add Entroly reference to prompt caching guide

juyterman1000:
## What this PR adds

Adds [Entroly](https://github.com/juyterman1000/entroly) as a See Also reference in the Prompt Caching documentation.

### Why it fits:

The prompt caching doc explains how to manually set `cache_control` to get Anthropic's cache discount. Entroly automates the complementary problem: **keeping the context prefix byte-stable** so the cache actually hits across requests.

Without prefix alignment, even with `cache_control` set, context re-ranking or dynamic content changes bust the cache. Entroly's Cache Aligner solves this transparently.

Additionally, Entroly compresses input context by 70-95%, reducing the tokens that need to be cached in the first place.

### Change:
- Added one line to the See Also section in `docs/concepts/prompt_caching.md`

Apache-2.0, local-first, no outbound analytics by default.


jordandunmire97-ai (approved):

jxnl:
Closing as an unsupported promotional/resource-only addition. The single link advertises an external proxy and a broad 70-95% compression claim without an Instructor-owned integration, maintained example, or regression surface. This is intentionally not part of the 1.16 documentation consolidation.

juyterman1000:
@jxnl Thanks for the clarification. We agree the original change was resource-only and did not provide an Instructor-owned integration, example, or regression surface.
For the record—not as a request to reopen—the original compression statement should have included its scope and evidence. On four deterministic agent-tool fixtures, Entroly 1.0.59 source reduced weighted input tokens by 95.1% while retaining all preregistered answer evidence under a shared tokenizer.
Benchmark report: https://github.com/juyterman1000/entroly/blob/main/docs/BENCHMARKS.md#same-input-compression-gauntlet
Raw artifact: https://github.com/juyterman1000/entroly/blob/main/benchmarks/results/compression_gauntlet.json

PR #2351, opened 2026-06-05
chore: improve instructor maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/test_utils.py, tests/v2/conftest.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -197,6 +197,23 @@ def test_classproperty():
     assert MyClass.my_property == 1
 
 
+def test_classproperty_instance_and_subclass():
+    """Test classproperty on instances and subclasses."""
+
+    class Base:
+        @classproperty
+        def prop(cls):
+            return cls.__name__
+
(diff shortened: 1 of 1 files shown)


jordandunmire97-ai (approved):

jordandunmire97-ai (approved):

jxnl:
Closing as behavior-free test-only maintenance. Current `main` already exercises `classproperty` binding in both utility test modules, and this patch does not reproduce a bug, change an implementation contract, or cover a previously failing path. The added subclass/instance assertions follow directly from the existing descriptor behavior and do not justify another duplicate test.

jordandunmire97-ai (approved):
I will see

PR #2361, opened 2026-06-10
fix(multimodal): raise explicit errors in Audio instead of assert

IgnazioDS:
## What & why

`Audio.from_url` and `Audio.from_path` validate inputs with bare `assert` statements:

```python
assert content_type in VALID_AUDIO_MIME_TYPES, "..."
assert path.is_file(), "..."
assert mime_type in VALID_AUDIO_MIME_TYPES, "..."
```

Two problems:

1. **`assert` is stripped under `python -O`** (PEP 230), so these validation checks silently vanish in optimized deployments — invalid audio then flows downstream instead of being rejected at the boundary.
2. When they *do* fire, they raise `AssertionError` rather than the `ValueError` callers expect.

This also leaves `Audio` inconsistent with its sibling classes: `Image.from_path` and `PDF.from_path` already raise `FileNotFoundError` for a missing file, `ValueError("... file is empty")` for an empty file, and `ValueError("Unsupported ... format: ...")` for an unsupported type.

## Change

Replace the asserts in `Audio` with the same explicit errors `Image`/`PDF` use:

- missing file → `FileNotFoundError`
- empty file → `ValueError("Audio file is empty")` *(new check, matching `Image`/`PDF`)*
- unsupported format (URL or path) → `ValueError("Unsupported audio format: ...")`

No behavior change for valid inputs.

## Tests

Adds 4 regression tests in `tests/v2/test_core_multimodal_runtime.py` — each fails on `main` and passes here:

- `from_url` with an unsupported content-type → `ValueError`
- `from_path` with an unsupported format → `ValueError`
- `from_path` with an empty file → `ValueError`
- `from_path` with a missing file → `FileNotFoundError`

Verification: `pytest tests/v2/test_core_multimodal_runtime.py` → 18 passed; `tests/multimodal/test_multimodal.py` (non-LLM) → 57 passed; `ruff check` / `ruff format --check` / `ty check` clean. `CHANGELOG.md` updated under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Fixed
 - **v2 cleanup**: Consolidate small provider/runtime fixes for Gemini JSON prompts, Cohere templating, JSON array extraction, iterable streaming, missing `jsonref` dependency guidance, retry semantics and hook metadata, and multimodal autodetection.
+- **Multimodal (Audio)**: `Audio.from_url` and `Audio.from_path` now raise `ValueError` (unsupported/empty) and `FileNotFoundError` (missing) instead of bare `assert` statements, which are silently stripped under `python -O`. Brings `Audio` validation in line with `Image`/`PDF`. ([#2361](https://github.com/567-labs/instructor/pull/2361))
 
 ### Tests / CI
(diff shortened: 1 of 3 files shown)


jxnl:
Consolidated this Audio validation fix into draft PR #2400. Closing this PR to keep the queue focused on the consolidated patch; the new PR carries the explicit `ValueError` / `FileNotFoundError` behavior and focused regression tests.

PR #2362, opened 2026-06-10
fix: warn against combining RESPONSES_TOOLS with text.format json_schema (closes #2300)

### LINKED ISSUE #2300: Responses API migration pain: RESPONSES_TOOLS can ignore structured text format and return empty `{}` tool args
- [ ] This is actually a bug report.
- [ x] I am not getting good LLM Results
- [ ] I have tried searching the documentation and have not found an answer.

**What Model are you using?**
- [ x] gpt-5-nano

**Description**
Hi Instructor team, thanks for the great library.
OpenAI is recommending migration from Chat Completions to the Responses API (https://developers.openai.com/api/docs/guides/migrate-to-responses) , so we moved our production pipeline accordingly. During that migration, we found a reliability issue in `RESPONSES_TOOLS` mode.

Environment:
- `instructor==1.14.4`
- OpenAI Python SDK `1.x`
- Azure OpenAI
- Python `3.11`
- Model: `gpt-5-nano`

Context:

- With Chat Completions + Instructor, our Pydantic extraction flow was stable.
- We migrated because OpenAI guidance suggests using Responses API going forward.
- In Responses API mode, we now see intermittent empty tool arguments (`{}`), which causes validation failures.

What we are sending:
- `mode=instructor.Mode.RESPONSES_TOOLS`
- `response_model=ResponseModel` (Pydantic)
- `text.format` set to `json_schema` (strict true/false depending on schema compatibility)

Observed behavior:
- Response includes both:
  - forced function tool call for `ResponseModel`
  - `text` config present
- But generation often returns tool call `arguments='{}'`
- Instructor then fails Pydantic validation (required fields missing)
- This appears more often when reasoning tokens are low

Representative failure:
- `ResponseFunctionToolCall(arguments='{}', name='ResponseModel', ...)`
- Validation error similar to:
  - `1 validation error for ResponseModel`
  - required top-level field missing (example: `data`)

Expected behavior:
- If `text.format` is provided, either:
  1. it is honored consistently, or
  2. Instructor rejects this combination with a clear error/warning.
- In `RESPONSES_TOOLS`, required schema fields should not degrade into empty `{}` tool args.
- Migration path from Chat Completions to Responses should preserve structured extraction reliability.

Why this matters:
- OpenAI recommends Responses API over Chat Completions, but current behavior makes migration risky for schema-driven extraction pipelines.
- We can’t reliably depend on required structured output if tool args can collapse to `{}`.
- The constan retries make the requests too expensive.

Request / possible improvements:

1. Add a Responses mode that uses `text.format` as the primary structured output path (no forced function call).
2. In `RESPONSES_TOOLS`, harden schema/tool argument enforcement to prevent empty `{}` when required fields exist.
3. Add explicit conflict handling for `text.format` + `RESPONSES_TOOLS` (document precedence and/or fail fast).
4. Add retry heuristic for empty tool args when response_model has required fields.

botbikamordehai2-sketch:
## What
In RESPONSES_TOOLS mode, setting text.format to json_schema can cause the model to return empty tool arguments (`{}`), resulting in Pydantic validation failures. This is due to the model treating tool calls as optional when a structured text format is also requested.

## Fix
Added a documentation note in the Mode enum clarifying that this combination is discouraged. The actual fix should be implemented in the response handling code for RESPONSES_TOOLS to either:
- Detect empty arguments and raise a clear ValidationError
- Retry with forced tool call prompting
- Reject the combination at configuration time

This PR addresses the immediate need for user awareness, but the core fix should be applied in the response parsing logic (see related PR for the handler).

Closes #2300

### DIFF
--- a/instructor/v2/core/mode.py
+++ b/instructor/v2/core/mode.py
@@ -29,6 +29,10 @@ class Mode(enum.Enum):
     RESPONSES_TOOLS = "responses_tools"
     RESPONSES_TOOLS_WITH_INBUILT_TOOLS = "responses_tools_with_inbuilt_tools"
 
+    # Note: When using RESPONSES_TOOLS, do NOT set text.format to json_schema
+    # as this combination is known to produce empty tool arguments ({}) intermittently.
+    # The handling code must detect empty args and raise a ValidationError or retry.
+
     # XAI modes
     XAI_JSON = "xai_json"
     XAI_TOOLS = "xai_tools"


jxnl:
Closing as superseded by the merged #2399, which closed #2300 by aligning RESPONSES_TOOLS text.format with the forced tool schema and adding targeted empty-args retry guidance/warnings in the 1.15.5 patch work.

PR #2364, opened 2026-06-11
fix(multimodal): support bytes in Image.autodetect, reject unsupported types

### LINKED ISSUE #2344: Image.autodetect returns None for non-str/non-Path sources, causing a confusing AttributeError
`Image.autodetect(source)` is annotated `-> Image` but only handles `str` and `Path`. For any other source (e.g. `bytes`) it falls off the end and implicitly returns `None`. Callers then access `.source`/`.media_type`/`.data` and get a misleading `AttributeError: 'NoneType' object has no attribute 'source'`.

```python
from instructor.v2.core.multimodal import Image
Image.autodetect(b"\xff\xd8\xff").source  # AttributeError: 'NoneType' ...
```

This is reachable from `ImageWithCacheControl.from_image_params` (and `convert_messages(..., autodetect_images=True)`), whose guard accepts a `source` of any type. The sibling `Audio.autodetect` already raises `ValueError` on an unknown source, so `Image` should too.

by-Kaimercer:
## Summary

`Image.autodetect()` previously returned `None` for non-str/non-Path sources (e.g. `bytes`), causing a confusing `AttributeError: 'NoneType' object has no attribute 'source'` downstream.

This fix:
- Adds `bytes` support: base64-encodes the bytes, then detects format from magic bytes (JPEG/PNG/GIF/WebP)
- Raises `ValueError` for truly unsupported types instead of silently returning `None`
- Matches the existing `Audio.autodetect()` behavior which already raises `ValueError`

## Changes

- `instructor/v2/core/multimodal.py`: Add `bytes` branch to `Image.autodetect()`, update type annotations on `Image.source` and `autodetect_safely()`
- `tests/v2/test_core_multimodal_runtime.py`: Add 6 new tests for bytes input (JPEG, PNG, GIF, WebP, unsupported bytes, unsupported type), update existing parametrize to no longer expect Image to reject bytes
- `CHANGELOG.md`: Add entry under [Unreleased] > Fixed

## Testing

- `test_image_autodetect_bytes_jpeg` — JPEG magic bytes → `image/jpeg`
- `test_image_autodetect_bytes_png` — PNG magic bytes → `image/png`
- `test_image_autodetect_bytes_gif` — GIF magic bytes → `image/gif`
- `test_image_autodetect_bytes_webp` — WebP magic bytes → `image/webp`
- `test_image_autodetect_bytes_unsupported_type` — non-image bytes → `ValueError`
- `test_image_autodetect_unsupported_type_raises` — dict input → `ValueError`

Fixes #2344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Fixed
 - **v2 cleanup**: Consolidate small provider/runtime fixes for Gemini JSON prompts, Cohere templating, JSON array extraction, iterable streaming, missing `jsonref` dependency guidance, retry semantics and hook metadata, and multimodal autodetection.
+- **Image.autodetect**: Accept `bytes` input (JPEG/PNG/GIF/WebP detected from magic bytes) and raise `ValueError` for unsupported types instead of silently returning `None`. ([#2344](https://github.com/567-labs/instructor/issues/2344))
 
 ### Tests / CI
 - **Type checking**: Upgrade to `ty` 0.0.44, enforce warning-free checks with GitHub annotations, cover V2 tests, validate supported Python versions and platforms, and strengthen installed-package public API typing tests.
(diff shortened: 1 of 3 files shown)


jxnl:
Consolidated in #2398 with raw-byte image autodetection tests.

PR #2372, opened 2026-06-15
fix(v2): avoid Python 3.9 runtime TypeError from PEP 604 unions in cast() calls

aqilaziz:
## Problem

`instructor/v2/auto_client.py` passes PEP 604 union syntax (`X | None`) as the **first argument** to `typing.cast()` inside `_build_openai()`:

- `cast(str | None, ...)` (organization)
- `cast(float | Timeout | None, ...)` (timeout)
- `cast(Mapping[str, str] | None, ...)` (default_headers)
- `cast(Mapping[str, object] | None, ...)` (default_query)
- `cast(httpx.AsyncClient | None, ...)` / `cast(httpx.Client | None, ...)` (http_client)

Unlike annotations, the first argument of `cast()` is **evaluated at runtime**. `from __future__ import annotations` (present at the top of the file) only defers *annotation* evaluation; it does not affect a normal call argument. On Python 3.9, evaluating `X | None` raises:

```
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

The project supports Python 3.9 (`requires-python = "<4.0,>=3.9"`), so any 3.9 user calling `instructor.from_provider("openai/...")` crashes here. CI only runs on 3.11+, so it isn't caught.

Minimal reproduction on Python 3.9:

```python
>>> from typing import cast
>>> cast(str | None, "x")
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

## Fix

Replace the runtime PEP 604 unions **inside `cast()`** with `typing.Optional` / `typing.Union`, which evaluate fine on 3.9. Annotations elsewhere are left untouched (they're safe under `from __future__ import annotations`). `cast(int, ...)` and `cast(Any, ...)` are already 3.9-safe and unchanged.

## Verification

- `ruff check` with the repo `.ruff.toml` (target `py39`, `select = ["UP", ...]`): **passes** — `Optional[...]`/`Union[...]` inside `cast()` are not re-flagged by pyupgrade.
- `ruff format --check`: already formatted.
- `python3.9 -m py_compile`: passes.
- Reproduced the original `TypeError` on 3.9 and confirmed the `Optional` form runs cleanly.

Distinct from #2371 (which fixed *import-time* annotation evaluation in v2 modules); this fixes *call-time* `cast()` evaluation in the OpenAI provider builder.

### DIFF
--- a/instructor/v2/auto_client.py
+++ b/instructor/v2/auto_client.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import importlib
-from typing import Any, Callable, Literal, Union, cast, overload
+from typing import Any, Callable, Literal, Optional, Union, cast, overload
 from instructor.v2.core.client import AsyncInstructor, Instructor
 from instructor import __version__
 from instructor.v2.core.mode import Mode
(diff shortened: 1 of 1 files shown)


jxnl:
Thanks for the focused Python 3.9 `typing.cast()` runtime fix. This was consolidated and merged in #2373 as part of the 1.15.3 release branch, so I’m closing this PR as absorbed.

PR #2377, opened 2026-06-19
fix(v2): treat list[A | B] PEP 604 unions as iterable of models

azizmabrouk11:
## Problem

`list[A | B]` (PEP 604 union syntax) is the idiomatic modern way to ask for "a list of A-or-B" and should behave identically to `list[Union[A, B]]`. It didn't:

| `response_model` | prepared as | schema root |
| --- | --- | --- |
| `list[Union[A, B]]` | `IterableModel` ✅ | `tasks` (array) |
| `Iterable[A \| B]` | `IterableModel` ✅ | `tasks` (array) |
| `list[A \| B]` | simple-type wrapper ❌ | `content` (single object) |

`A | B` reports `types.UnionType` as its origin (not `typing.Union`), so `prepare_response_model._is_model_type` didn't recognize it as a union of models and `list[A | B]` fell through to the simple-type `ModelAdapter`. This silently changes the JSON schema sent to the model and breaks multi-item extraction — no error, just wrong results.

## Reproduction

```python
from typing import Union
from instructor.v2.core.response_model import prepare_response_model

class A(BaseModel): x: int
class B(BaseModel): y: str

prepare_response_model(list[A | B]).model_json_schema()        # before: root "content" (wrong)
prepare_response_model(list[Union[A, B]]).model_json_schema()  # root "tasks" (correct)
```

## Fix

Recognize `types.UnionType` alongside `typing.Union` (Python 3.10+), following the existing convention in `dsl/parallel.py`:

1. `prepare_response_model._is_model_type` — routes `list[A | B]` to `IterableModel`.
2. `IterableModel` task-name derivation — gives pipe unions a clean name (`IterableAOrB`) instead of leaking `Iterable<module>.A | <module>.B` into the schema title.

Extends the PEP 604 support added for `Partial` in #2200 to list/iterable response models.

## Tests

Two regression tests in `tests/dsl/test_simple_types.py` (skipped < 3.10), verified red-green. Mixed unions (`list[A | int]`) stay consistent with `list[Union[A, int]]` (both simple).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **Response models**: `list[A | B]` written with PEP 604 union syntax is now treated as an iterable of models, identical to `list[Union[A, B]]`. Previously the `|` form was misrouted to the simple-type wrapper (root property `content`) instead of an iterable (root property `tasks`), silently changing the JSON schema sent to the model and breaking multi-item extraction. Pipe-union iterables also get a clean generated name (e.g. `IterableAOrB`) instead of leaking a module-qualified `str(types.UnionType)`.
+
+---
+
 ## [1.15.3] - 2026-06-15
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


jxnl:
Closing as superseded by #2399, which includes the PEP 604 list[A | B] iterable response-model fix and regression coverage for the 1.15.5 patch candidate.

PR #2382, opened 2026-06-19
fix(v2): prepare response_model before handler dispatch in patch.py

### LINKED ISSUE #2374: Pydantic related error: type object 'list' has no attribute 'model_json_schema'
- [ ] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

- [ ] gpt-3.5-turbo
- [ ] gpt-4-turbo
- [ ] gpt-4
- [x] Other (please specify) gemini

**Describe the bug**
Since the release of 1.15.3 we are getting this error:  type object 'list' has no attribute 'model_json_schema'
Unfortunatelly we use Databricks and they disabled showing the stack traces, so I do not have that at this point.

**To Reproduce**
Steps to reproduce the behavior, including code snippets of the model and the input data and openai response.
Not sure which aspect of our pydantic model causes this, I cannot copy the whole code, if you ask specific question happy to help.

**Expected behavior**
Should work like with 1.15.1

**Screenshots**

azrirefik:
## Summary

Fixes #2374

`patch.py` was not calling `prepare_response_model()` before dispatching to provider handlers, unlike `response.py` which does (`response.py:416`). This is a v2 refactor oversight.

Gemini handlers are the only providers that don't call `prepare_response_model` internally — all others (OpenAI, Anthropic, GenAI, Cohere, xAI, Bedrock, Writer, Mistral) call it in their `prepare_request` methods. As a result, raw `list[Model]` types reached:
- Cache key generation (`cache/__init__.py:174`) → `model_json_schema()` on `list` → `AttributeError`
- Gemini TOOLS handler → `response_model.gemini_schema` on `list` → `AttributeError`

## Fix

Added `prepare_response_model(response_model)` call in both sync and async wrappers in `patch.py`, mirroring `response.py:416`. Safe because `prepare_response_model` is idempotent (skips already-prepared models via `issubclass` check).

## Testing

- 3 regression tests in `tests/v2/test_issue_2374.py`:
  - Unit test: `prepare_response_model(list[User])` returns `IterableBase` subclass
  - Regression: patched create with `list[User]` + `AutoCache()` + Gemini MD_JSON
  - Regression: patched create with `list[User]` + Gemini TOOLS
- All 3 tests fail without the fix, pass with it
- Full suite: no regressions (71 pre-existing failures are all missing API key errors)
- `ty check` clean, `ruff check` + `ruff format` clean

## Checklist

- [x] Fix applied
- [x] Tests added and passing
- [x] CHANGELOG.md updated
- [x] `ty check` passes
- [x] `ruff check` + `ruff format` pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **v2 core**: Call `prepare_response_model()` in `patch.py` before dispatching to provider handlers, fixing `AttributeError: type object 'list' has no attribute 'model_json_schema'` when using `response_model=list[Model]` with Gemini (and other providers whose handlers do not internally prepare the model). This restores parity with `response.py` and resolves a regression introduced in the v2 refactor ([#2374](https://github.com/567-labs/instructor/issues/2374)).
+
+---
+
 ## [1.15.3] - 2026-06-15
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


jxnl:
Consolidated in #2398 with regression coverage for #2374.

PR #2385, opened 2026-06-21
fix(v2): preserve backticks inside JSON string values during stream extraction

hoangsonww:
## Describe your changes

`extract_json_from_stream` and `extract_json_from_stream_async` (`instructor/v2/core/json.py`) treat any backtick as a potential Markdown code-fence delimiter, even while inside a JSON string value. For an unfenced streamed response, backticks inside a string were silently diverted into the code-fence buffer and dropped, corrupting the extracted content while still producing valid-looking JSON:

~~~text
{"md":"use `pip install`"}   ->   {"md":"use pip install"}
{"code":"```py```"}          ->   {"code":"py```"}
~~~

This is common in practice, since models routinely emit code and markdown (with backticks) inside string fields.

The fix only treats a backtick as a code-fence candidate when not inside a JSON string (`not (json_started and in_string)`). Inside a string, backticks are appended as literal content. It is applied identically to the sync and async extractors.

None. Bug found while reading the streaming extractor.

Added offline unit tests in `tests/v2/test_json_helpers.py` (inline and triple backticks in unfenced streamed JSON, sync and async, plus a non-regression test that backticks survive inside a fenced block). Verified red/green: the new tests fail on `main` and pass with the fix; the existing JSON-extraction suites (95 tests) still pass. `ruff check`, `ruff format --check`, and `ty check` are clean. Added an `[Unreleased]` to `Fixed` entry in `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ---
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


jxnl:
Consolidated in #2398 with regression tests for streamed JSON backticks.

Issue #2205, opened 2026-03-23
RAM Spike when I import instructor


AbdealiLoKo:
Hi, I've been using the SDKs for my LLMs (`openai`, `google.genai`, etc.) so far which I use in a fastapi application I am working on which aims to support multiple LLM providers.
But recently got introduced to instructor which I was hoping can reduce some of my boiler plate code.

When I tried `instructor` I found my RAM utilization was spiking. When I debugged, I found that `instructor` was the cause.

Basically RAM stats when I import:
1. `sys` (python builtin) -> 15MB
2. `instructor` -> ~290MB (19x more)
3. `google.genai` -> ~90MB
4. `openai` -> ~54MB
5. `langgraph/langchain` -> ~15MB

Any other library doesn't seem to be taking more than ~90MB making instructor 3x more RAM intensive than others

Reproducible snippets:
```py
$ /usr/bin/time venv/bin/python -c "import sys"
0.02user 0.00system 0:00.03elapsed 91%CPU (0avgtext+0avgdata 15104maxresident)k
0inputs+0outputs (0major+1724minor)pagefaults 0swaps

$ /usr/bin/time venv/bin/python -c "import numpy"
1.16user 0.02system 0:00.21elapsed 547%CPU (0avgtext+0avgdata 30680maxresident)k
0inputs+0outputs (0major+3966minor)pagefaults 0swaps

$ /usr/bin/time venv/bin/python -c "import instructor"
7.64user 1.23system 0:08.89elapsed 99%CPU (0avgtext+0avgdata 290912maxresident)k
0inputs+0outputs (0major+83542minor)pagefaults 0swaps

$ /usr/bin/time venv/bin/python -c "import google.genai"
0.99user 0.11system 0:01.11elapsed 99%CPU (0avgtext+0avgdata 90708maxresident)k
0inputs+0outputs (0major+18901minor)pagefaults 0swaps

$ /usr/bin/time venv/bin/python -c "import openai"
0.80user 0.07system 0:00.87elapsed 99%CPU (0avgtext+0avgdata 54152maxresident)k
0inputs+0outputs (0major+10622minor)pagefaults 0swaps

$ /usr/bin/time venv/bin/python -c "import langgraph"
0.03user 0.00system 0:00.03elapsed 94%CPU (0avgtext+0avgdata 14976maxresident)k
0inputs+0outputs (0major+1723minor)pagefaults 0swaps

$ /usr/bin/time venv/bin/python -c "import langchain"
[...]

PR #2390, opened 2026-06-24
perf: defer openai SDK import in v2 core modules

mimran-khan:
Partially addresses #2205.

## Problem

With `openai`, `anthropic`, and `google-genai` installed, accessing `instructor.from_provider` or `instructor.Instructor` eagerly loaded the entire openai SDK (~30MB, 558 modules) even though nothing OpenAI-specific was needed yet. The root cause: several `v2/core/` modules imported `openai` at the module level for type annotations and isinstance checks that only matter at runtime when an OpenAI client is actually in play.

## Fix

Moved openai imports behind `TYPE_CHECKING` in four files:

- **`v2/core/client.py`** - `import openai` + `ChatCompletionMessageParam` (used only in annotations thanks to `from __future__ import annotations`; two isinstance checks now use a local import inside the conditional block that already guards for RESPONSES_TOOLS mode)
- **`v2/core/patch.py`** - `AsyncOpenAI`, `OpenAI` (annotation-only)
- **`v2/core/messages.py`** - `ChatCompletionMessage`, `ChatCompletionMessageParam` (annotation-only)
- **`v2/core/usage.py`** - `CompletionUsage` (moved to local import inside `update_total_usage()` where the isinstance check lives)

## Results

Benchmarked on Python 3.12 with `openai`, `anthropic`, and `google-genai` installed:

| Access pattern | Before | After | Reduction |
|---|---|---|---|
| `instructor.from_provider` | 26.7 MB / 942 modules | 13.0 MB / 304 modules | **51% memory, 68% modules** |
| `instructor.Instructor` | 29.0 MB / 976 modules | 14.8 MB / 338 modules | **49% memory, 65% modules** |
| `import instructor` | 0.6 MB / 78 modules | 0.6 MB / 78 modules | (already lazy) |

Provider-specific access (`from_openai`, `from_anthropic`, etc.) still correctly loads the respective SDK on demand - that's expected and intentional.

## Included

- Benchmark script (`scripts/benchmark_import.py`) for reproducible profiling
- Regression test (`tests/test_import_lazy_openai.py`) verifying openai isn't pulled in by core symbol access

## Scope

This is intentionally a focused fix for the `v2/core/` import chain. The top-level `__init__.py` lazy loading (already on main) handles the outer layer; this PR closes the gap in the inner layer that still triggered openai through `core/__init__.py` → `patch.py` → `messages.py` / `usage.py`.

### DIFF
--- a/instructor/v2/core/client.py
+++ b/instructor/v2/core/client.py
@@ -1,10 +1,7 @@
 from __future__ import annotations
 
-import openai
-from instructor.v2.core.mode import Mode
-from instructor.v2.core.providers import Provider
-from openai.types.chat import ChatCompletionMessageParam
 from typing import (
+    TYPE_CHECKING,
     TypeVar,
     Callable,
     overload,
@@ -14,6 +11,13 @@ from typing import (
     get_origin,
     get_args,
 )
+
+if TYPE_CHECKING:
+    import openai
+    from openai.types.chat import ChatCompletionMessageParam
+
+from instructor.v2.core.mode import Mode
+from instructor.v2.core.providers import Provider
 from tenacity import (
     AsyncRetrying,
     Retrying,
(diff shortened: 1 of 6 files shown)


jxnl:
Closing as superseded by #2399, which consolidates the lazy OpenAI import work with the union response-model and RESPONSES_TOOLS fixes for the 1.15.5 patch candidate.

PR #2296, opened 2026-05-02
feat(core): add token_budget parameter to limit retry token usage

Jwrede:
## Summary

Refs #2056

Adds an optional `token_budget` parameter that stops retries when cumulative token usage exceeds the specified limit. This mitigates the retry amplification vector where adversarial or malformed LLM outputs cause repeated validation failures, leading to unbounded context growth and cost amplification.

## Usage

```python
client.chat.completions.create(
    response_model=MyModel,
    max_retries=10,
    token_budget=5000,  # stop retrying after 5000 total tokens used
)
```

When the budget is exceeded between retry attempts, the retry loop stops and raises `InstructorRetryException` (same as when `max_retries` is exhausted).

## Implementation

- Adds `token_budget` and `usage_container` parameters to `initialize_retrying()`
- Uses a closure-based tenacity stop condition that checks cumulative usage (supports both OpenAI `total_tokens` and Anthropic `input_tokens + output_tokens`)
- Extracts `token_budget` from kwargs in both `retry_sync` and `retry_async` (same pattern as existing `timeout`)
- No breaking changes -- parameter is optional, behavior is unchanged when not set

## Test plan

- [x] `test_token_budget_stops_retries` -- verifies retries stop before max_retries when budget exceeded
- [x] `test_token_budget_not_set_retries_normally` -- verifies default behavior unchanged
- [x] `test_token_budget_success_before_limit` -- verifies successful responses return normally
- [x] Existing retry tests pass (`test_retry_json_mode.py`)
- [x] ruff check + format clean

PR #2392, opened 2026-06-24
feat: expose cumulative token usage on success + budget enforcement

### LINKED ISSUE #2391: feat: expose cumulative token usage on successful responses + budget enforcement
## Problem

When retries succeed, the accumulated `total_usage` (token counts across all retry attempts) is computed internally but never exposed to the user. It's only available when retries **fail** (via `InstructorRetryException.total_usage`).

This means:
- If you retry 3 times and succeed on the 4th, there's no way to know total tokens consumed
- There's no mechanism to cap runaway retry costs (mentioned in #2056 as a production concern)
- The `completion:usage` telemetry hook doesn't exist, so observability tools can't track per-attempt token accumulation

## Proposed Solution

1. **Attach `_total_usage` to successful responses** - just like `_raw_response` is already attached to parsed models, attach the cumulative usage object so users can inspect it after a successful call.

2. **Add `completion:usage` hook** - fires after each API attempt with the running total of tokens consumed. Enables integration with observability platforms without modifying core logic.

3. **Add `token_budget` parameter** - optional parameter on `.create()` that raises `TokenBudgetExceeded` if cumulative tokens across all attempts exceed the budget. This prevents runaway costs from repeated retries against complex schemas.

## Usage Example

```python
import instructor
from openai import OpenAI

# After extraction, check how many tokens were used across all retries
user = client.chat.completions.create(
    model="gpt-4o",
    response_model=User,
    messages=[...],
    max_retries=5,
    token_budget=2000,  # abort if cumulative tokens exceed 2000
)

# Access total usage (available on success, not just failure)
print(f"Total tokens: {user._total_usage.total_tokens}")
```

```python
# Observability via hooks
def on_usage(usage, *, attempt_number=0):
    metrics.record("instructor.tokens", usage.total_tokens, attempt=attempt_number)

client.on("completion:usage", on_usage)
```

## Context

- Directly addresses production concern raised in #2056 (retry amplification leading to hundreds of dollars/day)
- The infrastructure already exists (usage tracking in `retry.py`, hooks system) - just needs to be connected

mimran-khan:
Fixes #2391. Related to #2056.

## Problem

The retry system tracks cumulative token usage across all attempts, but this data is only accessible when retries **fail** (via `InstructorRetryException.total_usage`). On success, it's computed and thrown away. Users have no visibility into how expensive a successful extraction actually was, and no way to cap runaway costs from complex schemas that trigger many retries.

This came up in the #2056 discussion where someone mentioned losing "hundreds of dollars a day" from retries they couldn't observe or control.

## What this PR does

Three things, all backward-compatible:

**1. `_total_usage` on successful responses**

After a successful extraction, the parsed model now has `_total_usage` attached (same pattern as `_raw_response`):

```python
user = client.chat.completions.create(
    model="gpt-4o",
    response_model=User,
    messages=[...],
    max_retries=5,
)
print(f"Total tokens across all retries: {user._total_usage.total_tokens}")
```

**2. `completion:usage` hook**

New hook that fires after each API attempt with the running total. Enables integration with metrics/observability without touching core logic:

```python
def on_usage(usage, *, attempt_number=0):
    metrics.gauge("instructor.tokens.cumulative", usage.total_tokens)

client.on("completion:usage", on_usage)
```

**3. `token_budget` parameter**

Optional parameter that raises `TokenBudgetExceeded` if cumulative tokens exceed the limit:

```python
from instructor.v2.core.errors import TokenBudgetExceeded

try:
    user = client.chat.completions.create(
        response_model=ComplexSchema,
        max_retries=10,
        token_budget=5000,  # hard cap across all attempts
        ...
    )
except TokenBudgetExceeded as e:
    print(f"Aborted: used {e.total_usage.total_tokens} tokens in {e.n_attempts} attempts")
```

## Files changed

- `instructor/v2/core/hooks.py` - added `COMPLETION_USAGE` hook name + `CompletionUsageHandler` protocol + `emit_completion_usage()` method
- `instructor/v2/core/errors.py` - added `TokenBudgetExceeded` exception
- `instructor/v2/core/retry.py` - emit usage hook, check budget, attach `_total_usage` on success, let `TokenBudgetExceeded` escape the retry wrapper
- `instructor/v2/core/patch.py` - pass `token_budget` from `create()` down to retry logic
- `tests/test_token_budget.py` - 6 tests covering all new behavior

## Backward compatibility

All changes are additive. Existing code is unaffected:
- `token_budget` defaults to `None` (no enforcement)
- `_total_usage` is only attached, never required
- The new hook only fires if handlers are registered

### DIFF
--- a/instructor/v2/core/errors.py
+++ b/instructor/v2/core/errors.py
@@ -255,6 +255,39 @@ class InstructorRetryException(InstructorError):
         super().__init__(*args, failed_attempts=failed_attempts, **kwargs)
 
 
+class TokenBudgetExceeded(InstructorError):
+    """Raised when cumulative token usage exceeds the configured budget.
+
+    This allows users to cap the total cost of retries. When a token_budget
+    is set and the cumulative tokens across all attempts exceed it, this
+    exception is raised instead of continuing to retry.
+
+    Attributes:
+        total_usage: Cumulative token usage at the time of budget breach
+        budget: The configured token budget that was exceeded
+        n_attempts: Number of attempts made before budget was exceeded
+    """
+
+    def __init__(
+        self,
+        *args: Any,
+        total_usage: Any,
+        budget: int,
+        n_attempts: int,
+        last_completion: Any | None = None,
+        **kwargs: Any,
+    ):
+        self.total_usage = total_usage
+        self.budget = budget
+        self.n_attempts = n_attempts
+        self.last_completion = last_completion
+        message = (
+            f"Token budget exceeded: used {getattr(total_usage, 'total_tokens', '?')} "
+            f"tokens across {n_attempts} attempts (budget: {budget})"
+        )
+        super().__init__(message, *args, **kwargs)
+
+
 class ValidationError(InstructorError):
     """Exception raised when LLM response validation fails.
 
--- a/instructor/v2/core/hooks.py
+++ b/instructor/v2/core/hooks.py
@@ -15,6 +15,7 @@ class HookName(Enum):
     COMPLETION_RESPONSE = "completion:response"
     COMPLETION_ERROR = "completion:error"
     COMPLETION_LAST_ATTEMPT = "completion:last_attempt"
+    COMPLETION_USAGE = "completion:usage"
     PARSE_ERROR = "parse:error"
 
 
@@ -44,6 +45,20 @@ class CompletionErrorHandler(Protocol):
     ) -> None: ...
 
 
+class CompletionUsageHandler(Protocol):
+    """Protocol for completion usage handlers.
+
+    Fired after each API attempt with cumulative token usage.
+    """
+
+    def __call__(
+        self,
+        usage: Any,
+        *,
+        attempt_number: int = ...,
+    ) -> None: ...
+
+
 class ParseErrorHandler(Protocol):
     """Protocol for parse error handlers."""
 
@@ -58,6 +73,7 @@ HookNameType = Union[
         "completion:response",
         "completion:error",
         "completion:last_attempt",
+        "completion:usage",
         "parse:error",
     ],
 ]
@@ -67,6 +83,7 @@ HandlerType = Union[
     CompletionKwargsHandler,
     CompletionResponseHandler,
     CompletionErrorHandler,
+    CompletionUsageHandler,
     ParseErrorHandler,
 ]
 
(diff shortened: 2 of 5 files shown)


jxnl:
Reviewed for merge readiness. Directionally this is the right PR to keep for #2391 / the #2056 retry-cost thread, and I closed the older conflicted #2296 in favor of this one. Before merge, I would like this refreshed with CI and one follow-up check: make sure cumulative usage is exposed consistently for non-BaseModel successful response shapes too, especially `list[Model]` / `ListResponse`, or explicitly document that `_total_usage` is only attached to `BaseModel` responses. The hook and `token_budget` path should also have async coverage matching the sync tests.

mimran-khan:
Thanks for the review and for closing #2296 in favor of this. Pushed an update addressing both points:

**ListResponse coverage**: `_total_usage` is now attached to `ListResponse` results (the `list[Model]` path). The `_finalize_parsed_response` helper handles both the case where the parser returns a plain list (converted to `ListResponse`) and where it returns a `ListResponse` directly. Added 3 dedicated sync tests for this path.

**Async coverage**: Added 5 async tests matching the sync suite - usage attachment, hook firing, budget exceeded, under-budget success, and list response attachment. All use `asyncio.run()` with `AsyncMock` to exercise `retry_async_v2` directly.

**Note on AdapterBase**: For primitive return types (like `str`, `int` via `AdapterBase`), Python doesn't allow attaching attributes to built-in types, so `_total_usage` is only available on `BaseModel` and `ListResponse` shapes. The hook and `token_budget` still work regardless of response shape since they operate before finalization.

14 tests total now, all passing.

jxnl:
Thanks @mimran-khan. I rebuilt this against current `main` as draft #2512 and credited this PR in the replacement. The canonical version keeps the product direction here while parsing successful responses before enforcing the budget, stopping at the exact boundary, emitting immutable usage snapshots, defining missing-usage and streaming behavior, preserving public typing/list metadata, and adding deterministic sync/async call-count coverage. I am keeping this PR open until #2512 lands; after that it can be closed as intentionally superseded.

jxnl:
Closing this PR as superseded by #2512. The replacement keeps the cumulative-usage intent while enforcing a positive budget before another provider attempt, preserving valid responses that cross the threshold, exposing immutable usage snapshots, covering sync/async response shapes, and passing the full matrix. Credit is retained in #2512. Issues #2391 and #2056 remain open until the behavior reaches `main`.

Issue #2395, opened 2026-06-27
Fix the bug on ollama.md


badaboot:
- [x] This is actually a bug report.
- [x] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

- [x] llama3.2

**Describe the bug**
On following instructions on ollama.md at https://python.useinstructor.com/examples/ollama/#ollama I get error.

**To Reproduce**
Follow steps in ollama.md on main branch. 
Get error

<img width="1306" height="530" alt="Image" src="https://github.com/user-attachments/assets/ab469d65-0253-42b2-a769-ce8b75b7effd" />

**Expected behavior**
No errors.

jxnl:
Fixed by #2398, which updates the Ollama example to llama3.2 and refreshes stale docs model strings.

PR #2396, opened 2026-06-27
update model name

badaboot:
This is a bug fix. This fixes the ollama.md bug on #2395.
Before: error
Screenshot 2026-06-26 at 6 57 03 PM

After: no error
Screenshot 2026-06-26 at 7 05 02 PM


jxnl:
Consolidated in #2398, which also updates the remaining stale docs model strings.

PR #2409, opened 2026-07-04
feat(bedrock): support Amazon Bedrock API key (bearer token) authentication

### LINKED ISSUE #2408: feat(bedrock): support Amazon Bedrock API key (bearer token) authentication

The Bedrock provider currently offers no way to authenticate with an [Amazon Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) (bearer token). Bedrock API keys let you call the service with a bearer token instead of SigV4-signed AWS credentials, and come in two flavors:

- **Short-term** – lasts up to 12 hours (or the duration of your session, whichever is shorter), inherits permissions from the IAM principal that generated it. Recommended for production.
- **Long-term** – lasts until a configured expiration date, backed by an auto-created IAM user. Recommended only for exploration.

Today, `_build_bedrock` in `instructor/v2/auto_client.py` explicitly ignores the `api_key` argument and only wires up classic credential parameters:

https://github.com/567-labs/instructor/blob/main/instructor/v2/auto_client.py

```python
def _build_bedrock(
    *,
    provider: str,
    model_name: str,
    async_client: bool,
    mode: Mode | None,
    api_key: str | None,  # noqa: ARG001   <-- silently ignored
    kwargs: dict[str, Any],
    ...
) -> InstructorType:
    ...
    for key in [
        "aws_access_key_id",
        "aws_secret_access_key",
        "aws_session_token",
    ]:
        ...
```

So this silently falls back to the default credential chain instead of using the key:

```python
client = instructor.from_provider(
    "bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0",
    api_key="bedrock-api-key-...",  # ignored, no warning
)
```

This hurts ergonomics for people trying out instructor with Bedrock: every other major provider (`openai/...`, `anthropic/...`, etc.) works with a single `api_key`, while Bedrock forces users to set up the full AWS credential chain even though the service itself now supports simple bearer-token auth.

Support Bedrock API keys in the Bedrock provider path:

1. In `_build_bedrock`, when `api_key` is passed (or `AWS_BEARER_TOKEN_BEDROCK` is set in the environment), configure the boto3 client to use bearer-token auth instead of SigV4 credentials. botocore natively supports this since boto3 1.39.0 via the `AWS_BEARER_TOKEN_BEDROCK` environment variable, so a minimal implementation could be:

   ```python
   if api_key:
       os.environ["AWS_BEARER_TOKEN_BEDROCK"] = api_key
   ```

   (or, more hygienically, by injecting the token into the botocore session/auth config rather than mutating `os.environ`).

2. Document the option in the Bedrock integration docs, including the `AWS_BEARER_TOKEN_BEDROCK` environment variable, so users know both paths exist.

3. If bearer auth can't be honored (e.g. boto3 too old), raise or warn instead of silently ignoring `api_key`.

- Setting `AWS_BEARER_TOKEN_BEDROCK` manually before creating the client — this works because botocore picks it up, but it is undocumented in instructor and doesn't help users who reach for the `api_key` parameter that every other provider accepts.
- Building a `boto3` client myself and using `instructor.from_bedrock(client)` — works, but defeats the purpose of the `from_provider` one-liner.
- At minimum, a warning when `api_key` is passed to `from_provider("bedrock/...")` would prevent silent misconfiguration.

**Additional context**

- AWS announcement: https://aws.amazon.com/blogs/machine-learning/accelerate-ai-development-with-amazon-bedrock-api-keys/
- Bedrock API key docs: https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html
- boto3 support for `AWS_BEARER_TOKEN_BEDROCK` landed in 1.39.0 (July 2025).

Happy to submit a PR for this if the approach above sounds reasonable.

Winne004:
## Describe your changes

Adds support for [Amazon Bedrock API keys](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) (bearer tokens) in the Bedrock provider path of `from_provider`.

Previously, `_build_bedrock` in `instructor/v2/auto_client.py` accepted `api_key` but silently ignored it (`# noqa: ARG001`), only wiring up `aws_access_key_id` / `aws_secret_access_key` / `aws_session_token`. Bedrock is the only major provider where `api_key` doesn't work, which hurts ergonomics for people trying instructor with Bedrock:

```python
client = instructor.from_provider(
    "bedrock/anthropic.claude-3-5-sonnet-20241022-v2:0",
    api_key="your-bedrock-api-key",  # was silently ignored — now works
)
```

**Implementation:** when `api_key` is provided, it is exported as `AWS_BEARER_TOKEN_BEDROCK` before the boto3 client is created. This is the mechanism botocore (>= 1.39.0) natively supports for Bedrock bearer-token auth, and it's how other libraries in the ecosystem (e.g. langchain-aws) enable it. When no `api_key` is passed, behavior is unchanged — including picking up an `AWS_BEARER_TOKEN_BEDROCK` the user already exported themselves.

**Changes:**
- `instructor/v2/auto_client.py` — honor `api_key` in `_build_bedrock` by exporting `AWS_BEARER_TOKEN_BEDROCK`
- `tests/v2/test_auto_client_deterministic.py` — two new deterministic tests: `api_key` sets the bearer-token env var; no `api_key` leaves it unset
- `docs/integrations/bedrock.md` — new "Bedrock API Key (Bearer Token) Authentication" section covering both the `api_key` parameter and the env var, with the boto3 version requirement and credential-precedence note

Verified locally: `pytest tests/v2` → 1498 passed, 182 skipped; `ruff check` / `ruff format` clean.

Closes #2408


jxnl:
1.16 triage: keeping this open with #2408, but not merging the current patch. Assigning `AWS_BEARER_TOKEN_BEDROCK` in `os.environ` mutates process-global credential state, can leak one caller's token into concurrent clients, and the tests assert that unsafe side effect. AWS documents the environment mechanism, but `api_key=` support needs a session-scoped SDK path plus live Bedrock verification; users can continue setting the documented environment variable themselves.

PR #2412, opened 2026-07-05
fix: build proper model instances for Optional[BaseModel] fields during partial streaming

shadowmodder:
## Problem

When streaming a `Partial[Model]` and the root JSON is still open (i.e. the
LLM hasn't closed the outer `}`), `_build_partial_object()` is called for
every chunk.  Fields annotated as `Optional[NestedModel]` (equivalently
`Union[NestedModel, None]`) were silently stored as raw `dict` objects instead
of `NestedModel` instances, because the guard used was:

```python
isinstance(field_type, type) and issubclass(field_type, BaseModel)
```

`Optional[NestedModel]` is a generic alias (`typing.Union[NestedModel, None]`),
not a `type`, so `isinstance` returns `False` and the dict falls through
unprocessed.  The symptom is an `AttributeError` when user code accesses
`chunk.inner.some_field` mid-stream.

This asymmetry is clearly a bug: a field typed `inner: Inner` correctly yields
a partial `Inner` instance during incomplete streaming, while `inner:
Optional[Inner]` yields a raw `dict`.

## Fix

Add a helper `_unwrap_optional_base_model(annotation)` that:

- Returns the class directly for a plain `BaseModel` subclass annotation.
- Unwraps `Optional[T]` / `Union[T, None]` (exactly **one** non-`None` arg
  that is a `BaseModel` subclass) and returns `T`.
- Returns `None` for all other annotations, including multi-member `Union`s
  (to avoid ambiguity about which branch to validate against).

Replace the three affected `isinstance(field_type, type)` guards in
`_build_partial_object` (complete-field path and dict-recurse path) and
`_build_partial_list` with this helper.

The **missing-fields** branch in `_build_partial_object` is intentionally left
unchanged: an absent `Optional[Inner]` field should remain `None` (its
default), not become an empty `Inner()` instance.

## Tests

Three new regression tests in `TestOptionalNestedBaseModelDuringPartialStreaming`:

1. `Optional[Inner]` present during an incomplete-root stream yields an `Inner`
   instance, not a `dict`.
2. Attribute access on that instance works without raising `AttributeError`.
3. When the `Optional[Inner]` field is absent altogether, the value remains
   `None` (default-contract preserved).

All 49 unit tests pass (`46` pre-existing + `3` new), `2` skipped
(require `OPENAI_API_KEY`).

### DIFF
--- a/instructor/v2/dsl/partial.py
+++ b/instructor/v2/dsl/partial.py
@@ -44,6 +44,27 @@ UNION_ORIGINS = (Union, UNION_TYPE) if UNION_TYPE is not None else (Union,)
 _processing_models: set[type] = set()
 
 
+def _unwrap_optional_base_model(annotation: Any) -> type[BaseModel] | None:
+    """Return the BaseModel subclass for a type annotation, unwrapping Optional[T].
+
+    Handles:
+    - Direct BaseModel subclass  →  returns it as-is
+    - ``Optional[T]`` / ``Union[T, None]`` where T is a BaseModel  →  returns T
+    - Arbitrary ``Union`` with multiple non-None args  →  returns None (ambiguous)
+    - Any other annotation  →  returns None
+    """
+    if isinstance(annotation, type) and issubclass(annotation, BaseModel):
+        return annotation
+    origin = get_origin(annotation)
+    if origin in UNION_ORIGINS:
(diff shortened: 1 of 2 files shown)


jxnl:
Superseded by #2434, which includes the Optional[BaseModel] partial-streaming fix and its regression coverage. Thank you for the focused diagnosis and implementation.

PR #2415, opened 2026-07-07
fix(openai): derive input_audio format from media type

vineethsaivs:
### Problem

`audio_to_openai` always sets `input_audio.format` to `"wav"`, ignoring the audio's `media_type`:

```python
def audio_to_openai(audio: Any, mode: Mode) -> dict[str, Any]:
    if mode in RESPONSES_MODES:
        raise ValueError("OpenAI Responses doesn't support audio")
    return {"type": "input_audio", "input_audio": {"data": audio.data, "format": "wav"}}
```

OpenAI Chat Completions `input_audio.format` only accepts `"wav"` or `"mp3"`. `Audio` accepts many audio MIME types (`VALID_AUDIO_MIME_TYPES`), and `Audio.from_path("clip.mp3")` stores `media_type="audio/mpeg"` (`mimetypes.guess_type` maps `.mp3` to `audio/mpeg`). So an mp3 gets sent labeled `"wav"` and is rejected or mis-decoded. The sibling `image_to_openai` and `pdf_to_openai` already read `media_type`; only the audio handler ignored it.

### Fix

Derive the format from `audio.media_type`: map the mp3 media types (`audio/mp3`, `audio/mpeg`, `audio/mpga`) to `"mp3"`, and default to `"wav"` otherwise (the two formats OpenAI Chat Completions supports; `"wav"` is the safe default and is no worse than today for the rarer types). WAV audio is unchanged.

### Test

`test_audio_to_openai_format_follows_media_type` asserts an `audio/mpeg` input yields `format == "mp3"` and an `audio/wav` input yields `format == "wav"`. It fails on the current code (mp3 comes back as `"wav"`) and passes with this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 - **Templating**: Use populated `contents` when `messages` is empty, avoid mutating nested caller input, and preserve uncopyable metadata during template expansion.
 - **Anthropic system messages**: Reject invalid new system-message values even when no existing system message is present.
 - **Python 3.9**: Include the required type-evaluation backport in minimal installs, keep overload metadata available, and avoid runtime evaluation of unsupported union syntax in the core response path and offline tests.
+- **OpenAI**: Send the correct `input_audio.format` for audio content by deriving it from the audio's media type (e.g. mp3) instead of always sending `"wav"`, so non-WAV audio is no longer mislabeled. ([#2415](https://github.com/567-labs/instructor/pull/2415))
 
 ### Tests / CI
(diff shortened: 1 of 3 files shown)


jxnl:

vineethsaivs:
Thanks, and no objection to the consolidation. Confirmed the fix landed intact in #2495: `audio_to_openai` now derives the format from `media_type` with the three mp3 aliases mapped, and `test_audio_to_openai_format_follows_media_type` came across as well.

Your version is stronger than mine on the fallback. I defaulted anything unrecognised to `"wav"`, which quietly mislabels a format OpenAI does not accept; raising instead surfaces it at the call site:

```python
if media_type not in format_by_media_type:
    raise ...  # Unsupported OpenAI audio format: ...
```

That is the better contract, since the whole bug was audio being sent under the wrong label. Worth keeping in mind for the sibling encoders if they ever grow the same table.

PR #2418, opened 2026-07-08
fix(v2): stop OpenAI-compat handlers from mutating caller's messages list

### LINKED ISSUE #2417: `client.create()` mutates the caller's `messages` list (and sometimes message dicts) in place — affects far more than OpenAI TOOLS/JSON_SCHEMA/RESPONSES_TOOLS
## Environment
- instructor `1.15.5` (commit `cd698be4e79b1b3aa64da31d9e551c4e8f60be11`, 2026-06-28)
- Python 3.x, `openai` SDK response types used in repro (no network call needed to reproduce)

## Summary
When a Pydantic validation error triggers instructor's reask/retry path — and in a couple of cases even on a single successful call with no retry — the `messages` list passed in by the caller can be mutated **in place**, i.e. the same list object (or the same message dicts inside it), not a copy.

This is surprising because most affected handlers do `kwargs = kwargs.copy()` (or `new_kwargs = kwargs.copy()`) before mutating, which looks like a defensive copy but is a **shallow** copy — the `messages` *value* inside the dict is still the same list reference, and the dicts inside that list are still the same dict references. `dict.copy()` protects neither.

Concretely: if application code holds a reference to its own message list (e.g. conversation state in a LangGraph node, a chat history object, a list reused across calls), calling `client.create(messages=my_messages, response_model=X)` can silently graft retry artifacts onto `my_messages` after the call returns — even though the call succeeded and even though nothing in the public API signature suggests `messages` is anything but an input.

I originally found this in OpenAI's TOOLS/JSON_SCHEMA/RESPONSES_TOOLS modes, but tracing the same root cause (aliasing, not copying, `messages`) across the rest of the codebase turned up a substantially wider blast radius. I'm splitting it into three categories below because they have different root causes and will need different fixes.

## Category 1 — shared handler classes (one code location, many providers, fixed together)

`instructor/v2/providers/openai/handlers.py` defines a handler class once and registers it for a *list* of providers via `@register_mode_handler(SOME_PROVIDER_LIST, Mode.X)`. None of these rebind `messages` to a new list, so every provider in the list shares the exact same bug from the exact same code:

- `OpenAIToolsHandler` (`Mode.TOOLS`, line 594-595) → registered for `OPENAI_COMPAT_PROVIDERS` = OpenAI, Anyscale, Together, Databricks, Deepseek, OpenRouter, Groq, Fireworks, Cerebras (9 providers)
- `OpenAIJSONSchemaHandler` (`Mode.JSON_SCHEMA`, line 714-715) → registered for `OPENAI_JSON_SCHEMA_PROVIDERS` = OpenAI, Anyscale, Together, Databricks, Deepseek, Groq, Fireworks, Cerebras (8 providers)
- `OpenAIParallelToolsHandler` (`Mode.PARALLEL_TOOLS`, line 982-983) → registered for `OPENAI_PARALLEL_TOOL_PROVIDERS` = OpenAI, Anyscale, Together, Databricks, Deepseek, OpenRouter, Cerebras (7 providers) — its `handle_reask` delegates to the same `reask_tools()` used by `OpenAIToolsHandler`
- `OpenAIResponsesToolsHandler` (`Mode.RESPONSES_TOOLS`, line 1062-1063) → OpenAI only

I verified with a script that iterates every registered `(Provider, Mode)` pair and checks `new_kwargs["messages"] is caller_messages` right after `prepare_request` — all of the above come back `True` (aliased). By contrast `OpenAIJSONHandler` (`Mode.JSON`) and `OpenAIMDJSONHandler` (`Mode.MD_JSON`), also registered for the same 9-provider list, come back `False` — they already rebuild `messages` into a new list, which is why they're unaffected.

**One fix in these three/four OpenAI-side classes remediates all providers listed above.**

## Category 2 — independently duplicated in provider-specific code (separate fixes needed)

These providers have their own `handlers.py` with their own `prepare_request`/`handle_reask`, and independently reproduce the same "shallow `kwargs.copy()`, no `messages` rebind" mistake:

- **Mistral** — `MistralToolsHandler.prepare_request` (`instructor/v2/providers/mistral/handlers.py:239`): `new_kwargs = kwargs.copy()`, never rebinds `messages`.
- **OpenRouter** — `OpenRouterJSONSchemaHandler.prepare_request` (`instructor/v2/providers/openrouter/handlers.py:23`): subclasses `OpenAIJSONSchemaHandler` but overrides `prepare_request` with `new_kwargs = kwargs.copy()`, reintroducing the exact same bug even though its parent class would otherwise be fixed by the Category 1 change.
- **Cohere** — `CohereJSONSchemaHandler.handle_reask` (`instructor/v2/providers/cohere/handlers.py:303` area): does `kwargs["messages"].append(...)` directly on whatever `prepare_request` handed it, and `prepare_request`'s `_convert_messages(kwargs)` does not copy the list either.
- **xAI** — `XAIToolsHandler.handle_reask` (`instructor/v2/providers/xai/handlers.py:385` area): `kwargs["messages"].append(assistant_msg)` on the aliased list from `prepare_request`'s bare `kwargs.copy()`.
- **Writer** — `WriterToolsHandler` (`instructor/v2/providers/writer/handlers.py:120`): same pattern — `prepare_request` never rebinds `messages`, and `reask_writer_tools` appends onto the shared list.

Notably, `CohereToolsHandler.prepare_request` (`Mode.TOOLS`) gets this *right* — it has an explicit comment "copy to avoid mutating caller's list" and does `new_kwargs["messages"] = list(new_kwargs["messages"])` before inserting. So the fix pattern is already present elsewhere in the same file; it just wasn't applied consistently to the JSON_SCHEMA/MD_JSON handlers in the same provider.

## Category 3 — worse: mutates on the very first call, before any retry

These two don't even need a validation failure to corrupt the caller's state — they mutate a message **dict** in place inside `prepare_request` itself, so a single successful call with zero retries still corrupts `my_messages`:

- **Mistral** — `MistralMDJSONHandler.prepare_request` (`instructor/v2/providers/mistral/handlers.py:471` area):
  ```python
  messages = new_kwargs.get("messages", [])
  if messages and messages[0]["role"] != "system":
      messages.insert(0, {...})
  elif messages and isinstance(messages[0]["content"], str):
      messages[0]["content"] += f"\n\n{message}"
  ```
  Both the `.insert(0, ...)` and the `+=` mutate the caller's list/dict directly — `new_kwargs = kwargs.copy()` never rebinds `messages`.

- **Cohere** — `CohereMDJSONHandler.prepare_request` (`instructor/v2/providers/cohere/handlers.py:381` area):
  ```python
  if new_kwargs["messages"]:
      last_msg = new_kwargs["messages"][-1]
      last_msg["content"] = f"{last_msg.get('content', '')}\n\n{instruction}"
  ```
  `last_msg` is the caller's own dict (from the aliased list); this line rewrites its `content` field in place.

These need a slightly different fix than "rebind the list" — the message *dicts themselves* need to be copied before mutation, not just the list.

## Unaffected (verified, no action needed)
- Anthropic (`AnthropicToolsHandler` and all other Anthropic modes) — `prepare_request` rebinds `messages` via a list comprehension before the retry loop runs.
- Bedrock, Gemini — all modes came back `aliased=False` in my scan.
- `Mode.JSON` / `Mode.MD_JSON` on the OpenAI-compatible providers (as noted above).

## Reproduction
Attached: `repro_final.py`. Builds real `openai.types.chat.ChatCompletion` objects (no network call), calls the actual `mode_registry` handlers and `retry_sync_v2` exactly as `patch.py`'s `new_create_sync` does, and forces one validation failure (missing `age` field) before a valid second attempt.

```
len(my_messages) BEFORE the call: 1
len(my_messages) AFTER  the call: 3
```

`my_messages` (the exact list object passed to `messages=`) ends up containing the failed assistant tool call and the synthetic validation-error tool message — despite `client.create()` having returned a valid, successfully-validated model.

## Expected behavior
`client.create()` should treat `messages=` as a read-only input, the same as any typical API wrapper. The caller's list (and the message dicts inside it) should be unchanged after the call returns, regardless of how many retries happened internally.

## Suggested fix
For Category 1/2 (list aliasing), rebind rather than alias at the top of each affected `prepare_request`:
```python
new_kwargs = kwargs.copy()
new_kwargs["messages"] = list(kwargs.get("messages", []))  # shallow copy of the list itself
```
For Category 3 (dict mutation), the specific message dict being edited needs its own copy before the in-place field assignment, e.g. `last_msg = dict(new_kwargs["messages"][-1])` before mutating `last_msg["content"]`.

Given the scope, I'd suggest landing this as more than one PR rather than one large diff:
1. The Category 1 fix (OpenAI base handlers) is small, high-confidence, and fixes 9 providers at once via the shared classes — good first PR.
2. The Category 2/3 fixes touch four independent provider files and would benefit from being reviewed (and tested) separately.

Happy to open the first PR (Category 1) now, and follow up with the others — or to do it differently if maintainers have a preferred granularity for this kind of cross-cutting fix. A regression test that iterates every registered `(provider, mode)` pair and asserts `new_kwargs["messages"] is not caller_messages` (plus, for Category 3, that individual message dicts aren't shared) would catch all of these and prevent regressions going forward.

---
*Repro script attached below.*

mikko-lab:
## Summary
- Fixes #2417: `client.create()` was mutating the caller's own `messages` list (and, in some cases, the message dicts inside it) in place, on the OpenAI-compat handlers.
- Two related root causes, both from the same `kwargs.copy()`-is-shallow mistake:
  1. **Reask-time list aliasing** (`Mode.TOOLS`, `Mode.JSON_SCHEMA`, `Mode.PARALLEL_TOOLS`, `Mode.RESPONSES_TOOLS`): `prepare_request` never rebound `new_kwargs["messages"]` to a new list, so `handle_reask` appended retry artifacts directly onto the caller's original list object.
  2. **First-call dict/list mutation** (`Mode.JSON`, `Mode.MD_JSON`): `prepare_request` injects a system/instruction message by mutating the `messages` list and its dict elements directly (`.insert()`, `.append()`, `message["content"] += ...`) *before* rebuilding into a fresh list via `merge_consecutive_messages()`. This one is worse — it corrupts the caller's state on every call, not just ones that hit a validation reask. (I initially and incorrectly marked these two modes "unaffected" in the issue; see the correction there and in this PR's comments.)
- Fix:
  - For (1): rebind `messages` to a fresh list right after the shallow copy, in `OpenAIToolsHandler`, `OpenAIJSONSchemaHandler`, `OpenAIParallelToolsHandler`, and `OpenAIResponsesToolsHandler`.
  - For (2): added `copy_messages_for_mutation()` (`instructor/v2/core/messages.py`), which copies the list and any dict elements (plus their `content` list items) before the in-place edits run. Used by `OpenAIJSONHandler` and `OpenAIMDJSONHandler`.
- Because all six of these handler classes are registered for multiple providers at once (`OPENAI_COMPAT_PROVIDERS` / `OPENAI_JSON_SCHEMA_PROVIDERS` / `OPENAI_PARALLEL_TOOL_PROVIDERS`), this fixes Anyscale, Together, Databricks, Deepseek, OpenRouter, Groq, Fireworks, and Cerebras too — 9 providers total from one focused diff.

## Scope note
The same anti-pattern is independently reproduced in provider-specific handlers that don't share these classes: Mistral (`Mode.TOOLS`, plus a first-call `Mode.MD_JSON` dict-mutation case matching (2) above), Cohere (`Mode.JSON_SCHEMA`, plus a first-call `Mode.MD_JSON` case), Writer (`Mode.TOOLS`), xAI (`Mode.TOOLS`), and OpenRouter's own `Mode.JSON_SCHEMA` override. Documented in #2417, intentionally left out of this PR to keep the diff reviewable — happy to follow up, or fold them in here, whichever's preferred.

## Test plan
- [x] `tests/v2/test_messages_not_mutated.py`:
  - Parametrized test over every `(provider, mode)` pair fixed by this PR (now including `Mode.JSON`/`Mode.MD_JSON`), snapshotting the caller's `messages` with `deepcopy` before the call and asserting it's byte-for-byte unchanged after — an identity-only check on the returned list is not sufficient for case (2) above, since the input can be mutated before being rebuilt into a new list.
  - An end-to-end test replaying a real validation-failure-then-success cycle through `retry_sync_v2`, asserting the caller's original list is unchanged afterward.
  - `xfail(strict=True)` cases documenting the still-broken Mistral/Cohere/Writer/xAI/OpenRouter pairs.
- [x] `pytest tests/v2/` — 1549 passed, 179 skipped (missing optional provider SDKs), 5 xfailed, 0 failed.
- [x] `pytest tests/` (excluding `tests/v2` and `tests/llm/test_vertexai`) — same 85 pre-existing failures as unpatched `main` (verified via `git stash`; all need live API credentials or are docs-snippet tests, unrelated to this change).

### DIFF
--- a/instructor/v2/core/messages.py
+++ b/instructor/v2/core/messages.py
@@ -19,6 +19,25 @@ def extract_messages(kwargs: dict[str, Any]) -> Any:
     return []
 
 
+def copy_messages_for_mutation(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Return a copy of `messages` that is safe to mutate in place.
+
+    Handlers that inject a system/instruction message do so by mutating a
+    `messages` list and its dict elements directly (`.insert()`, `.append()`,
+    `message["content"] += ...`). Since `kwargs.copy()` is shallow, `messages`
+    is otherwise the exact list (and dict) objects the caller passed in, so
+    those in-place edits would corrupt the caller's own conversation state.
+    """
+    copied = [dict(message) for message in messages]
+    for message in copied:
+        content = message.get("content")
+        if isinstance(content, list):
+            message["content"] = [
+                dict(item) if isinstance(item, dict) else item for item in content
+            ]
+    return copied
+
+
 def dump_message(message: ChatCompletionMessage) -> ChatCompletionMessageParam:
     ret: ChatCompletionMessageParam = {
         "role": message.role,
--- a/instructor/v2/providers/openai/handlers.py
+++ b/instructor/v2/providers/openai/handlers.py
@@ -607,6 +607,7 @@ class OpenAIToolsHandler(OpenAIHandlerBase):
     ) -> tuple[Any, dict[str, Any]]:
         """Prepare request with tool definitions."""
         new_kwargs = kwargs.copy()
+        new_kwargs["messages"] = list(kwargs.get("messages", []))
 
         if response_model is None:
             return None, new_kwargs
@@ -732,6 +733,7 @@ class OpenAIJSONSchemaHandler(OpenAIHandlerBase):
             return None, kwargs
 
         new_kwargs = kwargs.copy()
+        new_kwargs["messages"] = list(kwargs.get("messages", []))
         schema = response_model.model_json_schema()
         new_kwargs["response_format"] = {
             "type": "json_schema",
@@ -995,6 +997,7 @@ class OpenAIParallelToolsHandler(OpenAIHandlerBase):
             return None, kwargs
 
         new_kwargs = kwargs.copy()
+        new_kwargs["messages"] = list(kwargs.get("messages", []))
         if new_kwargs.get("stream", False):
             raise ConfigurationError(
                 "stream=True is not supported when using PARALLEL_TOOLS mode"
@@ -1074,6 +1077,7 @@ class OpenAIResponsesToolsHandler(OpenAIHandlerBase):
         self._register_streaming_from_kwargs(response_model, kwargs)
 
         new_kwargs = kwargs.copy()
+        new_kwargs["messages"] = list(kwargs.get("messages", []))
 
         # Handle max_tokens to max_output_tokens conversion
         if new_kwargs.get("max_tokens") is not None:
--- a/instructor/v2/providers/openai/handlers.py
+++ b/instructor/v2/providers/openai/handlers.py
@@ -41,7 +41,11 @@ from instructor.v2.core.json import (
     extract_json_from_stream,
     extract_json_from_stream_async,
 )
-from instructor.v2.core.messages import dump_message, merge_consecutive_messages
+from instructor.v2.core.messages import (
+    copy_messages_for_mutation,
+    dump_message,
+    merge_consecutive_messages,
+)
 from instructor.v2.providers.openai.schema import generate_openai_schema
 from instructor.v2.core.decorators import register_mode_handler
 from instructor.v2.core.handler import ModeHandler
@@ -819,7 +823,7 @@ class OpenAIJSONHandler(OpenAIHandlerBase):
             Make sure to return an instance of the JSON, not the schema itself
             """
         )
-        messages = new_kwargs.get("messages", [])
+        messages = copy_messages_for_mutation(new_kwargs.get("messages", []))
         if messages and messages[0]["role"] != "system":
             messages.insert(
                 0,
@@ -909,7 +913,7 @@ class OpenAIMDJSONHandler(OpenAIHandlerBase):
         )
 
         # Add system message with schema
-        messages = new_kwargs.get("messages", [])
+        messages = copy_messages_for_mutation(new_kwargs.get("messages", []))
         if messages and messages[0]["role"] != "system":
             messages.insert(
                 0,
--- a/tests/v2/test_messages_not_mutated.py
+++ b/tests/v2/test_messages_not_mutated.py
@@ -1,9 +1,20 @@
 """Regression tests for the messages= aliasing bug (see GitHub issue #2417).
 
 `client.create()` must treat `messages=` as a read-only input. Internally,
-`prepare_request` must never hand back the caller's own list object (or leave
-the caller's message dicts open to in-place mutation during a reask), because
-`handle_reask` appends retry artifacts onto whatever list it is given.
+`prepare_request` must never hand back the caller's own list object, or
+leave the caller's message dicts open to in-place mutation, because both
+`prepare_request` itself (JSON/MD_JSON's system-message injection) and
+`handle_reask` (the retry path) mutate whatever `messages` list/dicts they
+are handed.
+
+Note on the aliasing check: comparing `new_kwargs["messages"] is not
+caller_messages` alone is not sufficient -- `OpenAIJSONHandler` and
+`OpenAIMDJSONHandler` mutate the *input* list/dicts in place before
+rebuilding `messages` into a genuinely new list via
+`merge_consecutive_messages()`, so an identity check on the final result
+would pass even though the caller's original objects were already
+corrupted. These tests instead snapshot `caller_messages` with `deepcopy`
+before the call and assert it is unchanged afterward.
 
 The provider lists below are imported directly from the shared OpenAI-compat
 handler module so this test stays in sync with the source registrations
@@ -12,6 +23,7 @@ rather than duplicating them.
 
 from __future__ import annotations
 
+from copy import deepcopy
 from typing import Any, Iterable
 
 import pytest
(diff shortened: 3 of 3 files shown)


mikko-lab:
Pushed an additional commit to this PR (before any review landed): `OpenAIJSONHandler` (`Mode.JSON`) and `OpenAIMDJSONHandler` (`Mode.MD_JSON`) turned out to have the same class of bug, and worse — they mutate the caller's `messages` list/dicts in place inside `prepare_request` itself (system-message injection via `.insert()`/`.append()`/`content +=`), before rebuilding into a new list via `merge_consecutive_messages()`. That means it corrupts caller state on *every* call, not just reask/retry ones. Details and the "I got this wrong the first time" writeup are in #2417.

Updated in this PR:
- New `copy_messages_for_mutation()` helper in `instructor/v2/core/messages.py`, used by both handlers to copy the list and its dict elements before any in-place edit.
- `tests/v2/test_messages_not_mutated.py` now also covers `Mode.JSON`/`Mode.MD_JSON` across all `OPENAI_COMPAT_PROVIDERS`, and the aliasing check was changed from an identity comparison to a deepcopy snapshot comparison — identity-only was exactly what let the JSON/MD_JSON case slip past the first pass.
- Full `tests/v2/` suite: 1549 passed, 179 skipped (missing optional provider SDKs), 5 xfailed (documented, still-open Category 2/3 cases), 0 failed.

PR title/description still describe this as the OpenAI-compat "shared handler" fix; happy to update the title if you'd like it to explicitly call out JSON/MD_JSON too.

mikko-lab:
Follow-up opened: #2419 fixes the same anti-pattern in Mistral/Cohere/Writer/xAI/OpenRouter's own handler code (the providers not covered by the shared classes here). It's stacked on this branch since it reuses `copy_messages_for_mutation()`.

jxnl:
Superseded by #2434, which consolidates this OpenAI-compatible message-aliasing fix with the provider follow-up and preserves contributor attribution. Thank you for the thorough test coverage.

PR #2419, opened 2026-07-08
fix(v2): stop Mistral/Cohere/Writer/xAI/OpenRouter from mutating caller's messages

mikko-lab:
## Stacked on #2418

This branch is built on top of #2418 (`fix/messages-aliasing-openai-handlers`) so it can reuse `copy_messages_for_mutation()` introduced there. Until #2418 merges, this diff will show both PRs' commits; GitHub will automatically narrow it down to just the new commit once #2418 merges into `main`. The only new commit here is `7ca74b0e`.

## Summary
Follow-up to #2417 / #2418: the same "shallow `kwargs.copy()`, `messages` never rebound/copied" bug, independently reproduced in five providers' own handler code (not the shared OpenAI-compat classes fixed in #2418):

**Reask-time list aliasing** (`prepare_request` never rebinds `messages`, so `handle_reask` appends retry artifacts onto the caller's original list):
- Mistral: `Mode.TOOLS`, `Mode.JSON_SCHEMA`
- Cohere: `Mode.JSON_SCHEMA` — fixed centrally in `_convert_messages_to_cohere_v2`, which also covers `Mode.TOOLS` (already had its own explicit fix) and `Mode.MD_JSON` since all three funnel through the same conversion function
- Writer: `Mode.TOOLS`, `Mode.JSON_SCHEMA`
- xAI: `Mode.TOOLS`, `Mode.PARALLEL_TOOLS`, `Mode.JSON_SCHEMA`
- OpenRouter: `Mode.JSON_SCHEMA` (its own `prepare_request` override)

**First-call dict/list mutation** (worse — corrupts caller state on every call, not just reasks; same category as OpenAI's `Mode.JSON`/`Mode.MD_JSON` fixed in #2418):
- Mistral: `Mode.MD_JSON`
- Cohere: `Mode.MD_JSON` (covered by the same central conversion fix above)
- Writer: `Mode.MD_JSON`
- xAI: `Mode.MD_JSON`

13 `(provider, mode)` pairs fixed in total — these were tracked as `xfail(strict=True)` known-broken cases in #2418's test file; this PR fixes them and removes the `xfail` list.

## Verification
Before writing the fix, empirically confirmed each of the 13 pairs actually mutates the caller's original `messages` on unpatched code:
- For the first-call category: called `prepare_request` alone and confirmed the caller's list/dict changed.
- For the reask-time category: `prepare_request` alone does *not* mutate anything for these modes (that's the bug — it silently hands back an aliased reference), so verifying required simulating the real retry path: `prepare_request` followed by `handle_reask` with a realistic mock response + validation error, confirming the caller's list gained the reask's appended messages.

After the fix, re-ran the exact same 13 checks and confirmed the caller's `messages` is untouched in every case, using both an identity check (`new_kwargs["messages"] is not caller_messages`) and a `deepcopy` snapshot content comparison (identity alone doesn't catch the first-call dict-mutation cases, per the correction in #2417).

## Test plan
- [x] `tests/v2/test_messages_not_mutated.py`: all 13 pairs folded into `FIXED_PAIRS`; new `test_reask_does_not_mutate_caller_messages` drives the full `prepare_request` → `handle_reask` cycle for the 9 reask-time pairs (a prepare_request-only check would miss this class of bug, since the mutation only happens once a reask actually runs).
- [x] `pytest tests/v2/` — 1571 passed, 179 skipped (missing optional provider SDKs), 0 failed, 0 xfailed.
- [x] `pytest tests/` (excluding `tests/v2` and `tests/llm/test_vertexai`) — same 85 pre-existing failures as unpatched `main` (verified via `git stash`; all need live API credentials or are docs-snippet tests, unrelated to this change).

### DIFF
--- a/instructor/v2/core/messages.py
+++ b/instructor/v2/core/messages.py
@@ -19,6 +19,25 @@ def extract_messages(kwargs: dict[str, Any]) -> Any:
     return []
 
 
+def copy_messages_for_mutation(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Return a copy of `messages` that is safe to mutate in place.
+
+    Handlers that inject a system/instruction message do so by mutating a
+    `messages` list and its dict elements directly (`.insert()`, `.append()`,
+    `message["content"] += ...`). Since `kwargs.copy()` is shallow, `messages`
+    is otherwise the exact list (and dict) objects the caller passed in, so
+    those in-place edits would corrupt the caller's own conversation state.
+    """
+    copied = [dict(message) for message in messages]
+    for message in copied:
+        content = message.get("content")
+        if isinstance(content, list):
+            message["content"] = [
+                dict(item) if isinstance(item, dict) else item for item in content
+            ]
+    return copied
+
+
 def dump_message(message: ChatCompletionMessage) -> ChatCompletionMessageParam:
     ret: ChatCompletionMessageParam = {
         "role": message.role,
--- a/instructor/v2/providers/cohere/handlers.py
+++ b/instructor/v2/providers/cohere/handlers.py
@@ -21,6 +21,7 @@ from instructor.v2.core.errors import ConfigurationError, ResponseParsingError
 from instructor.v2.core.json import extract_json_from_codeblock
 from instructor.v2.core.decorators import register_mode_handler
 from instructor.v2.core.handler import ModeHandler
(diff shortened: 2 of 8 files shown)


jxnl:
Superseded by #2434, which consolidates this provider-wide message-mutation fix with its stack parent and preserves contributor attribution. Thank you for tracing the complete provider surface.

PR #2420, opened 2026-07-08
fix(json stream): emit all objects in a chunk, not just the first

Diwak4r:
## Summary
- `extract_json_from_stream` / `extract_json_from_stream_async` built a JSON char buffer and `break` the inner per-char loop once a complete object's closing delimiter was hit. The `break` abandoned the rest of the chunk, so any JSON object following in the same chunk was silently dropped (e.g. `{"a":1}{"b":2}` yielded only `{"a":1}`).
- Changed `break` to `continue` so remaining chars are still scanned, matching the plain object-start path.

## Test plan
- Added `test_extract_json_from_stream_yields_all_objects_in_one_chunk` and `..._in_fenced_chunk` to `tests/v2/test_json_helpers.py`.
- Fail-before/pass-after verified locally (18 passed); ruff clean.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/instructor/v2/core/json.py
+++ b/instructor/v2/core/json.py
@@ -117,7 +117,7 @@ def extract_json_from_stream(chunks: Iterable[str]) -> Generator[str, None, None
                             yield from buffer
                             buffer = []
                             json_started = False
-                            break
+                            continue
 
                 buffer.append(char)
                 continue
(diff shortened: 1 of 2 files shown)


jxnl:
Superseded by #2434, which includes the same-chunk JSON streaming fix plus async regression coverage. Thank you for the concise fix.

PR #2421, opened 2026-07-09
fix(openai): stop mutating caller's messages list in place (issue #2417)

### LINKED ISSUE #2417: `client.create()` mutates the caller's `messages` list (and sometimes message dicts) in place — affects far more than OpenAI TOOLS/JSON_SCHEMA/RESPONSES_TOOLS

gautamkishore:
## Summary

Fixes #2417 (Category 1). The shared OpenAI v2 handlers only performed a shallow `kwargs.copy()` in `prepare_request`, so the caller's `messages` list (and the message dicts inside it) stayed aliased. When a validation failure triggered the reask/retry path, synthetic messages were appended **onto the caller's own list** — corrupting conversation state even though the call eventually succeeded and `messages=` looks like a read-only input.

This PR copies the `messages` list in `prepare_request` for the four affected OpenAI modes:
- `Mode.TOOLS` (shared by OpenAI + 8 OpenAI-compatible providers)
- `Mode.JSON_SCHEMA` (shared by OpenAI + 7 providers)
- `Mode.PARALLEL_TOOLS` (shared by OpenAI + 6 providers)
- `Mode.RESPONSES_TOOLS` (OpenAI)

Because `retry_sync_v2` / `retry_async_v2` receive the prepared `new_kwargs` (which now carry the copied list) and the reask handlers append onto that copy, the caller's list is never mutated. As noted in the issue, the JSON/MD_JSON OpenAI handlers already rebuild the list and are unaffected.

## Why this is safe

- Shallow copy of the list is exactly what the issue author recommends and matches the pattern already used correctly inside `CohereToolsHandler`.
- Reask handlers only append **new** message dicts; they never mutate existing message dicts in these four modes, so a list-level copy fully closes the gap.
- The early-return paths (`response_model is None`) do not mutate and are unchanged.

## Test plan

Added `tests/v2/test_issue_2417.py`:
- `test_openai_prepare_request_does_not_alias_caller_messages` — for each of the 4 OpenAI modes, asserts the prepared request uses a different list object and that the caller's list is byte-for-byte unchanged.
- `test_compat_providers_reuse_fixed_openai_tools_handler` — confirms every OpenAI-compatible provider sharing the `TOOLS` handler gets the fix.
- `test_reask_does_not_mutate_caller_messages` — full end-to-end reproduction through `retry_sync_v2` + the real handlers: a validation failure forces a reask, and the caller's `messages` list remains `[{"role": "user", "content": "What is 2+2?"}]`.

All 13 new tests pass; existing `test_openai_compat_handlers`, `test_retry_runtime`, `test_registry`, and `test_handler_registration_unified` suites still pass.

## CHANGELOG

Added an `[Unreleased] > Fixed` entry referencing #2417.

📌 Note: as suggested in the issue, I've scoped this to Category 1 (the shared OpenAI handlers). The Category 2/3 fixes for Mistral/OpenRouter/Cohere/xAI/Writer can follow in separate PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **v2 OpenAI handlers**: Stop mutating the caller's `messages` list in place. `prepare_request` for `TOOLS`, `JSON_SCHEMA`, `PARALLEL_TOOLS`, and `RESPONSES_TOOLS` now copies the `messages` list, so validation retries (reask) no longer graft synthetic messages onto the caller's own conversation state. ([#2417](https://github.com/567-labs/instructor/issues/2417))
+
+---
+
 ## [1.15.5] - 2026-06-28
 
 ### Fixed
--- a/instructor/v2/providers/openai/handlers.py
+++ b/instructor/v2/providers/openai/handlers.py
@@ -607,6 +607,9 @@ class OpenAIToolsHandler(OpenAIHandlerBase):
     ) -> tuple[Any, dict[str, Any]]:
         """Prepare request with tool definitions."""
         new_kwargs = kwargs.copy()
+        # Copy the caller's messages list so validation retries (reask) cannot
+        # mutate the caller's own list in place. See issue #2417.
+        new_kwargs["messages"] = list(new_kwargs.get("messages", []))
 
         if response_model is None:
             return None, new_kwargs
@@ -732,6 +735,10 @@ class OpenAIJSONSchemaHandler(OpenAIHandlerBase):
             return None, kwargs
 
         new_kwargs = kwargs.copy()
+        # Copy the caller's messages list so validation retries (reask) cannot
+        # mutate the caller's own list in place. See issue #2417.
+        new_kwargs["messages"] = list(new_kwargs.get("messages", []))
+
         schema = response_model.model_json_schema()
         new_kwargs["response_format"] = {
             "type": "json_schema",
@@ -995,6 +1002,10 @@ class OpenAIParallelToolsHandler(OpenAIHandlerBase):
             return None, kwargs
 
         new_kwargs = kwargs.copy()
+        # Copy the caller's messages list so validation retries (reask) cannot
+        # mutate the caller's own list in place. See issue #2417.
+        new_kwargs["messages"] = list(new_kwargs.get("messages", []))
+
         if new_kwargs.get("stream", False):
             raise ConfigurationError(
                 "stream=True is not supported when using PARALLEL_TOOLS mode"
@@ -1074,6 +1085,9 @@ class OpenAIResponsesToolsHandler(OpenAIHandlerBase):
         self._register_streaming_from_kwargs(response_model, kwargs)
 
         new_kwargs = kwargs.copy()
+        # Copy the caller's messages list so validation retries (reask) cannot
+        # mutate the caller's own list in place. See issue #2417.
+        new_kwargs["messages"] = list(new_kwargs.get("messages", []))
 
         # Handle max_tokens to max_output_tokens conversion
         if new_kwargs.get("max_tokens") is not None:
--- a/tests/v2/test_issue_2417.py
+++ b/tests/v2/test_issue_2417.py
@@ -1,0 +1,166 @@
+"""Regression tests for issue #2417.
+
+``client.create()`` (and the underlying v2 handlers) must treat the caller's
+``messages`` list as a read-only input. Before the fix, the shared OpenAI
+handlers only performed a shallow ``kwargs.copy()`` in ``prepare_request``, so
+the ``messages`` value was still the caller's own list. When a validation
+failure triggered the reask/retry path, synthetic messages were appended to
+that list in place -- corrupting the caller's conversation state even on a
+successful call.
+
+See https://github.com/567-labs/instructor/issues/2417 (Category 1).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+from types import SimpleNamespace
+from typing import Any
+
+import pytest
+from openai.types.chat import (
+    ChatCompletion,
+    ChatCompletionMessage,
+    ChatCompletionMessageToolCall,
+)
+from openai.types.chat.chat_completion_message_tool_call import Function
+from pydantic import BaseModel, ValidationError
+from tenacity import Retrying, retry_if_exception_type, stop_after_attempt
+
+from instructor import Mode, Provider
+from instructor.v2.core.retry import retry_sync_v2
+from instructor.v2.core.registry import mode_registry
+
+OPENAI_COMPAT_PROVIDERS = (
+    Provider.ANYSCALE,
+    Provider.TOGETHER,
+    Provider.DATABRICKS,
+    Provider.DEEPSEEK,
+    Provider.OPENROUTER,
+    Provider.GROQ,
+    Provider.FIREWORKS,
+    Provider.CEREBRAS,
+)
+
+
+class Answer(BaseModel):
+    answer: float
+
+
+def _tool_call_response(arguments: str) -> ChatCompletion:
+    message = ChatCompletionMessage(
+        role="assistant",
+        content=None,
+        tool_calls=[
+            ChatCompletionMessageToolCall(
+                id="call_0",
+                type="function",
+                function=Function(name="Answer", arguments=arguments),
+            )
+        ],
+    )
+    choice = {"index": 0, "message": message, "finish_reason": "stop"}
+    return ChatCompletion(
+        id="cmpl-issue-2417",
+        choices=[choice],  # type: ignore[arg-type]
+        created=0,
+        model="gpt-4o-mini",
+        object="chat.completion",
+    )
+
+
+@pytest.mark.parametrize(
+    "mode",
+    [Mode.TOOLS, Mode.JSON_SCHEMA, Mode.PARALLEL_TOOLS, Mode.RESPONSES_TOOLS],
+)
+def test_openai_prepare_request_does_not_alias_caller_messages(mode: Mode) -> None:
+    caller_messages = [{"role": "user", "content": "What is 2+2?"}]
+    original = [dict(m) for m in caller_messages]
+
+    # PARALLEL_TOOLS requires an Iterable-wrapped response model.
+    response_model: Any = Iterable[Answer] if mode is Mode.PARALLEL_TOOLS else Answer
+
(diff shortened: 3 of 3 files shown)


mikko-lab:
Hey @gautamkishore — nice independent catch, and it's good validation that this bug is real and worth fixing: you landed on the exact same root cause and the same fix pattern (rebinding `messages` to a fresh list right after the shallow `kwargs.copy()`) for all four handlers.

Flagging two things so effort isn't duplicated / no one's working off stale info:

1. **Overlap**: #2418 already fixes this same Category 1 scope (`OpenAIToolsHandler`, `OpenAIJSONSchemaHandler`, `OpenAIParallelToolsHandler`, `OpenAIResponsesToolsHandler` — same 4 handlers, same fix). Since both PRs touch the same lines in `instructor/v2/providers/openai/handlers.py`, whichever lands second will conflict with the other.

2. **The "JSON/MD_JSON are unaffected" note needs correcting**: that was the original issue text, but it turned out to be wrong — I said the same thing initially and had to walk it back. `OpenAIJSONHandler`/`OpenAIMDJSONHandler` don't just rebuild `messages` safely; they mutate the *input* list/dicts in place (`.insert()`, `.append()`, `message["content"] += ...`) before that rebuild ever runs, so they corrupt the caller's `messages` on **every call**, not just reask ones — actually worse than the Category 1 bug. See the correction here: https://github.com/567-labs/instructor/issues/2417#issuecomment-4916843937. #2418 covers this too (via a new `copy_messages_for_mutation()` helper).

There's also #2419, which covers the remaining Category 2/3 providers (Mistral, Cohere, Writer, xAI, OpenRouter) if you're looking for more scope that isn't already spoken for.

Given the overlap, might be worth closing this one in favor of #2418 (which is a superset), or the maintainers can pick whichever they'd rather review — not my call to make, just didn't want the duplication or the outdated claim to slip through unnoticed. Thanks for jumping on this either way.

gautamkishore:
Closing in favor of #2418 which covers this scope and more. Thanks @mikko-lab for the heads up.

mikko-lab:
Appreciate the quick, gracious close, @gautamkishore — that's exactly the right call and saved everyone a merge headache. Good bug to have caught independently either way.

PR #2425, opened 2026-07-10
fix(v2): make ModeRegistry.get_handlers() thread-safe with double-checked locking

### LINKED ISSUE #2422: Thread-unsafe lazy loading in ModeRegistry.get_handlers() causes RegistryError under concurrency
- [x] I have tried asking for help in the community on discord or discussions and have not received a response.

**What Model are you using?**

- [ ] gpt-3.5-turbo
- [ ] gpt-4-turbo
- [ ] gpt-4
- [x] Other (please specify): N/A — this is provider-agnostic. Reproduced via `instructor.from_litellm` against AWS Bedrock (Nova/Claude) models, but the bug is entirely inside `instructor`'s mode-handler registry and never reaches the model/provider. The minimal repro below doesn't call any model at all.

**Describe the bug**

`ModeRegistry.get_handlers()` in `instructor/v2/core/registry.py` is not thread-safe. On the *first* concurrent access to a given `(provider, mode)` key in a process, exactly one thread wins and every other thread that was already dispatched raises `KeyError` (surfaced elsewhere as `RegistryError`), permanently — there is no retry, and the failure is not intermittent, it happens on nearly every request in the batch.

Root cause — `get_handlers` (simplified):

```python
    mode_key = (provider, normalize_mode(provider, mode))
        loader = self._lazy_loaders.pop(mode_key)   # <-- mutates shared dict, no lock
        handlers = loader()                          # <-- imports a module; can be slow
        self._handlers[mode_key] = handlers          # <-- only set AFTER loader() returns
        return handlers
    raise KeyError(...)                               # <-- every other concurrent caller lands here
```

`self._handlers` and `self._lazy_loaders` are plain `dict`s on a module-level singleton (`mode_registry`), shared across all threads, with no lock around this check → pop → import → set sequence.

Sequence of events with N threads calling `get_handlers(Provider.OPENAI, Mode.TOOLS)` concurrently for the first time in a process:

1. `instructor.v2.core.registry` registers a *lazy loader* (not yet the real handler) for every `(provider, mode)` pair at **import time**, via `_register_default_lazy_handlers()`. So `is_registered()` returns `True` from the very start of the process — that check is not where the race lives.
2. Thread A calls `get_handlers`, finds the key in `_lazy_loaders`, **pops it**, and starts the (relatively slow) module import (`__import__("instructor.v2.providers.openai.handlers")`, which runs that module's top-level `@register_mode_handler(...)` decorators).
3. While that import is in flight, every other thread that already passed its own initial check evaluates `mode_key in self._handlers` (not yet set — thread A hasn't finished) and `mode_key in self._lazy_loaders` (already popped by thread A) — both `False` — and falls through to `raise KeyError`.
4. Net effect: exactly one thread wins the race and succeeds; every thread that was in flight before the winner's import finished fails permanently for that call — the lazy loader entry is gone, so there is no retry path.
5. Once the first successful `get_handlers()` call completes anywhere in the process, `_handlers[mode_key]` is permanently populated and every later call (even concurrent ones) succeeds. So only the very first concurrent burst per `(provider, mode)` key, per process, is affected — but it affects nearly the entire burst.

This was found via `instructor.from_litellm(...)` (used to talk to AWS Bedrock through litellm), which hardcodes `provider=Provider.OPENAI` and calls `patch_v2(func=completion, provider=Provider.OPENAI, mode=Mode.TOOLS)` — so any concurrent structured-generation workload that creates a fresh `instructor` client per request/thread (a common and recommended pattern to avoid sharing a *client* across threads) hits this on its first batch, regardless of which real provider/model is behind the `completion` callable.

**To Reproduce**

No API key, network access, or model call needed — this hits the registry directly and fails before any request would be sent:

```python
# save as repro_registry_race.py, run in a FRESH process each time
from concurrent.futures import ThreadPoolExecutor


def check(i):
    try:
        mode_registry.get_handlers(Provider.OPENAI, Mode.TOOLS)
        return i, "OK"
    except Exception as e:
        return i, f"{type(e).__name__}: {e}"[:100]


with ThreadPoolExecutor(max_workers=10) as pool:
    results = list(pool.map(check, range(50)))

fails = [r for r in results if r[1] != "OK"]
print(f"failures: {len(fails)}/50")
for i, err in fails[:5]:
    print(i, err)
```

Run with `python repro_registry_race.py` — must be a fresh interpreter process each time; re-running in the same process will not reproduce it, since the registry is already warm after the first successful resolution.

Observed output (unpatched `instructor`, 3/3 fresh-process attempts):

```
failures: 49/50
1 KeyError: "Mode (<Provider.OPENAI: 'openai'>, <Mode.TOOLS: 'tool_call'>) is not registered. Availabl
2 KeyError: "Mode (<Provider.OPENAI: 'openai'>, <Mode.TOOLS: 'tool_call'>) is not registered. Availabl
3 KeyError: "Mode (<Provider.OPENAI: 'openai'>, <Mode.TOOLS: 'tool_call'>) is not registered. Availabl
4 KeyError: "Mode (<Provider.OPENAI: 'openai'>, <Mode.TOOLS: 'tool_call'>) is not registered. Availabl
5 KeyError: "Mode (<Provider.OPENAI: 'openai'>, <Mode.TOOLS: 'tool_call'>) is not registered. Availabl
```

Confirmed present in `instructor==1.15.3` and `instructor==1.15.4` (latest at time of writing). Confirmed **absent** in `instructor==1.15.1` and `1.15.2`, which predate the `instructor.v2` registry module entirely and use a different (non-lazy, non-racy) mode-dispatch mechanism in `instructor/processing/response.py`. So this is a regression introduced when the `v2` registry/lazy-loading architecture landed, between `1.15.2` and `1.15.3`.

Via any real downstream caller that does `instructor.from_litellm(litellm.completion)` fresh per thread inside a `ThreadPoolExecutor` (e.g. a batched/parallel structured-generation helper), this surfaces as almost every request in the first batch of a fresh process failing with:

```
RegistryError: Mode Mode.TOOLS is not registered for provider Provider.OPENAI. Available modes: [...]
```

even though the actual backing model/provider (Bedrock, Anthropic, etc.) is never OpenAI — `from_litellm` just hardcodes `Provider.OPENAI` for registry-lookup purposes.

**Expected behavior**

A losing thread should either block and wait for the winning thread's in-flight resolution to complete and reuse its result, or retry the lookup, rather than permanently raising `KeyError`/`RegistryError` for a mode that genuinely is registered (just not finished resolving yet). Concretely, `ModeRegistry.get_handlers()`'s pop → import → set sequence needs a lock — e.g. a per-`mode_key` lock (or a single registry-wide lock held across the full lazy-load resolution) — so concurrent callers serialize on the same in-flight load instead of racing to pop a shared dict entry.

**Screenshots**

N/A — this is a pure control-flow/concurrency bug with no visual component. See the reproduction output above.

**Workaround** (for anyone hitting this before it's fixed upstream): force a synchronous, single-threaded call to `mode_registry.get_handlers(provider, mode)` for every `(provider, mode)` pair you'll use, once, before starting any concurrent/threaded structured-generation calls. This resolves the lazy loader into `_handlers` up front, so no thread ever races the pop/import/set sequence.

nexiouscaliver:
## Summary

Fixes the thread-safety race condition in `ModeRegistry.get_handlers()` reported in #2422.

Concurrent first-access to a lazy-loaded `(provider, mode)` pair caused **49/50 threads** to fail with `KeyError`/`RegistryError` because the winning thread popped the lazy loader entry from `_lazy_loaders` before the others could use it. Losing threads found the key absent from both `_handlers` and `_lazy_loaders` with no retry path.

## Root Cause

The check → pop → import → set sequence in `get_handlers()` operated on shared module-level dicts with no synchronization:

```python
# RACY: no lock
    loader = self._lazy_loaders.pop(mode_key)   # ← gone for other threads
    handlers = loader()                          # ← slow import
    self._handlers[mode_key] = handlers          # ← set only after import returns
    return handlers
raise KeyError(...)                               # ← every concurrent loser lands here
```

## Fix

Added `threading.Lock` to `ModeRegistry` with double-checked locking:

- **Fast path** (no lock): `_handlers` dict read — safe under the GIL for lookups, and once a handler is cached it is never removed
- **Slow path** (locked): `_lock` protects the entire pop → `loader()` → set sequence so concurrent callers serialize on the same in-flight load
- Also synchronized `register()` and `register_lazy()` since they mutate the same shared dicts

## Verification

New regression test in `tests/core/test_registry_thread_safety.py`:
- **`test_concurrent_first_access_no_failures`**: 50 threads via `ThreadPoolExecutor(max_workers=10)` hit the same lazy-loaded mode simultaneously. Before fix: 49/50 fail. After fix: 0/50 fail.
- **`test_loader_called_exactly_once`**: Confirms the lock serializes lazy loading — loader is invoked exactly once even under heavy concurrency.
- **`test_double_checked_locking_resolves_quickly`**: Verifies fast path after first resolution.
- **`test_unregistered_mode_still_raises`**: Ensures `KeyError` is still raised for genuinely unregistered modes.
- **`test_concurrent_different_modes`**: Concurrent access to different keys does not interfere.

```
$ pytest tests/core/test_registry_thread_safety.py -v
5 passed in 0.67s
```

## Files Changed

- `instructor/v2/core/registry.py` — added `threading.Lock`, double-checked locking in `get_handlers()`, synchronized `register()` and `register_lazy()`
- `tests/core/test_registry_thread_safety.py` — new regression tests
- `CHANGELOG.md` — entry under `[Unreleased]`

Closes #2422

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **v2 registry**: Fix thread-safety race condition in `ModeRegistry.get_handlers()` where concurrent first-access to a lazy-loaded mode caused `KeyError`/`RegistryError` for all threads except the winner (49/50 failures in testing). Added double-checked locking with `threading.Lock` around the lazy-load path. ([#2422](https://github.com/567-labs/instructor/issues/2422))
+
+---
+
 ## [1.15.5] - 2026-06-28
 
 ### Fixed
--- a/instructor/v2/core/registry.py
+++ b/instructor/v2/core/registry.py
@@ -6,6 +6,7 @@ Supports lazy loading, dynamic registration, and queryable API.
 
 from __future__ import annotations
 
+import threading
 from dataclasses import dataclass
 from typing import Callable
 
@@ -80,6 +81,7 @@ class ModeRegistry:
         """Initialize empty registry."""
         self._handlers: dict[tuple[Provider, Mode], ModeHandlers] = {}
         self._lazy_loaders: dict[tuple[Provider, Mode], Callable[[], ModeHandlers]] = {}
+        self._lock = threading.Lock()
 
     def register(
         self,
@@ -110,18 +112,19 @@ class ModeRegistry:
             ConfigurationError: If mode is already registered with different handlers
         """
         mode_key = (provider, mode)
-        if mode_key in self._lazy_loaders:
-            self._lazy_loaders.pop(mode_key, None)
+        with self._lock:
+            if mode_key in self._lazy_loaders:
+                self._lazy_loaders.pop(mode_key, None)
 
-        self._handlers[mode_key] = ModeHandlers(
-            request_handler=request_handler,
-            reask_handler=reask_handler,
-            response_parser=response_parser,
-            stream_extractor=stream_extractor,
-            stream_extractor_async=stream_extractor_async,
-            message_converter=message_converter,
-            template_handler=template_handler,
-        )
+            self._handlers[mode_key] = ModeHandlers(
+                request_handler=request_handler,
+                reask_handler=reask_handler,
+                response_parser=response_parser,
+                stream_extractor=stream_extractor,
+                stream_extractor_async=stream_extractor_async,
+                message_converter=message_converter,
+                template_handler=template_handler,
+            )
 
     def register_lazy(
         self,
@@ -144,10 +147,11 @@ class ModeRegistry:
         from instructor.v2.core.errors import ConfigurationError
 
         mode_key = (provider, mode)
-        if mode_key in self._handlers or mode_key in self._lazy_loaders:
-            raise ConfigurationError(f"Mode {mode_key} is already registered")
+        with self._lock:
+            if mode_key in self._handlers or mode_key in self._lazy_loaders:
+                raise ConfigurationError(f"Mode {mode_key} is already registered")
 
-        self._lazy_loaders[mode_key] = loader
+            self._lazy_loaders[mode_key] = loader
 
     def get_handlers(self, provider: Provider, mode: Mode) -> ModeHandlers:
         """Get all handlers for a mode.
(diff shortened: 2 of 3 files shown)


nexiouscaliver:
Hey @jxnl @ivanleomk — this fixes the thread-safety race condition in `ModeRegistry.get_handlers()` reported in #2422. The bug causes ~49/50 concurrent threads to permanently fail with `KeyError` on first lazy-load of any `(provider, mode)` pair. Would appreciate a review when you get a chance!

**TL;DR:** The `check → pop → import → set` sequence in `get_handlers()` had no lock. Added double-checked locking with `threading.Lock`. Includes regression tests that reproduce the 49/50 failure before the fix and 0/50 after.

nexiouscaliver:
Friendly nudge! Just pushed a cleanup commit fixing ruff lint/format in the test file. CI workflows need maintainer approval to run on fork PRs — could someone approve? 5/5 tests pass, ruff clean. @jxnl @ivanleomk

jxnl:
Superseded by #2495, which uses a lazy-load-specific lock and avoids coupling registration to loader resolution.

PR #2340, opened 2026-06-02
fix(v2): consolidate minor runtime fixes

jxnl:
## Summary

Consolidates the small, straightforward v2 fixes that had accumulated as overlapping one-off PRs:

- Guard Cohere templating when first-turn calls provide `message` without `chat_history`.
- Guard Gemini JSON mode when `messages` is empty or absent.
- Preserve array-root JSON in codeblock and streaming extraction.
- Make `IterableBase.get_object` ignore braces inside JSON strings.
- Raise actionable missing-`jsonref` errors for Gemini and VertexAI structured-output paths.
- Emit retry attempt metadata through `parse:error`, `completion:error`, and `completion:last_attempt` hooks.

Closes #2335.
Closes #2331.
Closes #2288.
Closes #2222.
Absorbs #2328.

Also absorbs the already-closed duplicate PRs: #2339, #2338, #2337, #2336, #2333, #2332, #2312, #2308, #2330, #2324, #2317, #2285, #2251, #2320, #2318, #2309, #2295, and #2293.

## Verification

- `uv run pytest tests/v2/test_json_helpers.py tests/v2/test_iterable_streaming.py tests/v2/test_core_provider_dispatch.py tests/v2/test_gemini_utils_deterministic.py tests/v2/test_vertexai_runtime.py tests/v2/test_retry_runtime.py -q` -> 55 passed
- `uv run ruff check <modified python files>` -> passed
- `uv run ruff format --check <modified python files>` -> passed
- `uv run ty check instructor/v2/core/json.py instructor/v2/dsl/iterable.py instructor/v2/core/templating.py instructor/v2/core/hooks.py instructor/v2/core/retry.py` -> passed
- `git diff --check` -> passed

Note: the repo-wide pre-commit `ty` hook currently fails in this local environment because optional provider SDKs are not installed (`google.genai`, `mistralai`, `cohere`, `groq`, `writerai`, `boto3`, `vertexai`, etc.). The touched core modules pass `ty`, and the provider behavior is covered by targeted tests with fakes.

---

> [!NOTE]
> **Medium Risk**
> Changes touch core JSON streaming, iterable parsing, and the v2 retry/hook path used on every structured call, but behavior is narrowly scoped with strong test coverage and no auth or data-store changes.
> 
> **Overview**
> This PR **bundles several small v2 runtime fixes** that were previously scattered across duplicate PRs, with matching tests and an Unreleased changelog entry.
> 
[...]

PR #2426, opened 2026-07-11
fix(v2): don't mistake a literal '[' inside a preceding string field for the tasks array start

roli-lpci:
## Describe your changes

`IterableBase.tasks_from_chunks` / `tasks_from_chunks_async` (`instructor/v2/dsl/iterable.py`), the
streaming path every provider's `Iterable[Model]` extraction goes through, detects the start of the
streamed `tasks` array with a naive `"[" in chunk` check on each individual chunk. If a field
preceding `tasks` in the response JSON contains a literal `[` inside a string value — for example a
free-text `note` field like `"check [priority] items first"` — that in-string bracket falsely
triggers array-start detection. The buffer gets sliced from the wrong offset, and every task
streamed after that point is silently dropped: the function returns an empty list, with no
exception anywhere.

```python
chunks = [
    '{"note": "check [priority] items first", ',
    '"tasks": [',
    '{"name": "alpha", "priority": 1}',
    ', {"name": "beta", "priority": 2}',
    "]}",
]
list(IterableModel(Task).tasks_from_chunks(chunks))
# []  <- both tasks vanished, no error
```

The fix adds `_find_unquoted_char`, a small string-aware scanner (tracks `in_string`/escape state,
the same pattern `get_object` a few lines below already uses for brace-matching) and uses it in
both `tasks_from_chunks` and `tasks_from_chunks_async` to find the real, unquoted `[` against the
full accumulated buffer instead of a naive per-chunk substring check. 38 lines changed across both
sync/async entry points plus the new helper; no signature changes.

Tests: added `test_iterable_tasks_from_chunks_ignores_bracket_inside_preceding_string` and its
async counterpart, mirroring the existing `test_iterable_tasks_from_chunks_handles_braces_inside_strings`
pattern, using the `note`-field repro above. Before fix: 2 failed — `assert [] == [Task(name='alpha',
priority=1), Task(name='beta', priority=2)]` (both streamed tasks silently dropped). After fix: all
6 tests in `tests/v2/test_iterable_streaming.py` pass. `ruff check` is clean on both changed files.

No existing issue tracks this bug. Distinct from the already-merged PR #2340 (a consolidation
bundle whose iterable.py portion fixed brace-in-string detection in `get_object`) — this fixes a
different detection bug (array-start bracket, not object-boundary brace) in a different function;
`get_object` itself is unchanged.

- [ ] If it is a core feature, I have added documentation. (Bug fix to internal streaming
  detection; no documented behavior changed.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **v2 iterable streaming**: `Iterable[Model]` / `IterableModel` streaming (`tasks_from_chunks` / `tasks_from_chunks_async`, used by every provider's streaming path) no longer mistakes a literal `[` inside a string value emitted before the `tasks` array (e.g. a free-text field) for the start of the array. Previously this silently dropped every streamed task with no error.
+
+---
+
 ## [1.15.5] - 2026-06-28
 
 ### Fixed
--- a/instructor/v2/dsl/iterable.py
+++ b/instructor/v2/dsl/iterable.py
@@ -102,9 +102,10 @@ class IterableBase:
         for chunk in json_chunks:
             potential_object += chunk
             if not started:
-                if "[" in chunk:
+                bracket_index = cls._find_unquoted_char(potential_object, "[")
+                if bracket_index != -1:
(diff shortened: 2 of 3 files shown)


jxnl:
Already fixed on main by #2434, including regression coverage for bracket characters inside preceding string fields. Closing this now-redundant patch.

PR #2430, opened 2026-07-11
fix(citations): escape substring quotes before fuzzy span matching

### LINKED ISSUE #2431: CitationMixin crashes with regex.error when a quote contains regex metacharacters

**What Model are you using?**

- [x] Other (provider-agnostic — the bug is in `CitationMixin` validation, independent of the model)

**Describe the bug**

`CitationMixin._get_span` (`instructor/v2/dsl/citation.py`) interpolates the LLM-generated quote directly into a `regex` fuzzy-match pattern without escaping it:

```python
minor = quote
s = regex.search(f"({minor}){{e<={errs_}}}", major)
```

Because `quote` comes straight from the model's `substring_quotes`, any regex metacharacter it contains — unbalanced parentheses/brackets, `*`, `+`, `?`, `\`, etc., all common in ordinary prose — is treated as part of the pattern. This raises `regex.error` (crashing validation) or silently matches the wrong span, instead of the documented behavior of finding the span or dropping the quote.

**To Reproduce**

```python


class Answer(CitationMixin):
    pass


context = "The margin is 50% (approx) this quarter."

# A quote with an unbalanced parenthesis, as an LLM might return:
Answer.model_validate(
    {"substring_quotes": ["50% (approx"]},
    context={"context": context},
)
```

Result:

```
regex.error: missing ) at position 19
```

**Expected behavior**

The span should be resolved (the quote normalized back to the matching substring of the context), or, if it does not match, the quote should be dropped from `substring_quotes` — never a `regex.error`. Fuzzy (edit-distance) matching should keep working.

The fix is to escape the quote with `regex.escape()` before building the pattern. Proposed in #2430.

RohitMudili:
## Describe your changes

`CitationMixin._get_span` interpolated the LLM-generated quote directly into a
`regex` fuzzy-match pattern:

```python
minor = quote
s = regex.search(f"({minor}){{e<={errs_}}}", major)
```

Because the quote is never escaped, any regex metacharacter in the model's
`substring_quotes` — unbalanced parentheses/brackets, `*`, `+`, `?`, `\`, etc.,
all common in real prose — is interpreted as part of the pattern. This either
raises `regex.error` (crashing validation) or silently matches the wrong span,
instead of the documented behavior of *finding the span or dropping the quote*.

Reproduction:

```python
context = "The margin is 50% (approx) this quarter."
Answer.model_validate(
    {"substring_quotes": ["50% (approx"]},   # unbalanced paren from the LLM
    context={"context": context},
)
# regex.error: missing ) at position 19
```

### Fix

Escape the quote with `regex.escape()` before building the pattern so it is
matched literally. Fuzzy (edit-distance) matching is unaffected — a quote with a
one-character typo still resolves within the default edit budget.

### Changes
- `instructor/v2/dsl/citation.py`: `regex.escape(quote)` before the fuzzy search.
- `tests/v2/test_citation.py`: new tests — metacharacter quotes resolve instead of crashing, non-matching quotes are dropped, fuzzy matching still works, and no-context input is untouched. (These fail on `main` and pass with the fix.)
- `CHANGELOG.md`: entry under `[Unreleased]`.

Closes #2431 — https://github.com/567-labs/instructor/issues/2431

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **Citations**: `CitationMixin` now escapes substring quotes before fuzzy-matching them against the context, so quotes containing regex metacharacters (e.g. unbalanced parentheses or brackets) resolve their span instead of raising `regex.error` during validation.
+
+---
+
 ## [1.15.5] - 2026-06-28
 
 ### Fixed
--- a/instructor/v2/dsl/citation.py
+++ b/instructor/v2/dsl/citation.py
@@ -80,7 +80,10 @@ class CitationMixin(BaseModel):
     ) -> Generator[tuple[int, int], None, None]:
         import regex
 
-        minor = quote
+        # Escape the quote so regex metacharacters in LLM-generated text
+        # (e.g. unbalanced parentheses or brackets) are matched literally
+        # instead of crashing the fuzzy search with a regex compile error.
+        minor = regex.escape(quote)
         major = context
 
         errs_ = 0
--- a/tests/v2/test_citation.py
+++ b/tests/v2/test_citation.py
@@ -1,0 +1,82 @@
+"""Tests for CitationMixin span resolution.
+
+Covers the regex-escaping behavior in ``_get_span``: LLM-generated quotes may
+contain regex metacharacters, which must be matched literally rather than
+compiled as a pattern.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+pytest.importorskip("regex")
+
+from instructor.v2.dsl.citation import CitationMixin  # noqa: E402
+
+
+class Answer(CitationMixin):
+    pass
+
+
+def test_quote_with_regex_metacharacters_does_not_crash() -> None:
+    """A quote with unbalanced parentheses must resolve, not raise regex.error."""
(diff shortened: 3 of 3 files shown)


RohitMudili:
Hey @jxnl @ivanleomk — this fixes the `regex.error` crash in `CitationMixin` reported in #2431. When the model returns a `substring_quotes` value containing regex metacharacters (unbalanced parentheses/brackets, `*`, `\`, etc. — common in ordinary prose), validation blows up with `regex.error` instead of resolving the span or dropping the quote. Would appreciate a review when you get a chance!

**TL;DR:** `_get_span()` interpolated the raw LLM quote straight into a fuzzy pattern `f"({minor}){{e<={errs_}}}"` with no escaping. Wrapped it in `regex.escape()` so the quote is matched literally; fuzzy (edit-distance) matching is unaffected. Includes regression tests that fail on `main` and pass with the fix.

ErenAta16:
Applied this locally and ran the new test file against current `main`: all 8 pass, including `test_fuzzy_matching_still_works_after_escaping`, which was the thing I wanted to check, `regex.escape()` on the quote before wrapping it in the `{e<=N}` fuzzy syntax doesn't interfere with the edit-distance matching itself, since escaping only affects how literal characters are interpreted, not the fuzzy-match operator appended outside the escaped substring.

Only blocker is the `DIRTY` merge state, looks like just a `CHANGELOG.md` conflict from newer entries landing since this was opened, the actual fix and tests applied cleanly.

jxnl:

PR #2435, opened 2026-07-13
feat(requesty): add Requesty provider

Thibaultjaigu:
## Summary

Adds [Requesty](https://requesty.ai), an OpenAI-compatible LLM router, as a provider, mirroring the existing OpenRouter provider (`instructor/v2/providers/openrouter/`). Requesty exposes 400+ models through one endpoint, addressed as `provider/model`.

## Changes

- `instructor/v2/providers/requesty/`: `from_requesty()` factory + `RequestyJSONSchemaHandler` (mirrors the OpenRouter provider; reuses the generic OpenAI-compatible JSON-schema wiring)
- `instructor/v2/core/providers.py`: `Provider.REQUESTY` + `get_provider()` URL detection
- `instructor/v2/core/provider_specs.py`: `PROVIDER_SPECS` entry (TOOLS / JSON_SCHEMA / MD_JSON / PARALLEL_TOOLS)
- `instructor/v2/providers/openai/handlers.py`: Requesty added to the OpenAI-compat + parallel-tools provider sets (same as OpenRouter)
- `instructor/__init__.py`: export + lazy import for `from_requesty`
- `instructor/v2/auto_client.py`: `_build_requesty()` (base URL `https://router.requesty.ai/v1`, env `REQUESTY_API_KEY`) + `requesty` in the provider builders and `supported_providers`
- `docs/integrations/requesty.md` + `mkdocs.yml` nav + `examples/open_source_examples/requesty.py` + CHANGELOG entry

## Testing

Verified end-to-end through the real `from_provider("requesty/openai/gpt-4o-mini")` code path against the live endpoint: a structured extraction into a Pydantic model returned the expected value. `ruff check` is clean on the touched files.

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.


jxnl:
1.16 triage: keeping this provider addition open, but outside maintenance consolidation. The 12-file Requesty patch conflicts with current provider code and adds no deterministic or live tests in the branch. It needs a current-v2 reconstruction, provider/mode contract coverage, live verification, and a maintenance owner before merge.

Thibaultjaigu:
Thanks for the clear list. The branch is now rebased on current main and built on the v2 layout (instructor/v2/providers/requesty with client and handlers, Provider.REQUESTY, a PROVIDER_SPECS entry and the auto_client builder), and I added Requesty next to OpenRouter in every shared contract test: the parametrized handler matrix for TOOLS, JSON_SCHEMA, MD_JSON and PARALLEL_TOOLS, the mode normalization checks (legacy modes normalize, the OpenRouter structured outputs mode is not accepted for Requesty), the messages not mutated tests for prepare_request and reask, the client and registry tests, the auto_client tail coverage for env key, default base URL, missing dependency and factory failure paths, the public typing surface, get_provider detection, and the docs guide lint. Locally that is 3539 passing deterministic tests with Requesty producing the same case set and the same skips as OpenRouter. For live verification I added requesty/openai/gpt-4o-mini to tests/llm/test_current_doc_models.py and wired REQUESTY_API_KEY into the workflow next to the OpenRouter secret; the case passes against the real router, and I am happy to provide a key for CI. I work at Requesty and will be the maintenance owner for this provider, so ping me on any breakage or follow up and I will handle it.

PR #2447, opened 2026-07-15
fix(bedrock): forward default model from provider

ShiroKSH:
## Summary

- Forward the model parsed by the Bedrock auto-client to the Bedrock client factory.
- Add a regression assertion for the forwarded default model.

## Root cause

The Bedrock auto-client selected a mode from the model string but never passed that string as the factory's default model. Calls that omitted `model` therefore reached Bedrock without the required `modelId`.

## Validation

- Targeted auto-client and Bedrock coverage tests: 31 passed.
- Ruff check for the two changed files passed.

### DIFF
--- a/instructor/v2/auto_client.py
+++ b/instructor/v2/auto_client.py
@@ -1023,6 +1023,7 @@ def _build_bedrock(
             client,
             mode=default_mode,
             async_client=async_client,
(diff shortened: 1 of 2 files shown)


jxnl:

PR #2448, opened 2026-07-15
fix(openai): handle responses with no tool_calls in reask_tools to avoid TypeError

Kropiunig:
## Problem

`reask_tools()` in `instructor/v2/providers/openai/handlers.py` is the reask handler for `Mode.TOOLS` (registered for OpenAI plus the OpenAI-compatible providers Together, Groq, Fireworks, Cerebras, DeepSeek, OpenRouter, Anyscale and Databricks) and is also used by `OpenAIParallelToolsHandler.handle_reask`.

On the non-streaming path it iterated the tool calls with no guard:

```python
reask_msgs: list[Any] = [dump_message(response.choices[0].message)]
for tool_call in response.choices[0].message.tool_calls:
    ...
```

**Failing scenario:** under `Mode.TOOLS`, a model returns a plain-text assistant message with **no** tool call. This is common with OpenAI-compatible providers that don't honor forced `tool_choice`, or a model that just answers in prose. Parsing then raises `ResponseParsingError` (which is retryable), and the retry loop calls `reask_tools`. But the OpenAI SDK sets `message.tool_calls` to `None` when absent, so `for tool_call in None` raises `TypeError: 'NoneType' object is not iterable`. Since `TypeError` is not retryable, the reask/retry flow aborts and the failure is wrapped into a confusing `InstructorRetryException` around a `TypeError` — the model is never actually re-asked.

## Repro

```python
from openai.types.chat import ChatCompletion
from instructor.v2.providers.openai.handlers import reask_tools

resp = ChatCompletion.model_validate({
    "id": "x", "object": "chat.completion", "created": 1, "model": "gpt-test",
    "choices": [{"index": 0, "finish_reason": "stop",
                 "message": {"role": "assistant", "content": "The answer is many."}}],
})
# resp.choices[0].message.tool_calls is None
reask_tools({"messages": [{"role": "user", "content": "How many?"}]},
            resp, ValueError("answer must be an int"))
# -> TypeError: 'NoneType' object is not iterable
```

## Fix

Normalize `message.tool_calls` to an empty list, and when there is no tool call, append a plain user correction instead of iterating `None`. This mirrors the existing streaming branch of `reask_tools` (which already falls back to a user correction) and the sibling `reask_genai_tools` handler, which explicitly handles the no-function-call case and has its own regression test. The tool-call path is unchanged.

Adds a regression test (`test_tools_reask_without_tool_call_adds_a_user_correction`) covering the no-tool-call case, and a CHANGELOG entry under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 - **Mistral/Vertex AI partial streaming**: Avoid forwarding iterable-only parser arguments into completed `Partial` responses, preventing final Pydantic validation errors for sync and async streams.
 - **Batch providers**: Handle missing optional SDKs safely, validate OpenAI batch input before client setup, report exhausted output-file retries clearly, and remove unreachable fallbacks.
 - **OpenAI/Writer tools**: Raise clear response-parsing errors for completions with no choices or tool calls instead of leaking attribute and index errors.
+- **v2 OpenAI tools reask**: Fall back to a plain user correction when a `Mode.TOOLS` response contains no tool calls (for example, OpenAI-compatible providers that ignore forced `tool_choice`, or a model that answers in prose) instead of iterating `None` and raising `TypeError` during reask.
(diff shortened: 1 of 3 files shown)


jxnl:

PR #2451, opened 2026-07-16
fix(openai): copy schema before strict mutation to prevent lru_cache poisoning

### LINKED ISSUE #2450: fix(openai): lru_cache poisoning when strict=True — schema persists across calls
## Bug

`generate_openai_schema` in `instructor/v2/providers/openai/schema.py` is decorated with `@functools.lru_cache(maxsize=256)`. It returns the **same dict object** on every call for a given model.

`OpenAIToolsHandler.prepare_request` writes `schema["strict"] = True` directly into that cached dict when `strict=True` is requested. This permanently mutates the cache entry, so all future calls for the same model  even those that never requested strict mode — receive a schema that already carries `"strict": True`.

## Reproduction

```python
from instructor.v2.providers.openai.schema import generate_openai_schema

class MyModel(BaseModel):
    value: int

s1 = generate_openai_schema(MyModel)
print("before:", "strict" in s1)   # False

# Simulate what the handler does when strict=True
s1["strict"] = True                 # mutates the cached object directly

s2 = generate_openai_schema(MyModel)
print("after:", "strict" in s2)    # True — cache poisoned
print("same object:", s1 is s2)    # True
```

## Impact

Once any call for a model is made with `strict=True`, every subsequent call for that model within the same process silently enables strict mode — even calls that never passed `strict=True`. The caller gets no error or warning; the schema is just silently wrong.

Affects long-running servers where one endpoint uses `strict=True` for a model that other endpoints use without it.

## Fix

Shallow-copy the cached dict before mutating:

```python
# Before:
schema = generate_openai_schema(response_model)

# After:
schema = dict(generate_openai_schema(response_model))
```

PR with fix and regression tests: coming.

RudraDudhat2509:
## Summary

- `generate_openai_schema` uses `@functools.lru_cache`, returning the same dict object on every call for a given model
- `OpenAIToolsHandler.prepare_request` was writing `schema["strict"] = True` directly into that cached object, permanently tainting it for all subsequent callers in the same process
- Fix: shallow-copy the dict before mutating (`schema = dict(generate_openai_schema(response_model))`)

Closes #2450

## What changed

`instructor/v2/providers/openai/handlers.py`  one-line fix, added comment explaining why the copy is required.

## Tests

Two regression tests added:

- `tests/core/test_schema_utils.py::test_cached_schema_same_object`  documents that `lru_cache` returns the same object (the precondition for the bug)
- `tests/v2/test_openai_compat_handlers.py::test_strict_mode_does_not_pollute_schema_cache`  calls `prepare_request` with `strict=True`, then again without it, and asserts the second result carries no `"strict"` key. This test fails on the unfixed code and passes after the fix.

## Test plan

- [ ] `uv run pytest tests/core/test_schema_utils.py tests/v2/test_openai_compat_handlers.py -v`
- [ ] `uv run ruff check instructor/v2/providers/openai/handlers.py tests/core/test_schema_utils.py tests/v2/test_openai_compat_handlers.py`

### DIFF
--- a/instructor/v2/providers/openai/handlers.py
+++ b/instructor/v2/providers/openai/handlers.py
@@ -648,7 +648,11 @@ class OpenAIToolsHandler(OpenAIHandlerBase):
             new_kwargs["tools"] = handle_parallel_model(cast(Any, response_model))
             new_kwargs["tool_choice"] = "auto"
         else:
-            schema = generate_openai_schema(response_model)
+            # Shallow-copy to avoid mutating the lru_cache return value.
+            # generate_openai_schema caches the same dict object; writing
+            # "strict" into it would permanently affect every future call for
+            # this model, even when strict=False.
+            schema = dict(generate_openai_schema(response_model))
 
             # Check for strict parameter
             use_strict = new_kwargs.pop("strict", False)
--- a/tests/core/test_schema_utils.py
+++ b/tests/core/test_schema_utils.py
@@ -215,5 +215,24 @@ def test_anthropic_schema_uses_openai_base():
     assert anthropic_schema["input_schema"] == TestModel.model_json_schema()
 
 
+def test_cached_schema_same_object():
+    """generate_openai_schema returns the same dict object on every call (lru_cache).
+
+    This means callers MUST NOT mutate the returned dict directly; they should
+    copy it first. OpenAIToolsHandler.prepare_request does this by shallow-copying
+    before writing "strict" into the schema.
+    """
+
+    class ImmutableModel(BaseModel):
+        value: int
+
+    s1 = generate_provider_openai_schema(ImmutableModel)
+    s2 = generate_provider_openai_schema(ImmutableModel)
+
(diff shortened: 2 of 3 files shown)


ErenAta16 (commented):
Verified this is real: generate_openai_schema is lru_cache'd, so schema["strict"] = True on the returned object does leak across calls for the same model. Checked the other v2 providers that call the same generate_openai_schema function (writer/handlers.py, mistral/handlers.py) to see if they have the same mutate-in-place problem - they don't, both just pass the schema through into new_kwargs["tools"] without writing into it, so this isn't a partial fix for a wider pattern, it's the one place that actually needed it. Test coverage is solid, the same-object assertion in test_cached_schema_same_object is a good way to pin the precondition down explicitly instead of just testing the symptom.

RudraDudhat2509:
thanks for the thorough review! let me know if anything's needed from my end

ErenAta16:
Nothing further needed from my side, the fix and tests are solid as they are.

jxnl:

PR #2453, opened 2026-07-17
fix(mistral): raise retryable error and reask correctly when response has no tool_calls

sohumt123:
## What happens

When a Mistral model replies with a plain-text assistant message and no tool call (models that ignore `tool_choice="any"`, content-filter/refusal text, or OpenAI-compatible gateways routed through the Mistral provider), `message.tool_calls` is `None` and TOOLS mode crashes twice:

1. **Parse layer** — `MistralHandlerBase._extract_tool_call_json` indexes `response.choices[0].message.tool_calls[0]` and raises `TypeError: 'NoneType' object is not subscriptable`. `TypeError` is not in `_RETRYABLE_PARSE_ERRORS` (`instructor/v2/core/retry.py`), so tenacity reraises immediately: the user gets an `InstructorRetryException` wrapping a bare `TypeError` after exactly **1** attempt, regardless of `max_retries`. The model is never re-asked.
2. **Reask layer** — even if parsing raised a retryable error, `MistralToolsHandler.handle_reask` iterates `for tool_call in response.choices[0].message.tool_calls:` on the same `None` and raises `TypeError: 'NoneType' object is not iterable`.

The parallel-tools generator in `parse_response` has the same unguarded iteration.

## Repro (no API key needed)

Driving the registered `(Provider.MISTRAL, Mode.TOOLS)` handlers with a fake `create()` that returns a prose response first and a valid tool call second:

- before: parse raises `TypeError: 'NoneType' object is not subscriptable`; reask raises `TypeError: 'NoneType' object is not iterable`; `retry_sync_v2` with `max_retries=2` aborts after **1** API call.
- after: parse raises a retryable `ResponseParsingError`, reask appends a plain user correction, and the retry loop makes a **2nd** API call and returns the validated model.

## Fix

Mirror the existing OpenAI handler contract in the Mistral provider:

- `_extract_tool_call_json` raises `ResponseParsingError("No tool calls found in Mistral response", ...)` when `tool_calls` is `None`/empty — retryable, so the retry machinery re-asks (the OpenAI parse path already does this).
- `handle_reask` falls back to a plain `user` correction message when there are no tool calls — there is no `tool_call_id` to attach a `tool`-role message to. Same shape as the OpenAI-side fix in #2448.
- The parallel-tools generator iterates `tool_calls or []` so it yields nothing instead of crashing.

This is the Mistral counterpart of #2448 (which touches only `instructor/v2/providers/openai/handlers.py`), plus the parse-layer guard that has no OpenAI analog.

## Testing

- New unit tests in `tests/v2/test_mistral_handlers.py` and `tests/coverage/test_mistral_coverage.py` (mock responses, no API key). All 4 fail on `main` with the exact `TypeError`s above and pass with this change.
- Full offline suites pass locally with the CI commands: `tests/v2` (1585 passed, 164 skipped), coverage job (`tests/coverage` + `tests/test_batch_processor_coverage.py`: 742 passed), and the full-coverage job (`pytest tests/ -k 'not docs'`: 2834 passed; coverage report 99%, 0 missing statements; `instructor/v2/providers/mistral/handlers.py` at 100% statement + branch).
- `ruff check` / `ruff format --check` on `instructor examples tests` and `ty check` (including `--python-version 3.9 --python-platform all`) all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 - **Mistral/Vertex AI partial streaming**: Avoid forwarding iterable-only parser arguments into completed `Partial` responses, preventing final Pydantic validation errors for sync and async streams.
 - **Batch providers**: Handle missing optional SDKs safely, validate OpenAI batch input before client setup, report exhausted output-file retries clearly, and remove unreachable fallbacks.
 - **OpenAI/Writer tools**: Raise clear response-parsing errors for completions with no choices or tool calls instead of leaking attribute and index errors.
+- **Mistral tools**: Raise a retryable response-parsing error when a completion contains no tool calls and fall back to a plain user correction during reask, so prose-only replies trigger a re-ask instead of aborting the retry loop with `TypeError`.
 - **Fireworks streaming**: Keep non-streaming async calls non-streaming and return streaming async generators without incorrectly awaiting them.
 - **GenAI uploads**: Respect `max_retries=0` without an unwanted sleep or polling request, allow recovery on the final permitted retry, and report nameless pending uploads clearly before polling.
 - **Gemini/GenAI messages**: Honor an explicit system message for unstructured requests, remove the unsupported raw `system` argument, and reject invalid scalar message content clearly.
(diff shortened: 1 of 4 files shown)


jxnl:

PR #2455, opened 2026-07-17
fix(cache): isolate retry-loop kwargs so cache store key matches lookup key

### LINKED ISSUE #2454: cache= silently stops caching any request that needed a retry (lookup/store key divergence)

**Describe the bug**

`cache=` silently stops working for any request that needed at least one retry. The result returned to the caller is still correct, but nothing gets cached under a key that will ever be looked up again, so an identical follow-up request always misses the cache and triggers a brand-new real LLM call.

**Root cause**

`_create_sync_wrapper` (`instructor/v2/core/patch.py`, mirrored in `_create_async_wrapper`) computes the cache lookup key from `new_kwargs.get("messages")` before calling `retry_sync_v2`, and recomputes the cache store key from the same `new_kwargs.get("messages")` after `retry_sync_v2` returns:

```python
# lines 234-246, lookup
if cache is not None and response_model is not None:
    key = make_cache_key(messages=new_kwargs.get("messages") or ..., model=new_kwargs.get("model"), ...)
    cached = load_cached_response(cache, key, response_model)
    if cached is not None:
        return cached

response = retry_sync_v2(..., kwargs=new_kwargs, ...)  # new_kwargs passed by reference, line 259

# lines 264-283, store
if cache is not None and response_model is not None:
    key = make_cache_key(messages=new_kwargs.get("messages") or ..., model=new_kwargs.get("model"), ...)
    store_cached_response(cache, key, response, ttl=cache_ttl)
```

`new_kwargs` is passed into `retry_sync_v2` by reference, not copied. When the first attempt fails Pydantic validation, `retry_sync_v2` (`instructor/v2/core/retry.py`) calls the provider's reask handler, e.g. `reask_tools` (`instructor/v2/providers/openai/handlers.py:157-190`, and identically `reask_md_json`, `reask_default`, `reask_responses_tools` in the same file, plus Anthropic's `handle_reask`):

```python
def reask_tools(...):
    kwargs = kwargs.copy()          # shallow copy: new dict, SAME "messages" list object
    ...
    kwargs["messages"].extend(reask_msgs)   # mutates that shared list in place
```

`kwargs.copy()` only copies the top-level dict. `kwargs["messages"]` is still the exact same list object `patch.py`'s outer `new_kwargs["messages"]` points to, so `.extend()` mutates it too. By the time `retry_sync_v2` returns, `new_kwargs["messages"]` now includes the failed tool-call turn and the "Validation Error found..." reask turn, so the cache *store* key computed from it no longer matches the cache *lookup* key computed before the retry happened. The successful result gets cached under a key nobody will ever compute again for that logical request.

**To Reproduce**

```python
from instructor.v2.core.patch import patch_v2
from instructor.cache import AutoCache
from tests.coverage._openai import chat_completion, tool_call

class Answer(BaseModel):
    value: int
    @field_validator("value")
    @classmethod
    def must_be_positive(cls, v):
        if v < 0:
            raise ValueError("value must be positive")
        return v

call_count = {"n": 0}
def fake_create(**kwargs):
    call_count["n"] += 1
    if call_count["n"] == 1:
        return chat_completion(tool_calls=[tool_call("Answer", {"value": -5})])  # fails validation
    return chat_completion(tool_calls=[tool_call("Answer", {"value": 42})])

create_fn = patch_v2(fake_create, provider=Provider.OPENAI, mode=Mode.TOOLS)
cache = AutoCache()
original_messages = [{"role": "user", "content": "What is 6 times 7?"}]

r1 = create_fn(model="gpt-4o-mini", messages=list(original_messages), response_model=Answer, max_retries=2, cache=cache)
print("calls after 1st create() (needed 1 retry):", call_count["n"])

r2 = create_fn(model="gpt-4o-mini", messages=list(original_messages), response_model=Answer, max_retries=2, cache=cache)
print("calls after 2nd, IDENTICAL create() (should stay the same, cache hit):", call_count["n"])
```

Actual output (ran against current `main`, `patch_v2` → real `mode_registry` handlers → real `retry_sync_v2` → real `instructor.cache`, `fake_create` standing in only for the network call):
```
calls after 1st create() (needed 1 retry): 2
calls after 2nd, IDENTICAL create() (should stay the same, cache hit): 3
```

**Expected behavior**

The second `create_fn(...)` call, with byte-identical arguments to the first, should hit the cache and not touch `fake_create` at all, keeping the call count at 2.

**Environment**

Reproduced against current `main` (`47fdb2c`, "fix(v2): consolidate correctness fixes (#2434)"). This is closely related to #2417/#2428 (fixed by #2434), which fixed `prepare_request` mutating the *caller's own* `messages` list; the reask handlers themselves (`reask_tools`, `reask_md_json`, `reask_default`, `reask_responses_tools`, and Anthropic's `handle_reask`) still mutate the request-local list in place via the same "shallow `.copy()` doesn't copy nested containers" pattern, which is invisible to a caller but corrupts `patch.py`'s own before/after cache-key computation.

**Suggested fix**

Either (a) have each `reask_*`/`handle_reask` function rebind `kwargs["messages"]` to a new list before appending/extending (`kwargs["messages"] = [*kwargs["messages"]]`), mirroring the fix already applied to `prepare_request` for #2417/#2428, or (b) have `patch.py` pass an isolated copy into `retry_sync_v2` (`{**new_kwargs, "messages": list(new_kwargs.get("messages", []))}`) so the outer `new_kwargs` used for the cache-store key can never observe retry-loop mutations regardless of what individual reask handlers do internally. Doing (b) alone is probably the more robust fix, since it closes the door for any future reask handler making the same mistake.

ErenAta16:
Fixes #2454.

## What's broken

`_create_sync_wrapper`/`_create_async_wrapper` (`instructor/v2/core/patch.py`) compute a cache lookup key from `new_kwargs["messages"]` before calling `retry_sync_v2`/`retry_async_v2`, and recompute the store key from the same `new_kwargs["messages"]` after the retry call returns, but pass `new_kwargs` into the retry call by reference.

Each reask handler (`reask_tools`, `reask_md_json`, `reask_default`, `reask_responses_tools` in `providers/openai/handlers.py`, and Anthropic's `handle_reask`) does `kwargs = kwargs.copy()` then `kwargs["messages"].append(...)`/`.extend(...)`. The shallow dict copy doesn't copy the `messages` list, so this mutates the exact list object `patch.py`'s `new_kwargs` still points to.

Any request that needed at least one retry ends up with a store key computed from the post-retry, reask-polluted `messages` list, different from the lookup key computed from the pristine one before the retry loop ran. The result is correct, but it gets cached under a key nothing will ever look up again, so an identical follow-up request always misses and re-triggers a real LLM call.

## Fix

Added `isolate_retry_kwargs()` in `messages.py`, right next to the existing `copy_messages_for_mutation` helper (added for #2417/#2428, which fixed the analogous problem one layer up in `prepare_request`). It returns `kwargs` with a shallow copy of whichever of `messages`/`contents`/`chat_history` is present, so mutations during the retry loop land on a private list and can never leak back into the `kwargs` dict `patch.py` reads again for the cache store key, regardless of what an individual reask handler does internally.

Used it at both call sites where `new_kwargs` is handed to the retry layer (sync and async).

## Testing

Added `test_auto_cache_prevents_duplicate_calls_after_a_retry` in `tests/cache/test_cache_integration.py`, next to the existing `test_auto_cache_prevents_duplicate_provider_calls`. It forces exactly one retry (first fake response fails a Pydantic validator, second succeeds), then makes an identical second call and asserts the provider function was only invoked twice total (once for the first call's retry, zero more for the second).

Verified both directions:
```
# against the unfixed patch.py/messages.py (git stash)
FAILED tests/cache/test_cache_integration.py::test_auto_cache_prevents_duplicate_calls_after_a_retry
AssertionError: Second, identical call should hit the cache instead of calling the provider again
assert 3 == 2

# against the fix
tests/cache/test_cache_integration.py::test_auto_cache_prevents_duplicate_calls_after_a_retry PASSED
```

Also ran the broader cache/retry suites to check for regressions, all green:
```
tests/cache/ tests/coverage/test_cache_coverage.py tests/coverage/test_core_patch_retry_coverage.py tests/v2/test_retry_runtime.py
55 passed
```

Fixes #2454

- [ ] If it is a core feature, I have added documentation. (not applicable — internal bug fix, no user-facing API or behavior change to document)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## [Unreleased]
 
 ### Fixed
+- **v2 caching**: Isolate retry-bound `kwargs` so reask-handler mutations to `messages` cannot leak back into the caller's `new_kwargs` and cause the cache lookup/store keys to diverge after any retry. Without this fix, `cache=` silently stops caching any request that needed at least one reask. ([#2454](https://github.com/567-labs/instructor/issues/2454))
 - **v2 message handling**: Preserve caller-owned message lists and nested content across request preparation and retries for OpenAI-compatible, Cohere, Mistral, OpenRouter, Writer, and xAI handlers. ([#2417](https://github.com/567-labs/instructor/issues/2417), [#2428](https://github.com/567-labs/instructor/issues/2428))
 - **v2 JSON extraction**: Prefer the final complete top-level JSON value in text responses and retain every JSON object when multiple objects arrive in one streaming chunk.
 - **v2 schemas**: Treat fields with Pydantic `default_factory` values as optional in generated OpenAI tool schemas.
--- a/instructor/v2/core/messages.py
+++ b/instructor/v2/core/messages.py
@@ -38,6 +38,23 @@ def copy_messages_for_mutation(messages: list[dict[str, Any]]) -> list[dict[str,
     return copied
 
 
+def isolate_retry_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
+    """Return kwargs with a shallow copy of the messages/contents/chat_history list.
+
+    Reask handlers append/extend that list in place during the retry loop (see
+    `copy_messages_for_mutation` above for the equivalent problem one layer up, in
+    `prepare_request`). Passing the caller-facing kwargs dict straight into the retry
+    loop means those mutations land on the exact same list object the caller passed
+    in, which corrupts any code that reads it again afterward, e.g. computing a cache
+    key from it both before and after the retry loop runs.
+    """
+    for key_name in ("messages", "contents", "chat_history"):
+        value = kwargs.get(key_name)
+        if isinstance(value, list):
+            return {**kwargs, key_name: list(value)}
+    return kwargs
+
+
 def dump_message(message: ChatCompletionMessage) -> ChatCompletionMessageParam:
     ret: ChatCompletionMessageParam = {
         "role": message.role,
--- a/instructor/v2/core/messages.py
+++ b/instructor/v2/core/messages.py
@@ -47,12 +47,18 @@ def isolate_retry_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
     loop means those mutations land on the exact same list object the caller passed
     in, which corrupts any code that reads it again afterward, e.g. computing a cache
     key from it both before and after the retry loop runs.
+
+    Each of the three candidate keys is isolated independently rather than stopping
+    at the first match, so a kwargs dict carrying more than one of them (not used by
+    any current provider, but not precluded by the shape of kwargs either) doesn't
+    leave the others aliased to the caller's list.
     """
+    isolated = dict(kwargs)
     for key_name in ("messages", "contents", "chat_history"):
-        value = kwargs.get(key_name)
+        value = isolated.get(key_name)
         if isinstance(value, list):
-            return {**kwargs, key_name: list(value)}
-    return kwargs
+            isolated[key_name] = list(value)
+    return isolated
 
 
 def dump_message(message: ChatCompletionMessage) -> ChatCompletionMessageParam:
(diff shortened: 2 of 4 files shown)


jxnl:

PR #2460, opened 2026-07-18
fix: prevent CitationMixin crash when context missing 'context' key

### LINKED ISSUE #2459: CitationMixin crashes when validation_context missing 'context' key
## Bug

## Reproduction

```python

class User(BaseModel):

# This crashes:
User.model_validate(
)
```

## Expected

## Actual

## Root Cause

`citation.py:70`:

```python
```

## Fix

truecallerabreham:
## Summary

Fix `CitationMixin.validate_sources` crash when `validation_context` is provided but doesn't contain the expected `"context"` key.

## Problem

`validate_sources` calls `info.context.get("context", None)` which returns `None` if the key is missing. This `None` is passed to `get_spans(None)`, which calls `_get_span(quote, None)`, which calls `regex.search(pattern, None)` — raising `TypeError: expected string or bytes-like object`.

## Fix

Add `if text_chunks is None: return self` after the `.get()` call (1 line).

## Test

Added `tests/dsl/test_citation.py` with 3 regression tests covering:
- Missing `"context"` key in validation_context
- None context
- Normal operation with valid context

## Issue

Closes #2459

### DIFF
--- a/instructor/v2/dsl/citation.py
+++ b/instructor/v2/dsl/citation.py
@@ -69,6 +69,9 @@ class CitationMixin(BaseModel):
         # Get the context from the info
         text_chunks = info.context.get("context", None)
 
+        if text_chunks is None:
+            return self
+
         # Get the spans of the substring_phrase in the context
         spans = list(self.get_spans(text_chunks))
         # Replace the substring_phrase with the actual substring
--- a/tests/dsl/test_citation.py
+++ b/tests/dsl/test_citation.py
@@ -1,0 +1,43 @@
+"""Regression test for CitationMixin crash when context missing 'context' key.
+
+See: https://github.com/567-labs/instructor/issues/2459
+"""
+import pytest
+from pydantic import BaseModel, Field
+
+from instructor import CitationMixin
+
+
+class CitationModel(CitationMixin, BaseModel):
+    answer: str = Field(description="The answer")
+    substring_quotes: list[str] = Field(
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Correcting my own review on #2492: I described that as one of two PRs for #2459 when there are actually four, and this one predates all of them by eight days. My duplicate check was too shallow, apologies.

The fix here is right, and identical in substance to #2488 and #2492: `info.context.get("context", None)` returns `None` when the validation context exists but has no `context` key, and both downstream uses assume a string:

```
```

with `text_chunks[span[0]:span[1]]` failing on `None` right after. Returning early is the correct shape, and it matches the `info.context is None` guard already two lines up.

Comparing the four for whoever triages:

| PR | date | test | notes |
|---|---|---|---|
| #2460 (this) | Jul 18 | yes, new `tests/dsl/test_citation.py` | earliest |
| #2471 | Jul 20 | - | closed |
| #2488 | Jul 26 | yes, in existing coverage file | adds CHANGELOG entry |
| #2492 | Jul 27 | no | |

This one has priority by date and ships a test. The only thing #2488 carries that this doesn't is the CHANGELOG line, which is trivial to add here. I'd land this one.

jxnl:
Closed as a duplicate implementation. The citation-context fix and regression coverage shipped in #2495 via the consolidated canonical patch.

PR #2462, opened 2026-07-18
fix: add threading lock for _processing_models set in partial.py

### LINKED ISSUE #2461: Thread-unsafe _processing_models set in partial.py
## Bug

The module-level `_processing_models: set[type]` in `partial.py` is mutated without locking in `_process_generic_arg` (lines 272-282) and `__class_getitem__` (lines 639-648). In concurrent async/threaded use (common with instructor), two threads processing different models can corrupt the set, leading to `RuntimeError: set changed size during iteration` or missed recursion guards.

## Reproduction

Not easily reproducible deterministically, but the race condition exists whenever:
1. Two concurrent async tasks or threads both call `Partial[ModelA]` and `Partial[ModelB]`
2. Both models have nested BaseModel fields that trigger `_process_generic_arg`
3. Both threads mutate `_processing_models` simultaneously

## Impact

- `RuntimeError: set changed size during iteration` in the worst case
- Missed recursion guards (self-referential models may infinite-recurse) in the worst case
- Silent data corruption in partial model generation

## Fix

Add a `threading.Lock()` to guard mutations of `_processing_models`. The lock should be held around the `add`/`discard` pair in both `_process_generic_arg` and `__class_getitem__`.

## Notes

This is a pre-existing race condition. The set is used as a recursion guard for self-referential models (e.g., `TreeNode` with `children: List["TreeNode"]`). Without locking, concurrent requests can interfere with each other's recursion detection.

truecallerabreham:
## Summary

Add a `threading.Lock()` to guard mutations of the shared `_processing_models` set in `partial.py`.

## Problem

The module-level `_processing_models: set[type]` is mutated without locking in `_process_generic_arg` (lines 272-282) and `__class_getitem__` (lines 639-648). In concurrent async/threaded use (common with instructor), two threads processing different models can corrupt the set, leading to `RuntimeError: set changed size during iteration` or missed recursion guards.

## Fix

Add `import threading` and `_processing_models_lock = threading.Lock()`. Wrap the `add`/`discard` mutations in both `_process_generic_arg` and `__class_getitem__` with `with _processing_models_lock:`.

The lock is held only around the check-and-modify pattern, not around the recursive call inside the `try` block, to avoid deadlocks.

## Issue

Closes #2461

### DIFF
--- a/instructor/v2/dsl/citation.py
+++ b/instructor/v2/dsl/citation.py
@@ -69,6 +69,9 @@ class CitationMixin(BaseModel):
         # Get the context from the info
         text_chunks = info.context.get("context", None)
 
+        if text_chunks is None:
+            return self
+
         # Get the spans of the substring_phrase in the context
         spans = list(self.get_spans(text_chunks))
         # Replace the substring_phrase with the actual substring
--- a/instructor/v2/dsl/partial.py
+++ b/instructor/v2/dsl/partial.py
@@ -9,6 +9,7 @@
 from __future__ import annotations
 
 import re
+import threading
 import types
 import warnings
 from collections.abc import AsyncGenerator, Callable, Generator, Iterable
@@ -42,6 +43,7 @@ UNION_ORIGINS = (Union, UNION_TYPE) if UNION_TYPE is not None else (Union,)
 # Track models currently being processed to prevent infinite recursion
 # with self-referential models (e.g., TreeNode with children: List["TreeNode"])
 _processing_models: set[type] = set()
+_processing_models_lock = threading.Lock()
 
 
 def _unwrap_optional_base_model(annotation: Any) -> type[BaseModel] | None:
@@ -269,9 +271,10 @@ def _process_generic_arg(
         return arg_origin[modified_nested_args]
     if isinstance(arg, type) and issubclass(arg, BaseModel):
         # Prevent infinite recursion for self-referential models
-        if arg in _processing_models:
-            return arg  # Already processing this model, return unwrapped
-        _processing_models.add(arg)
+        with _processing_models_lock:
+            if arg in _processing_models:
+                return arg  # Already processing this model, return unwrapped
+            _processing_models.add(arg)
         try:
             return (
                 _make_partial_type(arg, make_fields_optional=True)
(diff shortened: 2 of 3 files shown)


ErenAta16 (commented):
Two problems here, and the second one is a regression rather than a scoping question.

**1. A lock doesn't fix the bug in #2461.**

The reported failure isn't set corruption, it's a logical race on shared guard state. Serializing access to a still-global set keeps the sharing intact:

```
thread A: acquires lock, Address not in set, adds it, releases lock, recurses into Address
thread B: acquires lock, sees Address IS in set -> returns raw unwrapped Address
```

Thread B still gets the wrong schema. The lock only prevents two threads mutating the set at the same instant, and that was never the observable failure. There's no `for x in _processing_models` anywhere in this file, only `in` / `.add` / `.discard`, which are individually atomic under the GIL, so `RuntimeError: set changed size during iteration` was never reachable either. The guard needs to be per-call-stack, not shared-with-a-mutex. That's what #2489 does with a `ContextVar`.

**2. The refactor moves `try` outside the `else` and disables the recursion guard.**

In the `__class_getitem__` hunk:

```python
with _processing_models_lock:
    if annotation in _processing_models:
        tmp_field.annotation = annotation      # guard hit: keep unwrapped
    else:
        _processing_models.add(annotation)
try:
    tmp_field.annotation = Partial[annotation]  # runs unconditionally now
finally:
    with _processing_models_lock:
        _processing_models.discard(annotation)  # discards even if this call never added
```

Originally the `try` lived inside the `else`. Now it runs on both branches, so the guard-hit path immediately overwrites `tmp_field.annotation` with `Partial[annotation]` and recurses anyway. Simulating both control flows on a self-referential model:

```
ORIGINAL   -> ok, max recursion depth 1
PR #2462   -> guard failed: runaway recursion
```

So a `TreeNode`-style model with `children: list["TreeNode"]` would recurse until it blows the stack. The `finally` is also now unbalanced: it discards the annotation even on the path that didn't add it, stomping the outer frame's guard entry.

**3. Scope:** this also carries the `citation.py` fix from #2459, which is a separate issue with its own PRs (#2460, #2488, #2492). Worth splitting so each can be judged on its own.

I'd close this in favour of #2489 for the partial.py half and one of the citation PRs for the other. Not a criticism of the instinct, a lock is the obvious first reach for anything labelled "thread-unsafe", it just isn't what this particular bug needs.

jxnl:
Superseded by #2495, which isolates the partial-model recursion guard with ContextVar state rather than sharing a process-wide lock.

PR #2463, opened 2026-07-19
fix(citations): honor the errs tolerance cap in CitationMixin fuzzy span matching

lntutor:
### What's broken

`CitationMixin._get_span` escalates the allowed regex edit distance in a loop whose guard is evaluated **before** the increment:

```python
errs_ = 0
s = regex.search(f"({minor}){{e<={errs_}}}", major)
while s is None and errs_ <= errs:   # errs defaults to 5
    errs_ += 1
    s = regex.search(f"({minor}){{e<={errs_}}}", major)
```

When `errs_ == errs` (5) the body runs once more, bumping `errs_` to 6 and searching with `{e<=6}`. So the effective fuzzy tolerance is `errs + 1 = 6` — a hallucinated `substring_quote` that is 6 edits from any real source span is accepted and silently rewritten to that unrelated context substring, instead of being dropped as "not found":

```python
class Answer(CitationMixin): pass
Answer.model_validate({"substring_quotes": ["0123ZZZZZZ"]}, context={"context": "0123456789"}).substring_quotes
# before: ['0123456789']   (6 edits away -- should be [])
```

This weakens the citation verification the mixin exists to provide.

### Fix

Change the guard to `errs_ < errs` so the final search runs at exactly `{e<=errs}`. Quotes within `errs` edits still match.

### Tests

Adds `tests/dsl/test_citation.py` (within-tolerance matches; beyond-tolerance dropped). The existing citation coverage in `tests/coverage/test_dsl_small_coverage.py` — which relies on a legitimate 1-edit match — stays green (16). Distinct from #2430 (regex-escape crash), which leaves the loop bound untouched.

### DIFF
--- a/instructor/v2/dsl/citation.py
+++ b/instructor/v2/dsl/citation.py
@@ -85,7 +85,7 @@ class CitationMixin(BaseModel):
 
         errs_ = 0
         s = regex.search(f"({minor}){{e<={errs_}}}", major)
-        while s is None and errs_ <= errs:
+        while s is None and errs_ < errs:
             errs_ += 1
             s = regex.search(f"({minor}){{e<={errs_}}}", major)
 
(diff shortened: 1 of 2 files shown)


jxnl:

PR #2466, opened 2026-07-19
fix: preserve empty-content function calls

### LINKED ISSUE #2464: dump_message drops legacy function calls when content is empty
## Bug

## Reproduction

```python

    role="assistant",
    content=None,
)

print(dump_message(message))
```

Current output:

```python
```

## Impact

lntutor:
## Describe your changes

Preserve legacy OpenAI `function_call` payloads when an assistant response has `None` or empty content. Previously, `dump_message` skipped serialization behind a truthy-content guard and retry messages silently lost the function call.

Adds a focused regression test and an Unreleased changelog entry.

Closes #2464

## Validation

- `uv run pytest -q tests/v2/test_messages.py`
- `uv run ruff check instructor/v2/core/messages.py tests/v2/test_messages.py`
- `uv run ruff format --check instructor/v2/core/messages.py tests/v2/test_messages.py`
- `uv run ty check instructor/v2/core/messages.py tests/v2/test_messages.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## [Unreleased]
 
 ### Fixed
+- **v2 message handling**: Preserve legacy OpenAI `function_call` payloads when assistant content is empty. ([#2464](https://github.com/567-labs/instructor/issues/2464))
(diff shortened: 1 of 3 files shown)


jxnl:
Closed as a duplicate implementation. Empty-content legacy function calls are preserved by the implementation shipped in #2495.

PR #2469, opened 2026-07-20
fix(gemini): treat caller kwargs as read-only in update_gemini_kwargs

### LINKED ISSUE #2465: update_gemini_kwargs mutates caller generation config
## Bug

`instructor.v2.providers.gemini.utils.update_gemini_kwargs()` shallow-copies the request dictionary but mutates the nested `generation_config` dictionary in place while translating OpenAI-style keys.

## Reproduction

```python
from instructor.v2.providers.gemini import utils

utils._default_safety_thresholds = lambda: None
generation_config = {"max_tokens": 5}

utils.update_gemini_kwargs({"generation_config": generation_config})
print(generation_config)
```

Current output:

```python
{"max_output_tokens": 5}
```

Expected:

```python
{"max_tokens": 5}
```

The function should treat caller-owned configuration as read-only and return a normalized copy.

## Impact

Reusing the same options dictionary for a retry or later request silently changes the accepted parameter names and can produce behavior that depends on call order.

feiiiiii5:
## What

`update_gemini_kwargs` no longer mutates the caller's nested `generation_config` and `safety_settings` dicts while translating OpenAI-style keys to Gemini names.

## Why

Fixes #2465. `update_gemini_kwargs` did `result = kwargs.copy()` (shallow), then mutated the nested `generation_config` via `gen_config.pop(openai_key)` / `gen_config[gemini_key] = val` — the nested dict was still the caller's reference. The same leak applied to `safety_settings` (the defaults-fill loop wrote back through the caller's dict). Reusing the same options dict for a retry or later request silently changed accepted parameter names and produced call-order-dependent behavior.

This is the same shallow-copy mutation leak class as #2417 (`messages=` aliasing).

## Approach

Shallow-copy the two nested dicts that `update_gemini_kwargs` mutates, immediately after pulling them off `result`:

```python
gen_config = {**result["generation_config"]}
result["generation_config"] = gen_config
# ... existing rename loop now mutates the copy, not the caller's dict
```

```python
safety_settings = {**result.get("safety_settings", {})}
result["safety_settings"] = safety_settings
# ... existing defaults-fill loop now mutates the copy
```

### Alternatives considered

- `copy.deepcopy(kwargs)` at the top — too broad: `kwargs` may contain Pydantic models, message lists, arbitrary user objects that don't need copying. Adds O(n) cost on every call. Rejected.
- Wrapping every dict operation in a "deep clone everything" helper — same cost issue, and introduces a new abstraction for a single-use purpose.
- Returning a new `generation_config` dict without writing it back to `result` — would require rewriting every subsequent consumer to read from the return value or a new local. More invasive change; current consumers already read `result["generation_config"]`.

## Testing

- New regression tests in `tests/v2/test_gemini_utils_deterministic.py` (5 added, all named for #2465):
  - `test_update_gemini_kwargs_does_not_mutate_caller_generation_config` — snapshots the caller's `generation_config`, asserts it is unchanged after the call.
  - `test_update_gemini_kwargs_does_not_mutate_caller_safety_settings` — same class, for `safety_settings`.
  - `test_update_gemini_kwargs_returns_result_with_translated_keys` — guards against an over-broad fix that breaks the translation itself.
  - `test_update_gemini_kwargs_handles_missing_generation_config` — edge case: no `generation_config` key.
  - `test_update_gemini_kwargs_handles_empty_generation_config` — edge case: `generation_config={}`.
- Full local run: `tests/v2/test_gemini_utils_deterministic.py` 23/23 ✅, `tests/test_genai_config_merging.py` 21/21 ✅, `tests/v2/test_messages_not_mutated.py` (excluding `mistralai`-blocked cases, pre-existing on main) 61/61 ✅.
- `ruff format` + `ruff check` clean.

### DIFF
--- a/instructor/v2/providers/gemini/utils.py
+++ b/instructor/v2/providers/gemini/utils.py
@@ -322,7 +322,12 @@ def update_gemini_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
     result = kwargs.copy()
 
     if "generation_config" in result:
-        gen_config = result["generation_config"]
+        # Shallow-copy the nested dict so the caller's `generation_config`
+        # is treated as read-only: `dict.copy()` only copies the outer dict,
+        # so mutating the nested one leaks back to the caller and silently
+        # changes accepted parameter names on retry (#2465).
+        gen_config = {**result["generation_config"]}
+        result["generation_config"] = gen_config
 
         for openai_key, gemini_key in _OPENAI_TO_GEMINI_MAP.items():
             if openai_key in gen_config:
@@ -338,7 +343,10 @@ def update_gemini_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
         result.setdefault("safety_settings", {})
         return result
 
-    safety_settings = result.get("safety_settings", {})
+    # Same shallow-copy defense for `safety_settings`: the defaults-fill loop
+    # below writes back through this dict reference, so copy it first to keep
+    # the caller's object unchanged (same bug class as `generation_config`).
+    safety_settings = {**result.get("safety_settings", {})}
     result["safety_settings"] = safety_settings
 
     for category, threshold in default_safety_thresholds.items():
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Closing this — my sibling PR #2457 already covers the kwargs-isolation issue and has had positive feedback, so this one is redundant while I trim my open PRs. The branch stays on my fork if it's worth revisiting on its own. Thanks!

PR #2470, opened 2026-07-20
fix: stop mutating generation_config; keep empty-content function_call

### LINKED ISSUE #2464: dump_message drops legacy function calls when content is empty

### LINKED ISSUE #2465: update_gemini_kwargs mutates caller generation config

Solaris-star:
## Summary
Two small, independent correctness bugs in instructor v2:

### #2465 — `update_gemini_kwargs` mutates caller `generation_config`
`kwargs.copy()` is shallow, so rewriting `max_tokens` → `max_output_tokens` mutated the caller's nested dict. Now the nested dict is copied first.

### #2464 — `dump_message` drops legacy `function_call` when content is empty
Tool/function responses often use `content=None`. The previous `and ret["content"]` guard skipped serialization entirely. Function calls are now always appended (empty string base when content is missing).

## Test plan
- [x] Repro scripts from both issues pass locally
- [x] Added `tests/v2/test_issue_2465_2464.py`

Fixes #2465
Fixes #2464

### DIFF
--- a/instructor/v2/core/messages.py
+++ b/instructor/v2/core/messages.py
@@ -45,11 +45,9 @@ def dump_message(message: ChatCompletionMessage) -> ChatCompletionMessageParam:
     }
     if hasattr(message, "tool_calls") and message.tool_calls is not None:
         ret["tool_calls"] = message.model_dump()["tool_calls"]
-    if (
-        hasattr(message, "function_call")
-        and message.function_call is not None
-        and ret["content"]
-    ):
+    if hasattr(message, "function_call") and message.function_call is not None:
+        # Preserve legacy function calls even when content is None/empty
+        # (common for pure tool/function responses) — see #2464.
         if not isinstance(ret["content"], str):
             response_message = ""
             for content_message in ret["content"]:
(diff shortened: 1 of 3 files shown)


Solaris-star:
Closing this to avoid duplicating an existing fix for the same issue. The earlier contributor PR already covers this bug, so keeping another implementation open would only add review noise. Thanks.

PR #2471, opened 2026-07-20
fix(citation): tolerate validation_context without a 'context' key

### LINKED ISSUE #2459: CitationMixin crashes when validation_context missing 'context' key

Solaris-star:
## Summary
`CitationMixin.validate_sources` crashed with:

```
TypeError: expected string or bytes-like object
```

when `validation_context` was provided but did not contain a `"context"` key (or contained a non-string value). `info.context.get("context", None)` returned `None`, which was passed into `regex.search`.

## Change
If the resolved context is not a `str`, return the model unchanged instead of attempting span resolution.

## Test plan
- [x] Added `tests/v2/test_citation_missing_context.py`
- Manual: repro from the issue no longer raises

Fixes #2459.

### DIFF
--- a/instructor/v2/dsl/citation.py
+++ b/instructor/v2/dsl/citation.py
@@ -66,8 +66,12 @@ class CitationMixin(BaseModel):
         if info.context is None:
             return self
 
-        # Get the context from the info
+        # Get the context from the info. Callers may pass a validation_context
+        # dict without a "context" key (or with a non-string value); treat that
+        # as "no source text" rather than crashing in regex.search (#2459).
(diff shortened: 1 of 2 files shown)


Solaris-star:
Closing this to avoid duplicating an existing fix for the same issue. The earlier contributor PR already covers this bug, so keeping another implementation open would only add review noise. Thanks.

PR #2473, opened 2026-07-20
fix(batch): accept empty structured results

hsusul:
## Summary

Allow valid empty structured objects in batch results to reach Pydantic validation.

## Problem

`BatchProcessor.parse_results()` treats an extracted `{}` response as an
extraction failure. Models whose fields have defaults therefore receive a
`BatchError` instead of a successfully validated result.

This affects both OpenAI JSON-schema batch responses and Anthropic tool-use
batch responses.

## Root cause

The parser used a truthiness check for `extracted_data`, conflating an empty
dictionary with the `None` sentinel returned when no structured result could
be extracted.

## Fix

Check explicitly for `None`. Empty dictionaries are now validated normally,
while absent and malformed results retain the existing error behavior.

## Test plan

- Added a regression covering an OpenAI batch response containing `{}`
- Added the equivalent Anthropic tool-use response
- Verified both fail before the fix and pass afterward
- Ran the focused batch suite, broad offline suite, Ruff, ty, and pre-commit

## Compatibility considerations

This is backwards-compatible. Only valid empty mappings change behavior;
missing or malformed results continue to produce `BatchError`.

N/A — independently discovered; no matching issue found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## [Unreleased]
 
 ### Fixed
+- **Batch results**: Accept empty structured objects so Pydantic models can apply field defaults for OpenAI and Anthropic batch responses.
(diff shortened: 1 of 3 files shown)


jxnl:

PR #2476, opened 2026-07-23
fix: build partial model instances for incomplete list[BaseModel] items during streaming

manjunathbhaskar:
### Summary
Fixes a type asymmetry in `_build_partial_list` where the trailing (still-streaming)
item in a `list[BaseModel]` field was appended as a raw `dict` instead of a partial
model instance.

### Problem
When the root JSON is incomplete during streaming, `_build_partial_object` correctly
builds a partial model instance (via `model_construct`) for a singular open nested
`BaseModel` field. `_build_partial_list` does the same for *complete* list items, but
for the *incomplete* trailing item it fell through to `result.append(item)`, leaving
a raw dict.

This breaks the type contract users rely on when iterating a streamed list
(`for item in partial.items: item.name`) -- the last item raises `AttributeError`
until it completes.

### Fix
Route incomplete list items through the same `_build_partial_object` call already
used for singular open nested fields, when an item model is known. 6 lines, no new
abstractions.

### Testing
- Added `test_partial_list_returns_model_instance_for_incomplete_trailing_item`,
  confirmed to fail on `main` and pass with this change.
- Updated one pre-existing test in `tests/coverage/test_dsl_partial_coverage.py`
  that asserted the old (inconsistent) raw-dict behavior for open list items --
  its own sibling test in the same file already established that singular open
  nested fields recurse into partial instances, so this aligns list items with
  that existing, tested contract.
- Full `tests/dsl/test_partial.py` + `tests/coverage/test_dsl_partial_coverage.py`
  suite: 65 passed, 2 skipped (pre-existing), 0 failed.

### DIFF
--- a/instructor/v2/dsl/partial.py
+++ b/instructor/v2/dsl/partial.py
@@ -241,6 +241,12 @@ def _build_partial_list(
             result.append(_item_model.model_validate(item, **kwargs))
             continue
 
+        if _item_model is not None and isinstance(item, dict):
+            result.append(
+                _build_partial_object(item, _item_model, tracker, item_path, **kwargs)
+            )
+            continue
+
         result.append(item)
 
     return result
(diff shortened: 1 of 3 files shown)


jxnl:

PR #2478, opened 2026-07-23
fix(anthropic): disable parallel tool use for forced single-tool requests

### LINKED ISSUE #2477: bug: Anthropic does not disable parallel tool calling

**What Model are you using?**

- [ ] gpt-3.5-turbo
- [ ] gpt-4-turbo
- [ ] gpt-4
- [x] Other (please specify)

Claude via Anthropic / Anthropic Bedrock (`instructor.from_anthropic(...)`, mode `ANTHROPIC_TOOLS`). Reproduced against Claude models that support tool use (e.g. Claude Sonnet 4 family).

**instructor version:** `1.15.4`

**Describe the bug**

In `Mode.ANTHROPIC_TOOLS` with a single `response_model`, instructor forces the schema via a forced `tool_choice`, but does **not** set `disable_parallel_tool_use: True`. As a result, the model is still free to emit **multiple** `tool_use` blocks in one response, which the parser then rejects — so a valid model response fails at parse time.

**Request side** — `instructor/v2/providers/anthropic/handlers.py:411`:

```python
new_kwargs["tool_choice"] = {
    "type": "tool",
    "name": getattr(response_model, "__name__", "response"),
}
# no disable_parallel_tool_use
```

**Parse side** — `instructor/v2/providers/anthropic/handlers.py:534`:

```python
tool_calls_validator = TypeAdapter(
    Annotated[list[Any], Field(min_length=1, max_length=1)]
)
tool_call = tool_calls_validator.validate_python(tool_calls)[0]
```

When `completion.content` contains more than one `type="tool_use"` block, this raises a Pydantic `ValidationError` (`too_long` / `max_length=1`).

The contradiction: for single-schema `ANTHROPIC_TOOLS`, the parser hard-requires **exactly one** `tool_use`, but the request never tells Anthropic to produce only one. Under ambiguous inputs (schema wants one value, document contains several candidates), Claude emits one `tool_use` per candidate and instructor fails — even though every tool payload is individually valid against the schema.

> Note: `instructor.processing.function_calls.parse_anthropic_tools` still exists but is now a **deprecated shim** (`v2/core/function_calls.py:271`) that delegates to the registry handler above; the real logic lives in the v2 Anthropic handler.

**To Reproduce**

1. Deterministic parse failure — simplified mirror of the validator logic in `handlers.py:529-537` (no LLM needed):

```python
import json
from typing import Any, Annotated
from pydantic import Field, TypeAdapter, ValidationError

def parse_like_instructor(response_content):
    tool_calls = [
        json.dumps(c["input"]) for c in response_content if c["type"] == "tool_use"
    ]
    validator = TypeAdapter(Annotated[list[Any], Field(min_length=1, max_length=1)])
    return validator.validate_python(tool_calls)[0]

# Ambiguous extraction: model emitted one tool_use per candidate value
content = [
    {"type": "tool_use", "id": "toolu_1", "name": "SSNExtract", "input": {"ssn": "123-45-6789"}},
    {"type": "tool_use", "id": "toolu_2", "name": "SSNExtract", "input": {"ssn": "987-65-4321"}},
]

parse_like_instructor(content)
# pydantic.ValidationError: List should have at most 1 item after validation, not 2 [type=too_long]
```

2. Live / flaky path (Anthropic API):

```python
import asyncio
from anthropic import AsyncAnthropic
import instructor

class SSNExtract(BaseModel):
    ssn: str = Field(description="Extract THE Social Security Number. Return exactly one value.")

PROMPT = """
Extract the SSN from this document excerpt.

Form W-2 Wage and Tax Statement
Employee A: John Smith   SSN: 123-45-6789
Employee B: Jane Doe     SSN: 987-65-4321
Spouse (joint return):   SSN: 555-12-3456

There are multiple SSNs. Pick one. Still return via the tool.
""".strip()

async def main():
    client = instructor.from_anthropic(AsyncAnthropic())  # default mode = ANTHROPIC_TOOLS
    resp, completion = await client.chat.completions.create_with_completion(
        model="claude-sonnet-4-20250514",
        max_tokens=1024,
        messages=[{"role": "user", "content": PROMPT}],
        response_model=SSNExtract,
        max_retries=0,
    )
    tool_uses = [b for b in completion.content if b.type == "tool_use"]
    print("tool_use count:", len(tool_uses))
    print(resp)

asyncio.run(main())
```

When Claude returns >1 `tool_use`, instructor fails in the Anthropic tools parser with a `ValidationError` on `max_length=1`. (This path is inherently flaky — whether the model emits 1 or N blocks depends on the input and sampling.)

**Expected behavior**

For single-schema `ANTHROPIC_TOOLS`, instructor should either:

1. **Set `disable_parallel_tool_use: True` on the forced `tool_choice`** (`handlers.py:411`):

```python
new_kwargs["tool_choice"] = {
    "type": "tool",
    "name": getattr(response_model, "__name__", "response"),
    "disable_parallel_tool_use": True,
}
```

so Anthropic returns a single `tool_use`. Since the parser already hard-requires exactly one `tool_use` in this mode, forcing `disable_parallel_tool_use: True` has **no downside** — it simply aligns the request with what the parser already assumes. (This should not be applied to `PARALLEL_TOOLS` mode or when `tool_choice` is `{"type": "auto"}`, e.g. thinking-enabled paths at `handlers.py:399`.)

**or**

2. Handle multiple same-tool `tool_use` blocks gracefully (e.g. take the first, or surface a clearer error) instead of hard-failing on `max_length=1`.

**Workaround**

Two options that work on the caller side without patching instructor:

- **`extra_body` (preferred, no reliance on internals):** instructor forwards `extra_body` to the Anthropic SDK untouched, and the SDK shallow-merges it over the request body (extra_body wins), so it overrides instructor's `tool_choice`:

```python
await client.chat.completions.create_with_completion(
    ...,
    response_model=MySchema,
    extra_body={
        "tool_choice": {
            "type": "tool",
            "name": MySchema.__name__,
            "disable_parallel_tool_use": True,
        }
    },
)
```

- **Monkey-patch** `client.messages.create` to inject `disable_parallel_tool_use: True` after instructor builds `tool_choice`.

**Screenshots**

N/A — failure is a Pydantic `ValidationError` (`list should have at most 1 item` / `max_length=1`) when `completion.content` contains multiple `type="tool_use"` blocks.

mittalpk:
Fixes #2477

## Summary

`AnthropicToolsHandler.prepare_request()` forces `tool_choice` to a specific named tool when a single (non-parallel, non-thinking) `response_model` is used, but never set `disable_parallel_tool_use`. Anthropic can still emit multiple `tool_use` blocks for that same forced tool, which the single-model parser then rejects — so a valid model response fails at parse time, exactly as described in #2477.

## Fix

Added `"disable_parallel_tool_use": True` to the forced `tool_choice` dict in the single-response-model branch only. Left the `is_parallel` (`Iterable[T]`) and `thinking_enabled` branches untouched — those intentionally use `{"type": "auto"}` and, in the parallel case, actually want multiple tool calls.

## Testing

Added `tests/v2/test_anthropic_handlers.py` with 3 unit tests (no API key/mocking needed, following the existing `test_xai_handlers.py` pattern of calling `request_handler` directly):
- forced single-tool `tool_choice` now includes `disable_parallel_tool_use: True`
- the parallel-tools case is unaffected (still `{"type": "auto"}`, no `disable_parallel_tool_use`)
- an explicitly caller-supplied `tool_choice` is still not overridden

Verified the new test fails against the pre-fix code and passes against the fix. Ran the full `tests/v2/` suite (1478 passed, 160 skipped, 105 deselected) — no regressions.

How I verified: `python -m pytest tests/v2/ -k "not mistral"` (mistral cases skipped/deselected only because the optional `mistralai` SDK wasn't installed in my env, unrelated to this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## [Unreleased]
 
 ### Fixed
+- **Anthropic tools**: Set `disable_parallel_tool_use` on the forced `tool_choice` for single-model requests, so Anthropic can no longer emit multiple `tool_use` blocks for the same forced tool and have an otherwise-valid response rejected by the parser. ([#2477](https://github.com/567-labs/instructor/issues/2477))
 - **v2 message handling**: Preserve caller-owned message lists and nested content across request preparation and retries for OpenAI-compatible, Cohere, Mistral, OpenRouter, Writer, and xAI handlers. ([#2417](https://github.com/567-labs/instructor/issues/2417), [#2428](https://github.com/567-labs/instructor/issues/2428))
 - **v2 JSON extraction**: Prefer the final complete top-level JSON value in text responses and retain every JSON object when multiple objects arrive in one streaming chunk.
 - **v2 schemas**: Treat fields with Pydantic `default_factory` values as optional in generated OpenAI tool schemas.
--- a/instructor/v2/providers/anthropic/handlers.py
+++ b/instructor/v2/providers/anthropic/handlers.py
@@ -402,6 +402,11 @@ class AnthropicToolsHandler(AnthropicHandlerBase):
                 new_kwargs["tool_choice"] = {
                     "type": "tool",
                     "name": getattr(response_model, "__name__", "response"),
+                    # Without this, Anthropic may still emit multiple tool_use
+                    # blocks for the same forced tool, which the single-model
+                    # parser then rejects even though the model's response was
+                    # otherwise valid.
+                    "disable_parallel_tool_use": True,
                 }
 
         return response_model, new_kwargs
--- a/tests/v2/test_anthropic_handlers.py
+++ b/tests/v2/test_anthropic_handlers.py
@@ -1,0 +1,72 @@
+"""Unit tests for Anthropic v2 handlers.
+
+These tests verify handler behavior without requiring API keys, by calling
+the request-preparation logic directly and inspecting the resulting kwargs.
+"""
+
+from __future__ import annotations
+
+from typing import Iterable
+
+import pytest
+from pydantic import BaseModel
+
+from instructor import Mode, Provider
+from instructor.v2.core.registry import mode_registry
+
+
+class Answer(BaseModel):
+    """Simple answer model for testing."""
+
+    answer: float
+
+
+@pytest.fixture
+def handler():
+    """Get the Anthropic tools handler from registry."""
+    return mode_registry.get_handlers(Provider.ANTHROPIC, Mode.TOOLS)
+
+
+class TestAnthropicToolsHandlerToolChoice:
+    """Regression tests for #2477: a forced single-tool tool_choice did not
+    set disable_parallel_tool_use, so Anthropic could still emit multiple
+    tool_use blocks for the same forced tool, which the single-model parser
+    then rejected even though the model's response was otherwise valid."""
+
+    def test_single_response_model_disables_parallel_tool_use(self, handler):
+        kwargs = {
+            "messages": [{"role": "user", "content": "What is 2+2?"}],
+            "max_tokens": 100,
+        }
+
+        _, result_kwargs = handler.request_handler(Answer, kwargs)
+
+        assert result_kwargs["tool_choice"] == {
+            "type": "tool",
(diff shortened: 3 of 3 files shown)


jxnl:

PR #2484, opened 2026-07-25
fix(v2): retain legacy function_call in dump_message when content is empty

### LINKED ISSUE #2464: dump_message drops legacy function calls when content is empty

cognis-digital:
### Problem
`dump_message` (`instructor/v2/core/messages.py`) only appended the legacy OpenAI `function_call` when the assistant message had truthy content. Tool/function-call responses commonly use `content=None`, so the serialized message dropped the function call entirely.

Because `dump_message` builds the assistant turn on the retry/reask path (OpenAI and Mistral handlers), a re-ask could omit the exact structured output that caused the validation failure, leaving the next request without the relevant assistant response.

Fixes #2464.

### Repro (before)
```python

msg = ChatCompletionMessage(role="assistant", content=None,
                            function_call=FunctionCall(name="lookup", arguments='{"id":7}'))
print(dump_message(msg))
# {'role': 'assistant', 'content': ''}   # function_call lost
```

### Fix
Remove the falsy-content guard so a present `function_call` is always serialized. Behavior for non-empty string content and for list content (text/refusal flattening) is unchanged.

### Tests
Added `test_dump_message_retains_function_call_when_content_is_empty`, which fails on `main` and passes with this change. Existing `dump_message` tests and the surrounding message/handler suites remain green. CHANGELOG updated under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## [Unreleased]
 
 ### Fixed
+- **v2 message handling**: Retain legacy `function_call` payloads in `dump_message` when the assistant message has empty or `None` content, so retry/reask flows no longer drop the model's prior function call. ([#2464](https://github.com/567-labs/instructor/issues/2464))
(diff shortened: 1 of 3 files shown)


jxnl:

PR #2486, opened 2026-07-25
fix(anthropic): return tool_result for every tool_use on reask

### LINKED ISSUE #2485: fix(anthropic): handle_reask drops tool_result for all but the last tool_use block in PARALLEL_TOOLS mode
## Bug: `handle_reask` drops `tool_result` for all but the last `tool_use` block in `PARALLEL_TOOLS` mode

**File:** `instructor/v2/providers/anthropic/handlers.py` — `AnthropicToolsHandler.handle_reask` (also used by `AnthropicParallelToolsHandler`)

### What's happening
When Anthropic returns **multiple** `tool_use` blocks (normal for `Mode.PARALLEL_TOOLS`) and a validation error triggers a reask, only the **last** `tool_use_id` gets a `tool_result`:

```python
tool_use_id = None
for content in response.content:
    if content.type == "tool_use":
        tool_use_id = content.id   # overwritten each time — only last id kept
```

### Why it breaks
Anthropic requires a `tool_result` for **every** `tool_use` block in the prior turn. Missing one makes the retry request itself fail:400 invalid_request_error: tool_use ids were found without tool_result blocks
immediately after: toolu_1. Each tool_use block must have a corresponding
tool_result block in the next message.

So instead of retrying, `max_retries` just throws a confusing API error.

### Repro
```python
response.content = [
    ToolUseBlock(type="tool_use", id="toolu_1", name="User", input={"name": ""}),
    ToolUseBlock(type="tool_use", id="toolu_2", name="Job", input={"title": "Engineer"}),
]
handler.handle_reask({"messages": []}, response, ValueError("bad name"))
# -> only one tool_result (toolu_2), toolu_1 is missing
```

### Note
The existing test `test_parallel_tools_prepare_reask_and_parse_real_tool_blocks` builds a response with 3 `tool_use` blocks but only asserts the last id gets a `tool_result` — it locks in the bug instead of catching it. Needs updating to assert all ids are present.

cc / mentions: @jxnl @ivanleomk

Solaris-star:
## Summary

When Anthropic returns multiple `tool_use` blocks (normal for `Mode.PARALLEL_TOOLS`) and a validation error triggers a reask, `handle_reask` only kept the **last** `tool_use_id`. Anthropic requires a `tool_result` for *every* prior `tool_use`, so the retry request itself failed with:

```
400 invalid_request_error: tool_use ids were found without tool_result blocks
```

## Fix

Collect all `tool_use` ids and emit one `tool_result` per id (same error content, `is_error=True`).

## Test

```text
python -m pytest tests/coverage/test_anthropic_handlers_coverage.py -q
# 24 passed
```

Regression assertion now checks all three `tool_use_id`s are present.

Fixes #2485

### DIFF
--- a/instructor/v2/providers/anthropic/handlers.py
+++ b/instructor/v2/providers/anthropic/handlers.py
@@ -426,7 +426,10 @@ class AnthropicToolsHandler(AnthropicHandlerBase):
             return kwargs
 
         assistant_content = []
-        tool_use_id = None
+        # Anthropic requires a tool_result for *every* tool_use in the prior turn.
+        # Collect all ids — overwriting a single id drops results for PARALLEL_TOOLS
+        # and makes the reask request fail with 400 (issue #2485).
+        tool_use_ids: list[str] = []
         for content in response.content:
             try:
                 dumped_content = content.model_dump(exclude_none=True)  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


AJ-ing:
Looks correct — collecting every `tool_use` id and asserting all `tool_result` blocks is exactly what Anthropic requires for `PARALLEL_TOOLS` reasks. Thanks for also updating the test that previously locked in last-id-only behavior.

jxnl:

PR #2488, opened 2026-07-26
fix(dsl): handle citation contexts without source text

lntutor:
## What\n\nPrevents CitationMixin validation from crashing when validation_context is present but does not include the expected context source text.\n\nCloses #2459.\n\n## Changes\n\n- Return citations unchanged when validation context has no context key.\n- Add a regression test for a non-empty citation list and missing source text.\n- Document the fix in the changelog.\n\n## Testing\n\n- uv run pytest tests/coverage/test_dsl_small_coverage.py\n- uv run ruff check instructor/v2/dsl/citation.py tests/coverage/test_dsl_small_coverage.py\n- uv run ruff format --check instructor/v2/dsl/citation.py tests/coverage/test_dsl_small_coverage.py

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ## [Unreleased]
 
 ### Fixed
+- **Citation validation**: Preserve LLM-provided citations when a validation context does not include source text, instead of raising a `TypeError`.
 - **v2 message handling**: Preserve caller-owned message lists and nested content across request preparation and retries for OpenAI-compatible, Cohere, Mistral, OpenRouter, Writer, and xAI handlers. ([#2417](https://github.com/567-labs/instructor/issues/2417), [#2428](https://github.com/567-labs/instructor/issues/2428))
 - **v2 JSON extraction**: Prefer the final complete top-level JSON value in text responses and retain every JSON object when multiple objects arrive in one streaming chunk.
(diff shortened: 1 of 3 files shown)


ErenAta16 (commented):
Correct fix, and in the right place.

Worth spelling out why the guard is needed at all given there's already one two lines up: `if info.context is None: return self` covers "no validation context at all", but not "context supplied without a `context` key". In that second case `text_chunks` comes back `None` from the `.get()` and flows into `get_spans` -> `_get_span` -> `regex.search(pattern, None)`:

```
```

and even if that were survived, `text_chunks[span[0]:span[1]]` on the next line would fail on `None` too. So both downstream uses assume a string, and returning early is the right call rather than patching either one.

Returning `self` unchanged is also the behavior that matches the existing contract: no source text means there's nothing to validate the quotes against, so leaving `substring_quotes` as the model produced them is consistent with the `info.context is None` path right above.

Test covers the exact shape (non-empty citation list, context dict missing the key), which is what distinguishes this from the already-handled `None` case.

Heads up that #2492 is the same one-line fix, opened a day later and without a test. This one is the more complete of the two.

ErenAta16:
Correction to my review above, where I said this was "the more complete of the two". There are four PRs for #2459, not two: #2460 (Jul 18, by the issue author, with a test), #2471 (closed), this one, and #2492. My check only covered recently-created PRs and missed the earlier ones.

Your version is still the only one carrying a CHANGELOG entry, and the fix and test are both right. But #2460 predates this by eight days and also has a test, so it has priority on ordering. Comparison table on #2460.

jxnl:

PR #2492, opened 2026-07-27
fix: prevent CitationMixin crash when validation_context is missing 'context' key (#2459)

### LINKED ISSUE #2459: CitationMixin crashes when validation_context missing 'context' key

shahriar654:
### Description
This PR addresses an issue where `CitationMixin.validate_sources` crashes with a `TypeError` if a user provides a `validation_context` dictionary but omits the expected `"context"` key. 

Previously, when the `"context"` key was missing, `info.context.get("context", None)` returned `None`. This `None` value was then passed directly into `self.get_spans()`, which caused the underlying `regex.search` engine to crash because it expected a string or bytes-like object.

### Changes Proposed
- Added a defensive check (`if text_chunks is None: return self`) directly after the `.get()` lookup in `instructor/v2/dsl/citation.py`.
- This safely skips source validation and returns the model instance untouched rather than crashing the application.

### Related Issues
Closes #2459

### DIFF
--- a/instructor/v2/dsl/citation.py
+++ b/instructor/v2/dsl/citation.py
@@ -69,6 +69,9 @@ class CitationMixin(BaseModel):
         # Get the context from the info
         text_chunks = info.context.get("context", None)
 
+        if text_chunks is None:
+            return self
+
         # Get the spans of the substring_phrase in the context
         spans = list(self.get_spans(text_chunks))
         # Replace the substring_phrase with the actual substring


ErenAta16 (commented):
The fix itself is right, `text_chunks` comes back `None` when the validation context exists but has no `context` key, and both downstream uses assume a string:

```
```

with `text_chunks[span[0]:span[1]]` failing on `None` as well. Returning early rather than patching either site is the correct shape, and it matches the `info.context is None` guard already two lines above.

Flagging an overlap though: #2488 landed the byte-identical change (`if text_chunks is None: return self`, same location) a day earlier, and additionally ships a regression test plus a CHANGELOG entry. Since this bug is a hard crash rather than a silent wrong result, a test is worth having so it can't come back, and #2488 already carries one.

Nothing wrong with this PR, it's just the strictly smaller subset of the earlier one. Probably worth consolidating on #2488 unless a maintainer prefers this one, in which case lifting the test across would be the thing to do.

ErenAta16:
Correction to my review above: I wrote "#2488 is the more complete of the two", but there are four PRs for #2459, not two. #2460 (Jul 18, by the issue author) predates both #2488 and this one by eight days and also ships a test, and #2471 was a fifth attempt that's since been closed. My duplicate check only looked at recently-created PRs, which missed the earlier ones. Apologies for the inaccurate framing.

The technical assessment of this PR stands, the fix is correct and the only gap is the missing test. But on merge order #2460 has priority rather than #2488.

Full comparison on #2460.

jxnl:
Closed as a duplicate implementation. The citation-context fix and regression coverage shipped in #2495 via the consolidated canonical patch.

PR #2498, opened 2026-07-30
fix(openai): add reask fallback when RESPONSES_TOOLS output has no tool calls

Kropiunig:
## What

`reask_responses_tools` builds its correction messages only from the tool calls in
`response.output`. When a failed Responses API completion contains no
`function_call` items — reasoning models routinely return only `reasoning`/`message`
output items instead of honoring the forced tool call — `reask_messages` ends up
empty and the handler returns the kwargs unchanged. Every retry then resends the
byte-identical request with no corrective feedback, so the full retry budget (and
its tokens) is burned on responses that can't improve.

Repro sketch:

```python
client = instructor.from_openai(OpenAI(), mode=instructor.Mode.RESPONSES_TOOLS)
# model answers with a reasoning/message output instead of the forced tool call
# -> parse raises ResponseParsingError (retryable)
# -> reask_responses_tools(kwargs, response, exc) returns kwargs UNCHANGED
# -> attempts 2..n are identical requests, then InstructorRetryException
```

## Why it's a bug, not a choice

- The `response is None or no .output` guard at the top of the same function *does*
  append a fallback user correction — the invariant "a reask always carries
  feedback" just never got applied to the has-output-but-no-tool-calls case.
- `reask_tools` (chat completions) gained exactly this fallback in #2495 (from
  #2448), and the Anthropic handler gained it in #2486 (#2485). RESPONSES_TOOLS is
  the remaining sibling without it.

## Fix

If the tool-call loop produced no messages, append the same plain user correction
used by the stream/None branch. One conditional, no behavior change when tool calls
are present.

## Tests

- `test_reask_responses_tools_no_tool_calls_adds_fallback_message` — reasoning +
  message output items, asserts a user correction containing the validation error is
  appended. Fails on `main` (`assert 1 == 2`, messages unchanged), passes with the fix.
- Full `tests/test_openai_responses_tools.py`, `tests/test_streaming_reask_bug.py`,
  `tests/coverage/test_openai_handlers_coverage.py`: 74 passed, 1 skipped.
- CHANGELOG entry added under `[Unreleased]`.


jxnl:
Consolidated in #2502, now merged to main. Your original commit and authorship were preserved in the merge, with the Responses API no-tool-call retry regression and changelog entry included. Closing this source PR as superseded by the consolidation.

PR #2499, opened 2026-07-31
fix: accumulate newer billable token fields in usage tracking

### LINKED ISSUE #2493: Retry usage accumulation drops newer billable token fields (Anthropic stale, OpenAI overwritten with None)

rkfshakti:
## Problem

`update_total_usage` was hand-maintaining a list of token fields that drifted from both SDKs. On Anthropic, `cache_creation`, `server_tool_use`, and `output_tokens_details` sub-models were never initialized on the accumulator, so their fields stayed stale across retries. On OpenAI, `prompt_tokens_details` sub-model fields (`cache_write_tokens`, `accepted_prediction_tokens`, `rejected_prediction_tokens`) were overwritten with `None` because the accumulator never populated them.

This was silent — `input_tokens`/`prompt_tokens` were correct, so the under-count only showed up when reconciling against a provider invoice.

## Root cause

Both implementations enumerate fields by hand and drift independently as the SDKs add counters. Each new billable field needs a matching edit in two places, and nothing fails when that edit is missed.

## Fix

**Anthropic** (`instructor/v2/providers/anthropic/usage.py`):
- Added `_ensure_sub_models()` that initializes `cache_creation`, `server_tool_use`, and `output_tokens_details` on the accumulator with zeroed values
- Accumulate `ephemeral_1h_input_tokens`, `ephemeral_5m_input_tokens` (cache_creation), `web_search_requests`, `web_fetch_requests` (server_tool_use), and `thinking_tokens` (output_tokens_details) across retries
- Write back the accumulated sub-models onto the response so callers see them

**OpenAI** (`instructor/v2/core/usage.py`):
- Accumulate `cache_write_tokens`, `accepted_prediction_tokens`, and `rejected_prediction_tokens` in `prompt_tokens_details`, using `hasattr` guards for forward compatibility with older SDK versions

## Verification

```python
# Anthropic: 3 retries each with thinking=40, web_search=2, ephemeral_5m=500
# Before: thinking=40 (stale), web_search=2 (stale), eph5m=500 (stale)
# After:  thinking=120, web_search=6, eph5m=1500 ✅
```

Fixes #2493

### DIFF
--- a/instructor/v2/core/usage.py
+++ b/instructor/v2/core/usage.py
@@ -41,6 +41,19 @@ def update_total_usage(
         ):
             tpd.audio_tokens = (tpd.audio_tokens or 0) + (rpd.audio_tokens or 0)
             tpd.cached_tokens = (tpd.cached_tokens or 0) + (rpd.cached_tokens or 0)
+            # Accumulate newer prediction-related fields (openai>=2.45.0).
+            if hasattr(rpd, "cache_write_tokens"):
+                tpd.cache_write_tokens = (getattr(tpd, "cache_write_tokens", None) or 0) + (
+                    rpd.cache_write_tokens or 0
+                )
+            if hasattr(rpd, "accepted_prediction_tokens"):
+                tpd.accepted_prediction_tokens = (
+                    getattr(tpd, "accepted_prediction_tokens", None) or 0
+                ) + (rpd.accepted_prediction_tokens or 0)
+            if hasattr(rpd, "rejected_prediction_tokens"):
+                tpd.rejected_prediction_tokens = (
+                    getattr(tpd, "rejected_prediction_tokens", None) or 0
+                ) + (rpd.rejected_prediction_tokens or 0)
         response_usage.completion_tokens = total_usage.completion_tokens
         response_usage.prompt_tokens = total_usage.prompt_tokens
         response_usage.total_tokens = total_usage.total_tokens
--- a/instructor/v2/providers/anthropic/usage.py
+++ b/instructor/v2/providers/anthropic/usage.py
@@ -17,6 +17,30 @@ def initialize_usage() -> Any:
     )
 
 
+def _ensure_sub_models(total_usage: Any) -> None:
+    """Ensure nested usage sub-models exist on the accumulator, zeroed."""
+    from anthropic.types.usage import (
+        CacheCreation,
+        ServerToolUsage,
+        OutputTokensDetails,
+    )
+
+    if not total_usage.cache_creation:
+        total_usage.cache_creation = CacheCreation(
+            ephemeral_1h_input_tokens=0,
+            ephemeral_5m_input_tokens=0,
+        )
+    if not total_usage.server_tool_use:
+        total_usage.server_tool_use = ServerToolUsage(
+            web_search_requests=0,
+            web_fetch_requests=0,
+        )
+    if not total_usage.output_tokens_details:
+        total_usage.output_tokens_details = OutputTokensDetails(
+            thinking_tokens=0,
+        )
+
+
 def update_total_usage(response_usage: Any, total_usage: Any) -> bool:
     """Accumulate Anthropic token usage into a running total when applicable."""
     from anthropic.types import Usage as AnthropicUsage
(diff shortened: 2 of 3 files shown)


ErenAta16 (commented):
I filed #2493, so I am not neutral here. Reading this against the pinned dependency turned up a blocker, and it also corrects something in my own issue.

**`_ensure_sub_models` raises `ImportError` on the version this repo pins.** `pyproject.toml` pins `anthropic==0.93.0` (line 60, and the `anthropic` extra on line 87). That release has no `output_tokens_details` at all. Installed and ran the exact import from this PR:

```
$ pip install anthropic==0.93.0
>>> from anthropic.types.usage import CacheCreation, ServerToolUsage, OutputTokensDetails
ImportError: cannot import name 'OutputTokensDetails' from 'anthropic.types.usage'

>>> sorted(Usage.model_fields)
['cache_creation', 'cache_creation_input_tokens', 'cache_read_input_tokens',
 'inference_geo', 'input_tokens', 'output_tokens', 'server_tool_use', 'service_tier']
```

`src/anthropic/types/output_tokens_details.py` does not exist at tag `v0.93.0`; it first appears later. Since `_ensure_sub_models` is called from `update_total_usage` on every attempt, this does not degrade Anthropic usage tracking, it stops it with an `ImportError` on the first call under the pinned dependency.

`CacheCreation` and `ServerToolUsage` are fine. Both exist at `v0.93.0` with exactly the fields used here (`ephemeral_1h_input_tokens`/`ephemeral_5m_input_tokens`, `web_fetch_requests`/`web_search_requests`), so those two thirds work.

**This is my error as much as this PR's.** I wrote #2493 against `anthropic==0.116.0` and listed `output_tokens_details.thinking_tokens` as one of the three unaccumulated fields. It is not a field on the version instructor actually installs, so that third of the issue does not apply to the pinned SDK. The `cache_creation` and `server_tool_use` halves do apply on `0.93.0` and are real. Sorry for sending you after a field that is not there.

**The narrower point this exposes.** A fix here has to survive a version range, and hand-enumerating fields cannot. The same drift that produced the original bug is what makes `OutputTokensDetails` unimportable now, and it will do it again the next time Stainless regenerates the model. A shape-driven merge (walk the response model's own numeric fields and sum whatever is present) gets the version tolerance for free. Both #2497 and #2500 go that route, which is worth weighing before this lands.

**Secondary, and it survives fixing the import.** Even guarded, the accumulator would attach `output_tokens_details` to a `0.93.0` `Usage`, which permits extra attributes:

```
>>> u = Usage(input_tokens=1, output_tokens=1)
>>> hasattr(u, "output_tokens_details")
False
>>> u.output_tokens_details = None      # accepted, phantom attribute
```

so the write-back at the end would put a field on the response object that the installed SDK does not model, and `model_dump()` consumers would see a key the provider never sent.

**Also worth flagging: the write-back reintroduces the pattern the OpenAI half of #2493 was about.**

```python
response_usage.cache_creation = total_usage.cache_creation.model_copy(deep=True)
```

This is the same unconditional overwrite that wipes fields the accumulator does not track. `CacheCreation` has exactly two fields today so nothing is lost yet, but that was equally true of `PromptTokensDetails` before OpenAI added `cache_write_tokens`. Merging into the response object rather than replacing it would close that off permanently.

No tests in this PR. Given the accumulation is order-dependent across attempts, I would want at least a three-attempt case pinning that the sub-model fields sum rather than latch, which is the exact symptom in the issue.

rkfshakti:
Thanks @ErenAta16 for the thorough review — all three points are addressed in commit e57355ed:

1. **ImportError on anthropic==0.93.0** — `OutputTokensDetails` import is now guarded with try/except ImportError. All `output_tokens_details` access uses `getattr`/`hasattr` since the field itself isn't a declared member of `Usage` in 0.93.0. The `_ensure_sub_models` function returns early when the import fails, so the `cache_creation` and `server_tool_use` halves still work.

2. **Write-back merges not replaces** — Changed from `model_copy(deep=True)` replacement to field-by-field merge for all three sub-models. Future SDK additions that the accumulator doesn't track are preserved on the response object.

3. **Regression tests added** — 8 tests in `tests/llm/test_anthropic/test_usage_accumulation.py` covering: sub-model initialisation, three-attempt accumulation for `cache_creation` and `server_tool_use`, no crash on pinned anthropic (the exact regression), and merge-not-replace write-back. The `output_tokens_details` accumulation test is skipped on 0.93.0 since the field doesn't exist there.

All 7 runnable tests pass on anthropic==0.93.0.

jxnl:
Superseded by #2502, now merged to main. The merged fix uses one recursive provider-agnostic accumulator instead of enumerating SDK fields, avoiding drift and pinned Anthropic SDK import incompatibilities. Closing this overlapping patch in favor of the consolidated implementation.

PR #2501, opened 2026-08-02
fix(v2): keep raw Iterable typehint for PARALLEL_TOOLS

RezaRahemtola:
## Describe your changes

`Mode.PARALLEL_TOOLS` raises `TypeError` on every `.create(...)` since #2495, making the mode unusable.

It's due to the patch wrapper running `prepare_response_model` before dispatch, converting `Iterable[X]` into a class (`get_origin -> None`), but `get_types_array` requires the raw `Iterable[...]` typehint.

This PR skips `prepare_response_model` for `Mode.parallel_modes()` and keep the raw typehint as the `response_model` handed to `parse_response`. Existing PARALLEL_TOOLS tests call `handlers.request_handler` directly and bypass the wrapper, so the new tests go through `patch_v2` to cover that path.

None, it's a pre-release regression since #2495 I noticed while running the version on `main` that I fixed directly.

### DIFF
--- a/instructor/v2/core/patch.py
+++ b/instructor/v2/core/patch.py
@@ -209,13 +209,15 @@ def _create_sync_wrapper(
         # Get handlers from registry
         handlers = mode_registry.get_handlers(provider, mode)
 
-        if response_model is not None:
+        if response_model is not None and mode not in Mode.parallel_modes():
             response_model = prepare_response_model(response_model)
 
         # Prepare request kwargs using registry handler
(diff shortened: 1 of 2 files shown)


jxnl:
Consolidated in #2502, now merged to main. Your original commit and authorship were preserved, including sync and async wrapper regression tests for raw parallel Iterable type hints. Closing this source PR as superseded by the consolidation.

PR #2507, opened 2026-08-06
fix: add request timeouts to multimodal URL fetches

bunlongheng:
## Problem

Several `requests.get` / `requests.head` calls in the multimodal helpers download image, audio, and PDF content from remote URLs **without passing a timeout**. Because Python's `requests` has no default timeout, a slow or unresponsive host can hang the calling thread indefinitely (uncontrolled resource consumption, CWE-400). This is a denial-of-service hardening gap: a single unreachable media URL can stall a request pipeline forever.

The GCS helpers in the same module (`from_gs_url`) already use a 30s timeout, and `bedrock/handlers.py` uses `timeout=30` too, so the codebase already treats 30s as the convention. This PR simply extends that same default to the remaining fetches.

## Fixed call sites

- `instructor/v2/core/multimodal.py` - `Image.from_url` (HEAD probe), `Image.url_to_base64`, `Audio.from_url`, `PDF.from_url` (HEAD probe)
- `instructor/v2/providers/anthropic/multimodal.py` - `pdf_to_anthropic`
- `instructor/v2/providers/genai/multimodal.py` - `image_to_genai`, `pdf_to_genai`
- `instructor/v2/providers/openai/multimodal.py` - `pdf_to_openai`

## Change

Each remote fetch now passes `timeout=30`, matching the existing convention. No behavior change for healthy hosts; unresponsive hosts now fail fast instead of hanging.

## Tests

Added two regression tests in `tests/multimodal/test_multimodal.py` asserting a timeout is passed on both the `requests.get` and `requests.head` paths. Full multimodal suite passes: 62 passed, 1 skipped.

### DIFF
--- a/instructor/v2/core/multimodal.py
+++ b/instructor/v2/core/multimodal.py
@@ -203,7 +203,7 @@ class Image(BaseModel):
 
         if not media_type:
             try:
-                response = requests.head(url, allow_redirects=True)
+                response = requests.head(url, allow_redirects=True, timeout=30)
                 media_type = response.headers.get("Content-Type")
             except requests.RequestException as e:
                 raise ValueError(f"Failed to fetch image from URL") from e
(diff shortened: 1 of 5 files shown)


jxnl:
Consolidated and merged in #2510. The original timeout changes and authorship were preserved; the consolidation also repaired four stale request mocks, asserted all changed timeout paths, and passed the full CI/release matrix.

PR #2508, opened 2026-08-06
fix(openai): route streaming retries through the streaming parser in TOOLS/JSON/MD_JSON

RezaRahemtola:
## Summary

`create_partial`/`create_iterable` with OpenAI TOOLS/JSON/MD_JSON modes fails on retry: the per-model streaming flag is consumed on the first attempt, and on retry the handler falls through to the non-streaming parser (but the response is still a `Stream`, producing `ResponseParsingError("No choices in OpenAI response")` instead of the real validation error). Even when the model self-corrects on retry, the corrected stream is discarded.

The JSON_SCHEMA and RESPONSES_TOOLS handlers already gate correctly the streaming path but the other 3 handlers were missing the `stream or` condition.

## Fix

Add `stream or` to the gate in all three handlers, mirroring the JSON_SCHEMA sibling.

### DIFF
--- a/instructor/v2/providers/openai/handlers.py
+++ b/instructor/v2/providers/openai/handlers.py
@@ -718,10 +718,11 @@ class OpenAIToolsHandler(OpenAIHandlerBase):
     ) -> Any:
         """Parse tool call response."""
         # Check for streaming
-        consume_streaming = isinstance(
-            response_model, type
-        ) and self._consume_streaming_flag(response_model)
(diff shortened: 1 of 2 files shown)


jxnl:
Consolidated and merged in #2510 with the contributor commit and authorship preserved. The final version shares one consumptive streaming gate across TOOLS, JSON, JSON_SCHEMA, and MD_JSON and adds retry coverage for every mode.

Issue #1523, opened 2025-05-13
Using | breaks with partial streaming


ahuang11:
- [ ] This is actually a bug report.

**What Model are you using?**

gpt-4o-mini

**Describe the bug**

Follow up on https://github.com/567-labs/instructor/issues/661 since that was closed before it was fixed

When streaming with a `|`, it crashes with `TypeError: type 'types.UnionType' is not subscriptable`

**To Reproduce**

```python
import instructor
from pydantic import BaseModel
from openai import OpenAI


class User(BaseModel):
    name: str
    age: int
    category: str | None

client = instructor.from_openai(OpenAI())

# Extract structured data from natural language
res = client.chat.completions.create_partial(
    model="gpt-4o-mini",
    response_model=User,
    messages=[{"role": "user", "content": "John Doe is 30 years old."}],
)
res
```

**Expected behavior**

Swapping to Union works, but I expect them both to work

```python
import instructor
from pydantic import BaseModel
from openai import OpenAI


class User(BaseModel):
    name: str
    age: int
    category: Union[str, None]

client = instructor.from_openai(OpenAI())

# Extract structured data from natural language
res = client.chat.completions.create_partial(
    model="gpt-4o-mini",
    response_model=User,
    messages=[{"role": "user", "content": "John Doe is 30 years old."}],
)
res
```

Issue #661, opened 2024-05-11
create_partial errors out with union types


wiseman:
- [ ] This is actually a bug report.
- [ ] I am not getting good LLM Results
- [ ] I have tried asking for help in the community on discord or discussions and have not received a response.
- [ x] I have tried searching the documentation and have not found an answer.

**What Model are you using?**

- [ x] gpt-3.5-turbo
- [ ] gpt-4-turbo
- [ x] gpt-4
- [ ] Other (please specify)

**Describe the bug**
`create_partial` doesn't work with models that contain union types.

**To Reproduce**
```
import instructor
import openai
from pydantic import BaseModel

class Age(BaseModel):
    description: str | int


client = instructor.from_openai(openai.OpenAI())
partial_age = client.chat.completions.create_partial(
    model="gpt-4",
    response_model=Age,
    messages=[
        {
            "role": "user",
            "content": "I am 25 years old.",
        },
    ],
    stream=True,
)
print(partial_age)
```

Running the above code results in this:

```
Traceback (most recent call last):
  File "/Users/wiseman/src/notams/bug.py", line 10, in <module>
    partial_age = client.chat.completions.create_partial(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wiseman/src/notams/venv/lib/python3.11/site-packages/instructor/client.py", line 137, in create_partial
    response_model = instructor.Partial[response_model]  # type: ignore
                     ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/Users/wiseman/src/notams/venv/lib/python3.11/site-packages/instructor/dsl/partial.py", line 285, in __class_getitem__
    **{
      ^
  File "/Users/wiseman/src/notams/venv/lib/python3.11/site-packages/instructor/dsl/partial.py", line 289, in <dictcomp>
    else _wrap_models(field_info)
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wiseman/src/notams/venv/lib/python3.11/site-packages/instructor/dsl/partial.py", line 269, in _wrap_models
    generic_base[modified_args] if generic_base else None
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: type 'types.UnionType' is not subscriptable
```

**Expected behavior**

I expected that create_partial wouldn't error out on a union type.

[...]

PR #2509, opened 2026-08-07
fix(v2): parse PEP 604 unions in iterable streaming

Kropiunig:
## Problem

Streaming a union of models works with `Union[A, B]` but crashes with the PEP 604
spelling `A | B`:

```python
client.chat.completions.create_iterable(
    response_model=Weather | GoogleSearch,
    messages=[...],
)
```

```
AttributeError: 'types.UnionType' object has no attribute 'model_validate_json'
```

It raises on the first streamed item. `response_model=Union[Weather, GoogleSearch]`
— the spelling used in `docs/concepts/iterable.md` — works fine, so the two
spellings of the same type disagree.

## Cause

`create_iterable` wraps the model as `Iterable[Weather | GoogleSearch]`, so
`IterableBase.task_type` holds a `types.UnionType`. In
`instructor/v2/dsl/iterable.py`, `extract_cls_task_type` matched only
`typing.Union`:

```python
if get_origin(cls.task_type) is Union:
```

On Python 3.10-3.13 `get_origin(A | B)` is `types.UnionType`, so PEP 604 unions
skipped the member-by-member branch and called `model_validate_json` on the union
object itself.

The same file already defines `_UNION_ORIGINS = (Union, UnionType)` (lines 17-22)
and `IterableModel` uses it, so `Iterable[A | B]` builds a correct
`IterableWeatherOrGoogleSearch` wrapper and then fails to parse into it.
`dsl/parallel.py`, `dsl/partial.py` and `core/response_model.py` all already match
on `UnionType`; `extract_cls_task_type` was the only site left out.

## Fix

Match `_UNION_ORIGINS` in `extract_cls_task_type`. No new imports; `typing.Union`
behaviour is unchanged, and the Python 3.9 fallback (`_UNION_ORIGINS = (Union,)`)
is preserved.

One helper covers four streaming entry points — `tasks_from_chunks`,
`tasks_from_chunks_async`, `tasks_from_task_list_chunks`,
`tasks_from_task_list_chunks_async` — so sync, async and task-list streaming are
all fixed.

## Tests

`tests/v2/test_iterable_streaming.py` gains parametrised cases running both union
spellings through `IterableModel` and through the `Iterable[...]` type that
`create_iterable` builds, plus the unmatched-payload error path.

Without the fix: `3 failed, 6 passed` (only the `pep604-union` cases fail).
With the fix: `9 passed`.

Surrounding modules (`tests/v2/`, `tests/dsl/`, `tests/processing/`,
`tests/coverage/test_dsl_iter_parallel_coverage.py`) go from `7 failed, 1812
passed` on `main` to `7 failed, 1817 passed` — the same pre-existing failures
(missing optional provider deps locally), plus the 5 new cases.

Relates to the `|`-vs-`Union` streaming issues #661 / #1523 / #2088, which were
fixed in `dsl/partial.py`; this applies the same treatment to `dsl/iterable.py`.
It is a separate call site, not a duplicate of those fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Iterable streaming unions**: Parse PEP 604 unions (`create_iterable(response_model=Weather | GoogleSearch)`, `Iterable[Weather | GoogleSearch]`) member by member instead of raising `AttributeError: 'types.UnionType' object has no attribute 'model_validate_json'`, matching the existing `Union[Weather, GoogleSearch]` behaviour.
+
 ## [1.15.5] - 2026-08-02
 
 ### Fixed
--- a/instructor/v2/dsl/iterable.py
+++ b/instructor/v2/dsl/iterable.py
@@ -140,7 +140,11 @@ class IterableBase:
         **kwargs: Any,
     ):
         assert cls.task_type is not None
-        if get_origin(cls.task_type) is Union:
+        # ``Iterable[A | B]`` stores a ``types.UnionType`` task type, whose origin is
+        # ``types.UnionType`` rather than ``typing.Union`` on Python 3.10-3.13. Match on
+        # ``_UNION_ORIGINS`` so PEP 604 unions take the member-by-member branch instead of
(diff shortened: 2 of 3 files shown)


jxnl:
Consolidated and merged in #2510 with the contributor commit and authorship preserved. The final regression suite also avoids evaluating PEP 604 unions during Python 3.9 collection and passed the full supported-version matrix.

PR #2086, opened 2026-02-16
feat(bedrock): add native structured outputs and strict tool use modes

ndgigliotti:
## Describe your changes

Add `BEDROCK_STRUCTURED_OUTPUTS` and `BEDROCK_TOOLS_STRICT` modes that leverage Bedrock's native constrained decoding ([announced Feb 2026](https://aws.amazon.com/about-aws/whats-new/2026/02/structured-outputs-available-amazon-bedrock/)) for guaranteed JSON schema compliance via the Converse API.

**What's included:**

- **`BEDROCK_STRUCTURED_OUTPUTS`** — uses `outputConfig.textFormat` with `json_schema` type for native schema-constrained text generation
- **`BEDROCK_TOOLS_STRICT`** — adds `strict: true` to `toolSpec` for native schema enforcement on tool use
- `_set_additional_properties_false()` helper to recursively patch Pydantic schemas (Bedrock requires `additionalProperties: false` on all object types)
- Handler, reask, and parser functions for both modes
- Auto-routing in `from_provider()` — models that support structured outputs default to `BEDROCK_STRUCTURED_OUTPUTS` instead of `BEDROCK_JSON`/`BEDROCK_TOOLS`
- 22 unit tests covering handlers, reask, parsing, registry, and mode classification

**Smoke-tested models (all pass first-try):**
- Anthropic: Claude Haiku 4.5, Sonnet 4.5, Opus 4.5, Opus 4.6
- Qwen: Qwen3 32B, Qwen3 Coder
- Google: Gemma 3 12B, 27B
- MiniMax: M2
- Mistral: Large 3, Ministral 8B, Voxtral Small
- Moonshot: Kimi K2 Thinking
- NVIDIA: Nemotron Nano 9B, 12B

**Known broken models** (excluded from auto-routing, documented):
- OpenAI GPT-OSS (all variants) — returns markdown-wrapped JSON, ignores `outputConfig`
- Mistral Magistral Small 2509 — returns free-form text, ignores `outputConfig`
- DeepSeek V3.2 — API rejects `outputConfig.textFormat` with `ValidationException`

## Issue ticket number and link

Closes #2084

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] If it is a core feature, I have added documentation.

PR #2514, opened 2026-08-08
fix(bedrock): parse reasoning-prefixed JSON

### LINKED ISSUE #2076: Allow MD_JSON for bedrock client

jxnl:
## Summary

- parse the final complete JSON object or array from Bedrock `MD_JSON` responses using the shared v2 extractor
- handle reasoning prose and `<think>...</think>` blocks even when they contain valid or malformed brace spans
- preserve escaped JSON content instead of deleting literal `\\n` sequences
- copy message lists before native-system conversion and Bedrock reasks so caller-owned input remains unchanged

This is the current-main canonical replacement for #2287 and preserves the intent and attribution of @octo-patch. It fixes #2076 once merged.

## Validation

- focused Bedrock and shared JSON tests: `66 passed`
- exact GitHub core-test scope: `2024 passed, 144 skipped`
- exact offline coverage scope: `760 passed`, `93%`
- focused branch coverage exercises every changed Bedrock handler statement
- maintained Ruff and format scopes: passed
- strict source and test `ty`: passed
- all pre-commit hooks, lock validation, and `git diff --check`: passed

## Intentionally separate

- #2086 / #2084 native Bedrock structured outputs: useful but needs a separate current-v2 reconstruction without hard-coded model routing
- #2409 / #2408 bearer-token authentication: held because the existing patch mutates process-global credential state

## Merge sequencing

This PR is draft while the already validated `1.15.5` candidate remains unpublished. It should not merge past `1454af92` until that release is published or the maintainer explicitly folds the version history into `1.16.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Bedrock reasoning JSON**: Parse the final complete JSON value after reasoning text or `<think>` blocks, preserve JSON escape sequences, and keep caller-owned messages unchanged during Bedrock request preparation and retries. ([#2076](https://github.com/567-labs/instructor/issues/2076), [#2287](https://github.com/567-labs/instructor/pull/2287))
+
 ## [1.15.5] - 2026-08-07
 
 ### Fixed
--- a/instructor/v2/providers/bedrock/handlers.py
+++ b/instructor/v2/providers/bedrock/handlers.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import base64
 import json
 import mimetypes
-import re
 from textwrap import dedent
 from typing import Any, cast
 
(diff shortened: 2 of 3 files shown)


jxnl:
1.16 integration update: this component remains independently green, and the exact six-PR aggregate is now fully green in draft #2517 at `0980b3da` across Ruff, `ty`, full coverage, Python 3.9-3.13, provider jobs, and nonpublishing release readiness. The stacked versioned `1.16.0` candidate is also fully green in draft #2518 at `de3ceaf5`, including artifact build and Python 3.9/3.13 wheel smokes. Keep this PR draft until the unpublished 1.15.5 sequencing decision is approved.

jxnl:
Bedrock reasoning-prefixed JSON parsing and caller-message preservation are now on `main` via #2517 at `ad95a3d4`. Closing this independently reviewed draft as consolidated; its focused history and attribution remain here, and the final folded 1.16 release metadata is in #2519.

PR #2521, opened 2026-08-11
fix(cli): add type annotation to MODEL_COSTS in usage.py

nishant4671:
## Summary
Adds explicit type annotation to `MODEL_COSTS` in `instructor/cli/usage.py` to fix static type-checking errors.

## Motivation
Running `mypy instructor/cli/usage.py` resulted in 7 `[return-value]` errors inside `get_model_cost()` because `MODEL_COSTS` contained mixed value types (`dict` and `float`) without a type hint, causing `mypy` to infer the return type as generic `object`.

## Changes
- Annotated `MODEL_COSTS` as `dict[str, Union[dict[str, float], float]]`.

## Verification
- Verified locally with `mypy instructor/cli/usage.py` (0 errors found).


jxnl:
Closing as consolidated by https://github.com/567-labs/instructor/pull/2564, which preserves this CLI cost-table annotation without the branch's unrelated commits. Thank you for the contribution.

PR #2524, opened 2026-08-12
fix(cache): include hoisted system prompt in the cache key

Kropiunig:
With `cache=` enabled on an Anthropic (or Bedrock) client, two calls that differ only in their system prompt hit the same cache entry, so the second one silently returns the first one's answer:

```python
client = instructor.from_anthropic(anthropic.Anthropic(), mode=instructor.Mode.JSON)
cache = AutoCache(maxsize=10)

def ask(system):
    return client.create(
        model="claude-3-5-sonnet-latest",
        max_tokens=100,
        messages=[
            {"role": "system", "content": system},
            {"role": "user", "content": "hello"},
        ],
        response_model=Answer,
        cache=cache,
    )

ask("Always answer in French.")   # provider called
ask("Always answer in German.")   # provider NOT called - returns the French answer
```

## Root cause

The Anthropic and Bedrock request handlers hoist `role="system"` messages out of `messages` into a separate top-level `system` parameter (`anthropic/handlers.py:346-352`, `bedrock/handlers.py:387`). `patch.py` computes the cache key *after* the request handler runs, and `make_cache_key` only hashes `messages` / `contents` / `chat_history` — so the system prompt is not part of the key at all for those providers.

OpenAI-style providers keep the system message inside `messages`, which is why this never showed up there. The docstring on `make_cache_key` and the cache-key table in `docs/concepts/caching.md` both claim the system prompt is covered, so the current behaviour also contradicts the documented contract.

## Fix

`make_cache_key` takes an optional `system` argument and adds it to the hashed payload only when it is not `None`. `patch.py` passes `new_kwargs.get("system")` at the four call sites (sync/async x lookup/store). Because the field is omitted when absent, keys for providers that keep the system prompt in `messages` are byte-identical to before, so no existing cache is invalidated by this change.

`system` is stable across the retry loop — the Anthropic reask handlers only append to `messages` — so the lookup key and the store key still agree after a reask (the case #2457 fixed).

## Tests

- `tests/cache/test_cache_integration.py::test_cache_key_accounts_for_hoisted_system_prompt` — end-to-end through the patch layer with a stubbed `messages.create`, no API key needed. Asserts a different system prompt calls the provider again, and that repeating the original request still hits the cache.
- `tests/cache/test_cache_key.py` — two unit tests: a changed system prompt busts the key, and passing `system=None` produces the same key as omitting it (backwards compatibility).

Before the fix:

```
        second = ask("Always answer in German.")
>       assert call_counter["n"] == 2, (
            "A different system prompt must not reuse the cached response"
        )
E       AssertionError: A different system prompt must not reuse the cached response
E       assert 1 == 2
```

After: `9 passed` for `tests/cache/`.

Note: the `genai`/`gemini` handlers put the system prompt inside a nested `config` object rather than a top-level kwarg, so they are not covered by this change — happy to follow up separately if you want that hashed too.

### DIFF
--- a/instructor/cache/__init__.py
+++ b/instructor/cache/__init__.py
@@ -148,6 +148,7 @@ def make_cache_key(
     model: str | None,
     response_model: type[BaseModel] | None,
     mode: str | None = None,
+    system: Any = None,
 ) -> str:  # noqa: ANN401
     """Compute a *deterministic* cache key.
 
@@ -157,6 +158,10 @@ def make_cache_key(
     Components that influence the key:
         • provider/model name
         • serialized *messages* (user + system prompt, etc.)
+        • *system* – providers such as Anthropic and Bedrock hoist system
+          messages out of ``messages`` into a separate top-level parameter,
+          so it has to be hashed separately or two calls that only differ in
+          their system prompt would collide.
         • *mode* (Tools, JSON, …) – helps when users change Instructor mode
         • *response_model* schema – so edits to field definitions or
           descriptions invalidate prior cache entries (critical!).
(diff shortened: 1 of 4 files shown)


jxnl:
Closing as consolidated by https://github.com/567-labs/instructor/pull/2564, which includes the hoisted system prompt in sync and async cache keys. Thank you for the contribution.

PR #2525, opened 2026-08-12
fix(core): normalize remote media types

weivwang:
## What

Normalize remote `Content-Type` headers before validating image, audio, and PDF media types.

## Why

HTTP media types are case-insensitive and may include optional parameters. Valid responses such as `application/pdf; charset=binary` or `Audio/WAV; codecs=1` were rejected because the complete header value was compared against the allowlist.

## Changes

- centralize media-type normalization in the v2 multimodal core
- strip optional parameters, surrounding whitespace, and casing differences
- apply the normalized value to HTTP and GCS image, audio, and PDF sources
- add deterministic regression coverage for all three media categories

## Testing

- `uv run pytest tests/coverage/test_core_multimodal_coverage.py -q` (28 passed)
- `uv run ruff check instructor/v2/core/multimodal.py tests/coverage/test_core_multimodal_coverage.py`
- `uv run ruff format --check instructor/v2/core/multimodal.py tests/coverage/test_core_multimodal_coverage.py`
- `uv run ty check instructor/v2/core/multimodal.py tests/coverage/test_core_multimodal_coverage.py`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Remote multimodal media types**: Accept valid case-insensitive HTTP `Content-Type` values with optional parameters when loading images, audio, and PDFs. ([#2525](https://github.com/567-labs/instructor/pull/2525))
+
 ## [1.16.0] - 2026-08-09
 
 ### Added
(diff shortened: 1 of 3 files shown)


jxnl:
Closing as consolidated by https://github.com/567-labs/instructor/pull/2564, with parameterized/case-insensitive media-type handling and the missing-header boundary covered. Thank you for the contribution.

PR #2527, opened 2026-08-13
fix(messages): preserve tool-call protocol fields

### LINKED ISSUE #2526: merge_consecutive_messages drops tool-call fields from conversation history
## Description

`merge_consecutive_messages()` rebuilds every output message from only `role` and `content`. This silently drops protocol fields such as `tool_calls`, `tool_call_id`, `name`, and `function_call`, even when the message is not merged with another message.

The helper is used by the OpenAI-compatible, Mistral, and Writer JSON / Markdown-JSON request handlers, so a caller that reuses tool-call history can have a valid assistant-tool exchange corrupted before the provider request is sent.

## Reproduction

```python
from instructor.v2.core.messages import merge_consecutive_messages

messages = [
    {
        "role": "assistant",
        "content": None,
        "tool_calls": [
            {
                "id": "call_1",
                "type": "function",
                "function": {"name": "lookup", "arguments": "{}"},
            }
        ],
    },
    {
        "role": "tool",
        "content": "result",
        "tool_call_id": "call_1",
    },
]

print(merge_consecutive_messages(messages))
```

Current output on `main`:

```python
[
    {"role": "assistant", "content": [{"type": "text", "text": ""}]},
    {"role": "tool", "content": [{"type": "text", "text": "result"}]},
]
```

Both `tool_calls` and `tool_call_id` are gone.

## Expected behavior

Messages carrying protocol-specific fields should preserve those fields and should not be merged across semantic tool/function-call boundaries. Plain adjacent role/content messages can continue using the existing merge behavior.

## Proposed fix

Only merge messages whose keys are limited to `role` and `content`. When a message has any additional fields, keep it as a separate message and copy those fields to the output. Add regressions for assistant `tool_calls`, tool `tool_call_id`, and consecutive tool-call-only assistant messages.

I am working on a focused patch with offline regression coverage.

FU-max-boop:
## Describe your changes

`merge_consecutive_messages()` rebuilt every message from only `role` and `content`. That silently removed protocol fields such as `tool_calls`, `tool_call_id`, `name`, and `function_call`, even when the message was not merged.

This change:
- merges only adjacent plain `role`/`content` messages;
- treats messages with protocol metadata as boundaries and preserves all their fields;
- keeps the existing string/list content normalization behavior;
- adds regression coverage for consecutive assistant tool calls and linked tool results;
- documents the fix in the changelog.

Fixes #2526

## Validation

- `pytest` across message processing and affected OpenAI-compatible/Mistral handler suites: 215 passed, 1 skipped
- `ruff check` and `ruff format --check` on changed Python files
- `ty check` on changed Python files
- `git diff --check`
- manual behavior matrix covering plain merges, alternating roles, consecutive tool-call messages, tool results, and input immutability

AI assistance was used for issue reproduction, implementation review, and test planning; all code paths and test outputs were manually verified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Message history**: Preserve tool-call and other protocol fields when normalizing consecutive messages, and keep protocol messages as separate turns. ([#2527](https://github.com/567-labs/instructor/pull/2527))
+
 ## [1.16.0] - 2026-08-09
 
 ### Added
--- a/instructor/v2/core/messages.py
+++ b/instructor/v2/core/messages.py
@@ -70,6 +70,7 @@ def dump_message(message: ChatCompletionMessage) -> ChatCompletionMessageParam:
 
 
 def merge_consecutive_messages(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Merge adjacent plain messages without crossing protocol boundaries."""
     if not messages:
         return []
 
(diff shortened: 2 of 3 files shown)


jxnl:
Closing as consolidated by https://github.com/567-labs/instructor/pull/2564, which preserves protocol fields and caller-owned list content during message normalization. Thank you for the contribution.

PR #2533, opened 2026-08-16
fix(v2): preserve protocol fields when merging consecutive messages

feiiiiii5:
## Description

Fixes #2526: `merge_consecutive_messages()` rebuilt every output message from only `role` and `content`, silently dropping protocol fields such as `tool_calls`, `tool_call_id`, `name`, and `function_call` — even for messages that were never merged. The helper is used by the OpenAI-compatible, Mistral, and Writer JSON / Markdown-JSON request handlers, so a caller reusing tool-call history had a valid assistant-tool exchange corrupted before the provider request was sent.

This PR preserves those fields on the rebuilt messages and never merges across tool/function-call boundaries; plain adjacent role/content messages still merge as before.

## Type of Change
- [x] Bug fix

## Test
- New `tests/core/test_messages.py` (4 unit tests): tool_calls/tool_call_id preserved; no merge across tool boundaries; plain messages still merge; name/function_call preserved.
- Pre-fix FAIL verified: with main's implementation, 3 of the 4 tests fail; post-fix 4/4 pass.
- `ruff check` and `ruff format --check` clean on both changed files.

## Checklist
- [x] Tests included
- [x] CHANGELOG.md entry added under [Unreleased]
- [x] Conventional commit format

Closes #2526

PR #2534, opened 2026-08-16
fix(gemini): preserve tool-call history in prompt conversion

feiiiiii5:
## Description

`transform_to_gemini_prompt` (used by `update_gemini_kwargs` for every Gemini/VertexAI v2 request) silently dropped two things from reused tool-call conversations:

1. **`tool` messages were discarded entirely** — the model received no function results, so any tool-based workflow behaved as if its tools never ran.
2. **`assistant` messages with `tool_calls`** were converted to content-only parts, losing the `functionCall`.

This PR converts assistant `tool_calls` into Gemini `functionCall` parts (resolving string arguments to JSON where possible) and `tool` messages into `functionResponse` user-turn parts, mapping `tool_call_id` back to the function name from the preceding assistant message (falling back to the id when no match exists). System-message merging and plain message conversion are unchanged.

## Type of Change
- [x] Bug fix

## Test
- 2 new unit tests in `tests/v2/test_gemini_utils_deterministic.py`: tool_calls→functionCall + tool→functionResponse with id→name mapping; id fallback when no assistant tool call precedes.
- Pre-fix FAIL verified: with the previous implementation both new tests fail; post-fix the full file passes 21/21.
- `ruff check` and `ruff format --check` clean on both changed files.

## Checklist
- [x] Tests included
- [x] CHANGELOG.md entry added under [Unreleased]
- [x] Conventional commit format

## Related
Complements #2533 (protocol-field preservation in `merge_consecutive_messages`) — this fixes the same corruption class one step further down the Gemini conversion path.


feiiiiii5:
Closing this to keep the review queue clean, as it has not attracted maintainer review and I do not want it sitting in your backlog indefinitely.

The branch is preserved on my fork, so nothing is lost. If this is still useful to the project, I am happy to reopen it or rebase it onto current main at any time. Anyone is also welcome to pick the change up.

Thanks for maintaining the project.

PR #2536, opened 2026-08-16
fix(core): keep lazy modes registered while loading

### LINKED ISSUE #2535: from_litellm raises RegistryError when several threads make their first structured call concurrently
### Summary

`ModeRegistry.get_handlers` pops a lazy loader out of `_lazy_loaders` *before*
publishing the loaded handlers into `_handlers`. Between those two statements the
mode key is in neither dict, and `is_registered` checks exactly those two dicts.
A second thread calling `from_litellm` during that window is told its mode is not
registered, and `patch_v2` raises `RegistryError`.

It only affects the *first* structured call in a process — once the handlers are
published, the window is closed for good. So it looks like a rare flake, but it
is fully deterministic if a cold process goes straight into concurrency: 3 of 4
threads failed in the reproduction below.

### Version

- instructor 1.15.4
- Python 3.14.3, Windows 11
- via `instructor.from_litellm(litellm.completion, mode=instructor.Mode.JSON)`

### The window

`instructor/v2/core/registry.py`:

```python
    ...

        loader = self._lazy_loaders.pop(mode_key)   # (1) popped
        handlers = loader()                         # (2) slow: imports a module
        self._handlers[mode_key] = handlers         # (3) published only now
        return handlers
```

and

```python
def is_registered(self, provider, mode) -> bool:
    ...
    return mode_key in self._handlers or mode_key in self._lazy_loaders
```

Thread A is between (1) and (2). Thread B calls `from_litellm` → `patch_v2` →
`RegistryValidationMixin.validate_mode_registration` → `is_registered` → `False`
→ `RegistryError`.

A detail that confirms the diagnosis rather than merely fitting it: the error's
own "Available modes" dump lists `OPENAI` with `JSON_SCHEMA`, `MD_JSON`,
`PARALLEL_TOOLS`, `RESPONSES_TOOLS` and `TOOLS`, but **not** `JSON` — while
`ANTHROPIC`, `ANYSCALE`, `DATABRICKS`, `DEEPSEEK`, `GENAI` and `TOGETHER` all
show `JSON`. `list_modes()` reads both dicts, so the one mode missing from the
listing is precisely the one in flight, belonging to neither.

### Reproduction

Deterministic on this machine. Needs any OpenAI-compatible endpoint; the calls
themselves are irrelevant, only their concurrency and their coldness.

```python
import sys
import threading

import instructor
import litellm


class Tiny(BaseModel):
    answer: str


def one(results, index):
    try:
        client = instructor.from_litellm(litellm.completion, mode=instructor.Mode.JSON)
        client.create_with_completion(
            response_model=Tiny,
            max_retries=0,
            model="openai/<your-model>",
            api_base="<your-endpoint>",
            api_key="<key>",
            messages=[{"role": "user", "content": "Say the word ok."}],
        )
        results.append((index, None))
    except Exception as exc:
        results.append((index, f"{type(exc).__name__}: {str(exc)[:60]}"))


def main():
    warm = "warm" in sys.argv
    if warm:
        one([], -1)          # one single-threaded call first

    results = []
    threads = [threading.Thread(target=one, args=(results, i)) for i in range(4)]
    for t in threads:
        t.start()
    for t in threads:
        t.join()

    failures = [r for r in results if r[1]]
    print(f"warm={warm} threads=4 failures={len(failures)}")
    for index, error in sorted(failures):
        print(f"  thread {index}: {error}")


if __name__ == "__main__":
    main()
```

Results:

```
$ python race.py
warm=False threads=4 failures=3
  thread 1: RegistryError: Mode Mode.JSON is not registered for provider Provider.OPENAI...
  thread 2: RegistryError: Mode Mode.JSON is not registered for provider Provider.OPENAI...
  thread 3: RegistryError: Mode Mode.JSON is not registered for provider Provider.OPENAI...

$ python race.py warm
warm=True threads=4 failures=0
```

### Impact

Any application that fans out structured calls across threads from a cold
process. In ours (a multi-agent framework) a run dispatches N agents through a
thread pool, and each one's first call is structured — so a fresh process
reliably loses one or more agents to this, while the same code run after any
single structured call never fails.

It is also easy to misattribute: the exception names a *mode/provider
combination* that is in fact perfectly valid and works one second later, so it
reads like a configuration error rather than a race.

### Suggested fix

Either publish before popping, or hold a lock across the load:

```python
    with self._lock:
        if mode_key in self._handlers:          # another thread got there first
        loader = self._lazy_loaders.get(mode_key)
        if loader is not None:
            handlers = loader()
            self._handlers[mode_key] = handlers  # publish...
            self._lazy_loaders.pop(mode_key, None)  # ...then retire the loader
            return handlers
```

The ordering alone (publish, then pop) fixes `is_registered` for readers even
without a lock, since the key is then in one dict or the other at every instant.
A lock is still worth having so the loader runs once rather than once per racing
thread.

### Workaround for anyone hitting this now

Force the load once, single-threaded, before fanning out — no network needed:

```python

mode_registry.get_handlers(instructor.Provider.OPENAI, instructor.Mode.JSON)
```

nightcityblade:
## Describe your changes

Keeps each lazy mode registered until its handlers are published, closing the window where concurrent `from_litellm` validation could raise `RegistryError`. Adds deterministic regression coverage and an Unreleased changelog entry.

Fixes #2535

- [x] If it is a core feature, I have added thorough tests. (Regression coverage added for this bug fix.)
- [x] If it is a core feature, I have added documentation. (N/A; this is a bug fix, and the changelog is updated.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+
+- **v2 mode registry**: Keep lazily loaded modes registered while their handlers initialize, preventing concurrent first calls from failing spuriously. ([#2536](https://github.com/567-labs/instructor/pull/2536), [#2535](https://github.com/567-labs/instructor/issues/2535))
+
 ## [1.16.0] - 2026-08-09
 
 ### Added
--- a/instructor/v2/core/registry.py
+++ b/instructor/v2/core/registry.py
@@ -188,11 +188,9 @@ class ModeRegistry:
         if mode_key in self._handlers:
             return self._handlers[mode_key]
 
-        # Try lazy loading. Locked because the pop -> import -> set sequence
-        # below is not atomic: without the lock, a thread that loses the race
-        # to pop self._lazy_loaders[mode_key] would find it already gone and
-        # self._handlers[mode_key] not yet set, and raise KeyError even
-        # though the mode genuinely is registered (just still resolving).
+        # Try lazy loading. Locked because resolving and publishing a loader is
+        # not atomic; without the lock, concurrent first callers could run it
+        # repeatedly.
         with self._lazy_load_lock:
             # Re-check: another thread may have finished loading this key
             # while we were waiting for the lock.
@@ -200,9 +198,10 @@ class ModeRegistry:
                 return self._handlers[mode_key]
 
             if mode_key in self._lazy_loaders:
-                loader = self._lazy_loaders.pop(mode_key)
+                loader = self._lazy_loaders[mode_key]
                 handlers = loader()
                 self._handlers[mode_key] = handlers
+                self._lazy_loaders.pop(mode_key, None)
                 return handlers
 
         raise KeyError(
--- a/tests/v2/test_registry.py
+++ b/tests/v2/test_registry.py
@@ -186,3 +186,36 @@ def test_get_handlers_concurrent_first_access_does_not_race():
     )
     # The loader must run exactly once, not once per racing thread.
     assert load_count == 1
+
+
+def test_registry_stays_registered_while_lazy_loader_runs():
+    """The mode remains registered while its lazy loader is in progress."""
+    import threading
+
+    from instructor.v2.core.registry import ModeHandlers, ModeRegistry
+
+    registry = ModeRegistry()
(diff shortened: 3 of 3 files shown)


jxnl:
Closing as consolidated by https://github.com/567-labs/instructor/pull/2564, with added coverage for mode visibility during loading and retry after a transient loader failure. Thank you for the contribution.

PR #2537, opened 2026-08-17
fix(validation): run declared async validators in the async retry loop

### LINKED ISSUE #2528: Your code contains @async_model_validator and @async_field_validator, but it seems to be dead code
I want to validate cypher queries when the LLM generates parameters, but I want it to happen in the retry loop, if possible. The fact that you have async validator decorators indicates that you may be planning for async validation, but the code is not really connected to anything (apparently dead code).

I would like to be able to use the async validator decorators and have instructor call them, and have them work for async functions.

I am trying manual validation, but that makes me accept bad input from the LLM and proceeding with it in some way, then only sending a message back with an error to try again, except it's disconnected from the original request. I also tried pydantic-async-validation, but it doesn't work, either. Neither does just putting the regular validator decorators on an async function.

**Additional context**
These seem to be the same validators in pydantic-async-validation, and if instructor worked with those, that would be fine, as well.

tonycoder-hub:
Fixes #2528.

@async_model_validator and @async_field_validator only attached metadata and were never awaited. After an async client parses a non-streaming response, the retry loop now awaits those validators (including nested models). Failures become AsyncValidationError and go through the existing reask path. Sync clients log a warning instead of silently returning an unvalidated model.

tests/core/test_async_validation.py: 6 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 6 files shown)


tonycoder-hub:

PR #2538, opened 2026-08-17
fix: copy messages in Perplexity prepare_request so reask does not mutate the caller

tonycoder-hub:
## Summary
The #2417 aliasing fix missed Perplexity: it registers its own \`MD_JSON\` handler instead of using \`OPENAI_COMPAT_PROVIDERS\`. \`prepare_request\` shallow-copied \`kwargs\` but not \`messages\`, so \`reask_perplexity_json\` \`.extend()\` appended retry turns onto the caller's list.

Copy the list the same way \`OpenRouterJSONSchemaHandler\` already does. Added Perplexity \`MD_JSON\` to the existing \`test_messages_not_mutated\` parametrizations.

## Test plan
- [x] \`uv run pytest tests/v2/test_messages_not_mutated.py -q\` → 68 passed
- [x] New Perplexity cases fail on current main (caller list gains assistant + correction turns) and pass after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


tonycoder-hub:

PR #2539, opened 2026-08-17
fix(genai): copy contents in structured-outputs reask so retry does not mutate the caller

tonycoder-hub:
## Summary
`reask_genai_structured_outputs` did a shallow `kwargs.copy()` then appended the validation-error turn onto `kwargs["contents"]`. The list is shared, so every retry mutated the caller's `contents`. Missing `contents` also raised `KeyError`.

`reask_genai_tools` already copied the list. Pulled that guard into `_copy_kwargs_with_contents` and used it in both reask builders.

Follow-up to #2538 (Perplexity `messages`) on the GenAI structured-outputs path.

## Test plan
- [x] `uv run pytest tests/test_genai_reask.py tests/v2/test_genai_handlers_deterministic.py tests/coverage/test_genai_coverage.py tests/coverage/test_gemini_coverage.py -q` → 58 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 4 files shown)


tonycoder-hub:

PR #2540, opened 2026-08-17
fix(batch): parse OpenAI batch jobs when request_counts or metadata is null

tonycoder-hub:
## Summary
`BatchJobInfo.from_openai()` assumed `request_counts` and `metadata` were always dicts. The OpenAI SDK dumps them as `None` when optional fields are unset, so `dict.get(key, {})` returns `None` and listing batches raises `AttributeError`. One such job empties the whole CLI table.

Also parse the Batch API error list (`{"object":"list","data":[...]}`) instead of reading `type`/`message`/`code` off the outer object.

Distinct from #2529 (`from_anthropic` only).

## Test plan
- [x] `python3 -m pytest tests/coverage/test_batch_api_coverage.py -q` → 17 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


tonycoder-hub:

PR #2532, opened 2026-08-14
fix(bedrock): preserve nested caller config

wiliyam:
## Describe your changes

Bedrock request preparation shallow-copies the caller's keyword arguments, then merges normalized `temperature` and `top_k` values into existing nested `inferenceConfig` and `additionalModelRequestFields` dictionaries. That silently changes caller-owned configuration and can affect later retries or reused requests.

This copies the two nested dictionaries before merging and adds a regression test covering both paths.

Validation:

- `uv run pytest tests/v2/test_bedrock_handlers.py -q` (21 passed)
- `uv run ruff check instructor/v2/providers/bedrock/handlers.py tests/v2/test_bedrock_handlers.py`
- `uv run ruff format --check instructor/v2/providers/bedrock/handlers.py tests/v2/test_bedrock_handlers.py`
- `uv run ty check instructor/v2/providers/bedrock/handlers.py tests/v2/test_bedrock_handlers.py`

## Issue ticket number and link

No issue; independently discovered while reviewing Bedrock request normalization. Searches of open and closed issues and PRs found no report or competing fix for this mutation.

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [ ] If it is a core feature, I have added documentation.

PR #2530, opened 2026-08-14
fix(bedrock): reject unsupported structured-output constraints

wiliyam:
## Describe your changes

Bedrock native structured outputs reject JSON Schema numerical constraints (`minimum`, `maximum`, and `multipleOf`) and string-length constraints (`minLength` and `maxLength`). Pydantic emits these keywords for constrained fields, but Instructor currently forwards them to Bedrock, where the request fails remotely.

This change validates the generated schema recursively and raises a local `ConfigurationError` that identifies the unsupported keyword and its schema path. It covers both `JSON_SCHEMA` and `TOOLS_STRICT` modes with a regression test.

Validation:

- `uv run pytest tests/v2/test_bedrock_handlers.py -q` (22 passed)
- `uv run ruff check instructor/v2/providers/bedrock/handlers.py tests/v2/test_bedrock_handlers.py`
- `uv run ruff format --check instructor/v2/providers/bedrock/handlers.py tests/v2/test_bedrock_handlers.py`
- `uv run ty check instructor/v2/providers/bedrock/handlers.py tests/v2/test_bedrock_handlers.py`

## Issue ticket number and link

No issue; independently discovered regression in the recently added Bedrock native structured-output handling.

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [ ] If it is a core feature, I have added documentation.

PR #2541, opened 2026-08-17
fix(bedrock): return a toolResult for every toolUse on reask

tonycoder-hub:
## Summary
`reask_bedrock_tools` took the first `toolUseId` and broke. Bedrock Converse requires a `toolResult` for every `toolUse` in the previous assistant turn, so parallel tool calls made the retry a 400 and validation never reached the model.

Same bug Anthropic already fixed in #2485. Distinct from #2530 and #2532.

## Test plan
- [x] New test fails on current main (`['tool-1']` vs `['tool-1', 'tool-2']`) and passes after
- [x] `python3 -m pytest tests/coverage/test_bedrock_coverage.py -q` → 31 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


tonycoder-hub:

PR #2542, opened 2026-08-17
fix(genai): keep OpenAI-style text blocks in the system instruction

tonycoder-hub:
## Summary
`extract_genai_system_message` only kept plain-string items in list-form `content`. An OpenAI-style system message like `{"role":"system","content":[{"type":"text","text":"..."}]}` produced an empty string, so GenAI got no `system_instruction`. `convert_to_genai_messages` already skips `role == "system"`.

Read `{"type":"text","text":...}` blocks the same way the legacy Gemini path already does.

## Test plan
- [x] New assertions fail on current main (`''` vs the prompt) and pass after
- [x] `uv run pytest tests/v2/test_gemini_utils_deterministic.py tests/coverage/test_gemini_coverage.py -q` → 42 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


tonycoder-hub:

PR #2543, opened 2026-08-17
fix: merge JSON schema instruction into the first text content part

tonycoder-hub:
## Summary
JSON and MD_JSON handlers append the generated schema instruction to `messages[0]["content"][0]["text"]`. If the first content part is not text (an image) this raises `KeyError: 'text'`; an empty content list raises `IndexError`. Confirmed on 21 (provider, mode) handler pairs.

`merge_system_instruction()` now appends to the first text part, or adds a new text part when there is none. OpenAI, Mistral, Writer, and xAI handlers share that helper.

Distinct from #2542 (GenAI system text blocks).

## Test plan
- [x] `uv run pytest tests/v2/test_system_instruction_merge.py -q` → 12 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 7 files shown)


tonycoder-hub:

PR #2544, opened 2026-08-17
fix: do not skip the next object when streamed JSON objects abut

tonycoder-hub:
## Summary
`IterableBase.get_object` returned `s[i + 2:]` to skip the comma between array elements. When two streamed objects abut with nothing between them (`}{`), that skip eats the next object's opening brace and the item is silently dropped.

Consume the separator only when it is actually a comma. Existing `},{` / `}, {` cases stay the same.

## Test plan
- [x] `uv run pytest tests/v2/test_iterable_streaming.py -q` → 11 passed


tonycoder-hub:

PR #2547, opened 2026-08-17
fix: skip whitespace-only chunks in iterable task-list streaming

tonycoder-hub:
IterableBase.tasks_from_task_list_chunks (and the async variant) only skipped falsy chunks before json.loads. A streamed delta of just whitespace like "\n" is truthy, so it raised JSONDecodeError and killed the Mistral/VertexAI Mode.TOOLS iterable stream.

Whitespace-only chunks are now skipped. None is still skipped without calling strip(). Malformed chunks still raise JSONDecodeError; missing tasks still raises KeyError.

Tests: uv run pytest tests/coverage/test_dsl_iter_parallel_coverage.py -q — 15 passed. Distinct from #2544.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


tonycoder-hub:

PR #2548, opened 2026-08-17
fix: tolerate OpenAI completions with an empty choices list

tonycoder-hub:
_handle_incomplete_output and _extract_text_content indexed choices[0] as soon as a completion had a choices attribute. An empty choices list (OpenAI-compatible gateways, Azure content filtering) raised IndexError. The 1.16.0 changelog covered handler methods; the shared core helpers were left behind.

Empty choices is now treated as no usable text. Anthropic stop_reason still runs afterwards.

Tests: uv run pytest tests/coverage/test_core_schema_coverage.py tests/processing/test_json_extraction.py -q — 46 passed. Distinct from #2537-#2547.

PR #2549, opened 2026-08-17
fix: extract Bedrock text from the first block that has it

tonycoder-hub:
_extract_text_content took the first Converse content block and returned block.get("text"). A leading toolUse or reasoningContent block returned None (annotated -> str), and callers then TypeError in json.loads. #2548 fixed the OpenAI empty-choices path in the same helper; this is the Bedrock branch.

Scan for the first string text, else "".

Tests: uv run pytest tests/coverage/test_core_schema_coverage.py tests/processing/test_json_extraction.py tests/coverage/test_bedrock_coverage.py tests/v2/test_bedrock_handlers.py -q — 98 passed. Distinct from #2537-#2548.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


tonycoder-hub:

PR #2545, opened 2026-08-17
fix(gemini): stop MD_JSON prepare_request from mutating caller messages

tonycoder-hub:
Issue #2417 was fixed for OpenAI-compat, Cohere, Mistral, OpenRouter, Writer, and xAI. Gemini MD_JSON was missed because the sweep passed model= and Gemini raises ConfigurationError before reaching the injection code.

handle_gemini_json inserted the schema instruction into the caller list, and get_message_content returned the caller content list so transform_to_gemini_prompt spliced the system prompt into it. Both now copy first.

Tests: uv run pytest tests/v2/test_messages_not_mutated.py tests/coverage/test_gemini_coverage.py tests/v2/test_gemini_utils_deterministic.py tests/v2/test_gemini_json_messages.py tests/processing/test_message_processing.py tests/processing/test_response_model_conversion.py -q -> 149 passed.

Distinct from #2538/#2539/#2542/#2543.

PR #2551, opened 2026-08-18
fix(gemini): preserve caller safety settings

tandede:
## Describe your changes

`update_gemini_kwargs` now copies a caller-provided `safety_settings` mapping before merging the Gemini default thresholds.

Previously, the function made only a shallow copy of the request kwargs, then reused and updated the original nested mapping. A request could therefore silently lower thresholds or add categories in application-owned configuration, so reusing the same dictionary produced stateful behavior outside Instructor.

The regression test verifies both sides of the contract: defaults are still applied to the outgoing request, while the original mapping remains unchanged and is not aliased by the result.

N/A — independently discovered while auditing Gemini request preparation. This is separate from #2545, which covers caller-owned `messages`.

## Validation

- `pytest tests/v2/test_gemini_utils_deterministic.py -q` — 19 passed
- v2, Gemini coverage, and processing regression suite — 1734 passed, 168 skipped
- non-LLM code suite excluding pre-existing docs-format baseline failures — 2969 passed, 199 skipped, 27 deselected
- full Ruff check and format check — passed
- `ty` checks for `instructor/` and `tests` — passed

- [x] I have added a focused regression test
- [x] Documentation is not required because this restores input immutability without changing the public API

### DIFF
--- a/instructor/v2/providers/gemini/utils.py
+++ b/instructor/v2/providers/gemini/utils.py
@@ -339,7 +339,7 @@ def update_gemini_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
         result.setdefault("safety_settings", {})
         return result
 
-    safety_settings = result.get("safety_settings", {})
+    safety_settings = result.get("safety_settings", {}).copy()
     result["safety_settings"] = safety_settings
 
     for category, threshold in default_safety_thresholds.items():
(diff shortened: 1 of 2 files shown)


jxnl:
Closing as consolidated by #2566, which copies caller-provided Gemini safety mappings before applying defaults. Thank you for the contribution.

PR #2559, opened 2026-08-24
fix: run async validators in v2 retry path

### LINKED ISSUE #2528: Your code contains @async_model_validator and @async_field_validator, but it seems to be dead code

mikemikimike:
## Summary

Closes #2528.

Async validators are now executed in the async v2 retry path after response parsing, including nested models and validation context. Validator failures are converted to `AsyncValidationError` so the existing retry/reask handling remains in control.

This change also keeps the v2 validation package imports lazy. Accessing async validation support no longer eagerly imports the OpenAI SDK; the existing `Validator`, `llm_validator`, and `openai_moderation` exports remain available and are loaded when accessed.

## Compatibility

The public validation exports and async retry behavior remain backward compatible. The OpenAI SDK is still loaded when an OpenAI-backed validator or provider is explicitly accessed.

## Validation

- `.venv\\Scripts\\python.exe -m pytest tests/test_import_lazy_openai.py tests/v2/test_async_validators_runtime.py -q` — 19 passed.
- `.venv\\Scripts\\python.exe -m ruff format --check instructor/v2/validation/__init__.py tests/v2/test_async_validators_runtime.py` — passed.
- `.venv\\Scripts\\python.exe -m ruff check --target-version py39 --select B,F401,E722,ARG,UP --ignore B006,B018 instructor/v2/validation/__init__.py tests/v2/test_async_validators_runtime.py` — passed.
- `.venv\\Scripts\\python.exe -m ty check --error-on-warning instructor/v2/validation/__init__.py tests/v2/test_async_validators_runtime.py` — passed.
- `git diff --check` — passed.
- GitHub Actions run [33098244274](https://github.com/567-labs/instructor/actions/runs/33098244274) — all required checks passed, including full coverage, all supported Python versions, provider tests, Ruff, and type-check.

The full repository workflow and pre-commit were not run locally; the remote workflow above completed the full CI validation.


mikemikimike:
CI note: the focused runtime tests, Ruff, and type-check passed. The failing Core Tests / Full Coverage jobs fail only at the pre-existing `tests/test_import_lazy_openai.py::test_instructor_class_access_does_not_load_openai` assertion (OpenAI modules eagerly loaded), outside the async-validator changes; Python 3.9 coverage reports the same unrelated failure.

PR #2522, opened 2026-08-11
fix: reject empty provider model names

wiliyam:
## Describe your changes

Reject provider strings with an empty provider or model component before constructing an SDK client.

Previously, `from_provider("openai/")` reached `openai.OpenAI` and failed on unrelated configuration such as a missing API key. The documented `provider/model-name` contract now produces the same actionable `ConfigurationError` used for strings without a provider prefix.

Adds a focused regression test for the empty-model case.

## Issue ticket number and link

Not applicable — this is a small, independently discovered validation fix, and no duplicate issue or PR was found.

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [ ] If it is a core feature, I have added documentation. (No documentation change: this enforces the existing documented format.)

Tests:

- `uv run pytest tests/v2/test_auto_client_deterministic.py -q`
- `uv run ruff check instructor/v2/auto_client.py tests/v2/test_auto_client_deterministic.py`
- `uv run ty check instructor/v2/auto_client.py tests/v2/test_auto_client_deterministic.py`

PR #2564, opened 2026-08-27
fix(core): consolidate 1.16.1 reliability fixes

### LINKED ISSUE #2526: merge_consecutive_messages drops tool-call fields from conversation history

jxnl:
## Summary

- prepare the source, lockfile, and changelog as the `1.16.1` maintenance line
- reject malformed `from_provider()` model strings before client construction
- include hoisted system prompts in sync and async cache keys
- normalize parameterized and case-insensitive remote media types
- preserve tool-call protocol fields and caller-owned content while merging messages
- include canceled and expired Anthropic requests in batch totals
- keep lazy registry entries visible and retryable while handlers load
- add the missing CLI cost-table annotation from #2521 without its stray commits

## Consolidated work

Supersedes #2521, #2522, #2524, #2525, #2527, #2529, and #2536.

Closes #2526 and #2535.

Original commits and authors are retained where the branches were clean. The follow-up commits resolve changelog overlap and strengthen boundary coverage: both empty provider components, list-content input immutability, the existing Anthropic count fixture, `list_modes()` visibility during lazy loading, retry after a transient loader failure, and the missing media-type guard.

## Validation

- release validator reports `1.16.1`
- `273 passed` across the affected cache, message, multimodal, batch, auto-client, and registry suites
- Ruff check passed on `instructor`, `examples`, and `tests`
- Ruff format check passed on `instructor`, `examples`, and `tests`
- strict `ty` source and changed-test checks passed
- `uv lock --check` passed
- commit preflight passed dependency installation, lock, Ruff, format, and type checks
- `git diff --check` passed

## Intentionally separate from 1.16.1

- OpenAI v3 support in #2558 needs bounded dependency and real sync/async/stream/retry compatibility coverage
- async validators in #2559 need correct return-value semantics and a clear sync-client failure contract
- #2534 and #2409 depend on provider protocol or SDK-internal decisions and need dedicated review
- broad provider additions, architecture changes, examples, and dependency batches remain outside this patch release

Small provider-correctness and documentation candidates will be evaluated independently after this PR lands; none are required for this consolidation's correctness.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -21,10 +21,13 @@ jobs:
         uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
         with:
           ref: ${{ github.event.release.tag_name }}
+          persist-credentials: false
       - name: Install uv
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
-          enable-cache: true
+          # Publishing runs with a long-lived registry credential. Avoid
+          # restoring a cache populated by less-privileged workflows.
+          enable-cache: false
           python-version: "3.11"
       - name: Download tested release artifacts
         env:
--- a/.github/workflows/scheduled-release.yml
+++ b/.github/workflows/scheduled-release.yml
@@ -46,6 +46,7 @@ jobs:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
         with:
           fetch-depth: 0
+          persist-credentials: false
       - name: Install uv
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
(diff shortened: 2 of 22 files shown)


jxnl:
All required checks are green for exact head `896cbb4fb9041b27791cfde1f82c1dc287ddebd2`, including full coverage, Python 3.9-3.13, provider suites, release artifacts, wheel smokes, Ruff, `ty`, and Workers. Auto-merge is armed. @vm or @jxnl-oai, could one of you provide the independent approval required by branch protection? This is the bounded `1.16.1` reliability consolidation; follow-on docs and provider fixes are isolated in draft PRs #2565 and #2566.

jxnl:
Security review update for exact head `532db517`:

- Reviewed all changed runtime, provider-adjacent, docs, lockfile, and release paths for cache isolation, caller-input mutation, URL/media handling, schema validation, credential handling, and workflow privilege boundaries; no material vulnerability or privilege regression was found in the release diff.
- Bandit 1.9.4 reports no medium-or-higher findings in changed production Python files.
- Hardened release workflows by disabling checkout credential persistence, disabling cache restore in the credentialed PyPI job, and moving release outputs into step environment variables before shell use.
- Zizmor now reports zero low/medium/high findings for the two release workflows; its remaining informational recommendation is PyPI Trusted Publishing, which requires external PyPI project configuration.
- `tests/test_prepare_release.py`: 8 passed; release metadata validator: 1.16.1; `git diff --check`: clean.

The broader dependency audit and disabled repository scanners are recorded as follow-up security/compatibility work rather than folded into this patch as an unreviewed Python 3.9-breaking dependency batch. Merge remains gated on every GitHub check for this new exact head.

PR #2568, opened 2026-08-27
fix: preserve optional TypedDict response fields

mikemikimike:
## Summary

Fix TypedDict response-model conversion so optional keys remain optional.

`prepare_response_model` now resolves TypedDict annotations with extras, preserves `total=False` and `NotRequired` metadata, and unwraps `Required`/`NotRequired` before creating the Pydantic model. This prevents valid responses that omit optional fields from being rejected.

## Tests

- `python -m pytest tests/v2/test_response_schema_compat.py -q` (15 passed)
- `ruff check instructor/v2/core/response_model.py tests/v2/test_response_schema_compat.py`
- `ruff format instructor/v2/core/response_model.py tests/v2/test_response_schema_compat.py`
- `git diff --check`

The repository's frozen environment could not build its pinned `pydantic-core` on Python 3.14; the focused tests were run successfully with the system Python environment.

### DIFF
--- a/instructor/v2/core/response_model.py
+++ b/instructor/v2/core/response_model.py
@@ -5,7 +5,16 @@ from __future__ import annotations
 import inspect
 import sys
 from collections.abc import Iterable
-from typing import Any, Callable, TypeVar, Union, cast, get_args, get_origin
+from typing import (
+    Any,
+    Callable,
+    TypeVar,
+    Union,
+    cast,
+    get_args,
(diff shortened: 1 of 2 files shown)


jxnl:
Closing as consolidated by #2566. The canonical patch preserves total=False, Required, and NotRequired semantics through one shared conversion path and also fixes iterable TypedDict response models. Thank you for the contribution.

PR #2577, opened 2026-08-28
fix(anthropic): read local PDF sources from disk instead of over HTTP

Nikhiljai03:
## Describe your changes

`pdf_to_anthropic` encoded any PDF whose `data` was unset by calling
`requests.get(str(pdf.source))`.

That branch is only reachable for **non-HTTP** sources, because a remote `http(s)`
URL returns earlier as a URL document block:

```python
if isinstance(pdf.source, str) and pdf.source.startswith(("http://", "https://")) and not pdf.data:
    return {"type": "document", "source": {"type": "url", "url": pdf.source}}
if not pdf.data:
    pdf.data = requests.get(str(pdf.source), timeout=30).content   # only non-HTTP sources get here
```

So every source that could actually reach the `requests.get` call (a local path,
`gs://`, `s3://`) is one `requests` cannot resolve. It raised `MissingSchema` /
`InvalidSchema` instead of encoding the document.

`PDF.source` is documented as *"URL, file path, or base64 data of the PDF"*, and
constructing `PDF(source=Path(...))` directly leaves `data` unset — only the
`from_path` / `autodetect` constructors pre-populate it. Building a PDF from a
local path and sending it to Anthropic therefore always failed, even with a
perfectly readable file on disk:

```python
pdf = PDF(source=Path("invoice.pdf"), media_type="application/pdf")
pdf_to_anthropic(pdf)
# requests.exceptions.InvalidSchema: No connection adapters were found for 'invoice.pdf'
```

### Changes

- Read local files from disk when `data` is unset.
- Raise the same clear `ValueError("PDF data is missing for base64 encoding.")`
  that `pdf_to_openai` already raises when no usable source is present, instead
  of surfacing an opaque `requests` error.
- This removes the last raw `requests` call left in a provider multimodal module
  after remote media fetching was centralized in #2569 (which updated the
  `openai`, `genai`, and `bedrock` multimodal paths but not `anthropic`).

`image_to_anthropic` is deliberately left alone — it has a different shape and no
fetch call, so it is out of scope here.

### Tests

`test_anthropic_multimodal_encodes_remote_image_and_local_pdf` asserted the broken
behaviour by mocking `requests.get`, and separately failed on Windows because it
compared a `Path` against a hardcoded POSIX string:

```
AssertionError: assert [('\tmp\example.pdf', 30)] == [('/tmp/example.pdf', 30)]
```

It now writes a real file to `tmp_path`, which fixes the Windows failure too. Two
regression tests were added:

- `test_anthropic_pdf_reads_local_path_without_network` — a `PDF` built from a
  local path encodes from disk (uses the real `PDF` model, not a stub).
- `test_anthropic_pdf_without_data_raises_clear_error` — an unusable source
  raises the clear `ValueError`.

All three fail on `main` and pass with this change.

No existing issue — found while reviewing provider multimodal code after #2569.

- [ ] If it is a core feature, I have added documentation. *(bug fix, no doc change needed)*

### Verification

- `pytest tests/coverage/test_anthropic_support_coverage.py` — 13 passed
- `pytest tests/ --ignore=tests/llm --ignore=tests/docs` — 2979 passed, 195 skipped.
  The 6 remaining failures are pre-existing `tests/providers/test_auto_client.py`
  cases that require live `GROQ_API_KEY` / `WRITER_API_KEY` / `CEREBRAS_API_KEY`.
- `ruff check` and `ruff format --check` pass
- `ty check` passes
- `git diff --check` clean

### DIFF
--- a/instructor/v2/providers/anthropic/multimodal.py
+++ b/instructor/v2/providers/anthropic/multimodal.py
@@ -3,9 +3,8 @@
 from __future__ import annotations
 
 import base64
+from pathlib import Path
 from typing import Any
-
-import requests
 
 
 def image_to_anthropic(image: Any) -> dict[str, Any]:
@@ -25,6 +24,26 @@ def image_to_anthropic(image: Any) -> dict[str, Any]:
     }
 
 
+def _read_local_pdf(source: Any) -> bytes | None:
+    """Return the bytes of ``source`` when it points at a readable local file.
+
+    Returns ``None`` when the source is not a usable filesystem path — for
+    example an inline base64 payload or a non-HTTP remote URI — so callers can
+    raise a clear error instead of failing deeper in the stack.
+    """
+    try:
+        path = Path(source)
+    except TypeError:
+        return None
+    try:
+        if not path.is_file():
+            return None
+        return path.read_bytes()
+    except OSError:
+        # Inline base64 payloads and remote URIs are not valid path names.
+        return None
+
+
 def pdf_to_anthropic(pdf: Any) -> dict[str, Any]:
     if (
         isinstance(pdf.source, str)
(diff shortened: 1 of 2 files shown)


jxnl:

PR #2586, opened 2026-09-02
fix(cache): include sampling parameters (temperature, seed, ...) in the cache key

### LINKED ISSUE #2585: AutoCache/DiskCache key ignores temperature/seed/etc. — calls that only differ in sampling params silently return a stale cached response
## What

`instructor.cache.make_cache_key` — the function documented in
[caching.md](docs/concepts/caching.md) as computing the "deterministic" key
for every `cache=` call — only hashes `model`, `messages`/`contents`,
`mode`, `system`, and `response_model`'s schema:

https://github.com/567-labs/instructor/blob/main/instructor/cache/__init__.py#L145-L189

It never inspects `temperature`, `top_p`, `seed`, `max_tokens`, or any other
sampling parameter, even though these are present in the same `kwargs` dict
(`new_kwargs`) that the two call sites in `patch.py` already have in scope:

https://github.com/567-labs/instructor/blob/main/instructor/v2/core/patch.py#L251-L259

Two `client.create(...)` calls with identical messages/model but different
`temperature` (or `seed`, `top_p`, ...) produce the *same* cache key. If the
first call is cached, the second silently returns the first call's answer
instead of hitting the provider — no error, no warning.

This is a normal, expected use pattern: sharing one `cache=` instance across
calls that vary `temperature` per-request (e.g. deterministic calls at
`temperature=0` alongside exploratory calls at higher temperature) is
exactly the kind of thing `cache=` is documented to support transparently.

## Reproduction

```python
from instructor.cache import make_cache_key, AutoCache

class Answer(BaseModel):
    value: int

messages = [{"role": "user", "content": "Roll a random number between 1 and 100"}]

key_a = make_cache_key(messages=messages, model="gpt-4o", response_model=Answer, mode="tool_call")
key_b = make_cache_key(messages=messages, model="gpt-4o", response_model=Answer, mode="tool_call")
# key_a == key_b always, regardless of temperature/seed/etc. -- the function
# has no parameter for them at all, so callers can't distinguish these calls
# no matter what they pass to client.create(...).

cache = AutoCache()
cache.set(key_a, Answer(value=7))
cache.get(key_b).value  # 7 -- a later, differently-sampled call reuses this
```

Confirmed end-to-end through the real client (not just the key function) via
`instructor.from_litellm` with a fake completion function that returns a
different value on each call: a `temperature=1.9` call after a cached
`temperature=0.0` call receives the `temperature=0.0` response without the
provider ever being invoked a second time.

## Expected behavior

Two calls that differ only in sampling parameters must not collide in the
cache.

## Fix

Extend `make_cache_key` with an optional `generation_kwargs` component and
populate it at the `patch.py` cache lookup/store sites from `new_kwargs`.
PR incoming.

Yashwanth-Kumar-Kotla:
## Describe your changes

Fixes #2585.

`make_cache_key` never looked at `temperature`, `top_p`, `seed`,
`max_tokens`, or any other sampling parameter, even though `patch.py`'s
cache lookup/store sites have them available in `new_kwargs`. Two calls
identical except for sampling parameters produced the same cache key, so
the second call silently received the first call's cached response instead
of hitting the provider.

Added `extract_generation_kwargs()` in `instructor/cache/__init__.py`,
which pulls a fixed allowlist of sampling-affecting keys (`temperature`,
`top_p`, `top_k`, `seed`, `max_tokens`, `max_completion_tokens`,
`max_output_tokens`, `frequency_penalty`, `presence_penalty`, `n`, `stop`,
`stop_sequences`, `logit_bias`) out of a kwargs dict, dropping anything
unset. `make_cache_key` gained an optional `generation_kwargs` parameter
that folds these into the hashed payload when present, so keys are
unchanged when no sampling parameters are passed. Wired
`generation_kwargs=extract_generation_kwargs(new_kwargs)` into all four
`make_cache_key` call sites (sync lookup/store, async lookup/store) in
`patch.py`.

Also updated the "Cache-key design" table in `docs/concepts/caching.md`.

Fixes #2585

## Proof of fix

Added:
- `tests/cache/test_cache_key.py`: unit tests on `make_cache_key`/`extract_generation_kwargs` (temperature/seed change the key; no-generation-kwargs keys are unchanged; irrelevant/unset keys are dropped).
- `tests/cache/test_cache_integration.py::test_cache_key_accounts_for_temperature`: end-to-end through the real client (`instructor.from_litellm`) — a `temperature=1.9` call after a cached `temperature=0.0` call must re-hit the provider.

Verified the regression tests actually catch the bug: reverted just the fix (kept the tests) and 5 of the 14 cache tests failed, including the integration test with `assert call_counter["n"] == 2` failing as `1 == 2` (the second call never reached the provider). Restored the fix — all 14 pass again.

```
tests/cache/test_cache_key.py::test_cache_key_changes_on_temperature_change PASSED
tests/cache/test_cache_key.py::test_cache_key_changes_on_seed_change PASSED
tests/cache/test_cache_key.py::test_cache_key_unchanged_when_no_generation_kwargs PASSED
tests/cache/test_cache_key.py::test_extract_generation_kwargs_drops_unset_and_irrelevant_keys PASSED
tests/cache/test_cache_integration.py::test_cache_key_accounts_for_temperature PASSED
============================== 14 passed in 0.59s ===============================
```

Ran the broader offline test suite (`tests/core`, `tests/dsl`, `tests/v2`, `tests/cache`, `tests/processing`): 2018 passed, 178 skipped, 6 failed. Confirmed all 6 failures are pre-existing on unmodified `main` too (missing optional `vertexai`/`mistralai`/`google-genai` SDKs in this environment), unrelated to this change.

`ruff check` and `ruff format --check` pass on all changed files.

## Scope

Two-file logic change (`instructor/cache/__init__.py`, `instructor/v2/core/patch.py`) plus tests and a one-line doc update. No other caching behavior changed.

### DIFF
--- a/docs/concepts/caching.md
+++ b/docs/concepts/caching.md
@@ -85,6 +85,7 @@ Components that influence the key:
 | `messages` / `contents`     | The full chat history is hashed              |
 | `mode`                      | JSON vs. TOOLS vs. RESPONSES changes formatting |
 | `response_model` schema     | The entire `model_json_schema()` is included so **any** change in field names, types or *descriptions* busts the cache automatically |
+| sampling parameters (`temperature`, `top_p`, `seed`, `max_tokens`, …) | Calls that only differ in these must not collide — otherwise a higher-`temperature` call would silently be served a lower-`temperature` call's cached answer |
 
 The function returns a SHA-256 hex digest; its length is constant regardless
 of prompt size, so it is safe to use as a Redis key, file path, etc.
--- a/instructor/cache/__init__.py
+++ b/instructor/cache/__init__.py
@@ -37,6 +37,7 @@ __all__ = [
     "AutoCache",
     "DiskCache",
     "make_cache_key",
+    "extract_generation_kwargs",
 ]
 
 
@@ -141,6 +142,36 @@ class DiskCache(BaseCache):
 # Cache-key helper
 # -------------------------------------------------------------------------
 
+# Sampling/generation parameters that affect the model's output and must
+# therefore be part of the cache key -- otherwise two calls that differ only
+# in, say, ``temperature`` collide and the second call silently returns the
+# first call's cached (and differently-sampled) response.
+_GENERATION_PARAM_KEYS: tuple[str, ...] = (
+    "temperature",
+    "top_p",
+    "top_k",
+    "seed",
+    "max_tokens",
+    "max_completion_tokens",
+    "max_output_tokens",
+    "frequency_penalty",
+    "presence_penalty",
+    "n",
+    "stop",
+    "stop_sequences",
+    "logit_bias",
+)
+
+
+def extract_generation_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
+    """Pull the sampling parameters relevant to :func:`make_cache_key` out of
+    a provider call's kwargs dict, dropping anything unset."""
+    return {
+        key: kwargs[key]
+        for key in _GENERATION_PARAM_KEYS
+        if key in kwargs and kwargs[key] is not None
+    }
+
 
 def make_cache_key(
     *,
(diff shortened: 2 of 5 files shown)


jxnl:

PR #2587, opened 2026-09-02
docs: fix blog code blocks that fail the docs lint

hassaanch23:
## Why

`tests/docs/test_posts.py` lints every code block under `docs/blog/posts` with black via pytest-examples. **34 of them fail on `main`**, which is why `weekly-tests` and the `release` job's test step are both red:

```
34 failed, 228 passed
```

This fixes **16 of the 34**, taking it to `18 failed, 244 passed`.

## Two were genuinely broken Python

A reader copying either of these gets a `SyntaxError`:

**`bad-schemas-could-break-llms.md`** — missing the `c`:

```diff
- lass AnswerWithIntermediateCalculations(BaseModel):
+class AnswerWithIntermediateCalculations(BaseModel):
```

**`jinja-proposal.md`** — missing comma, so the two strings concatenate and the following dict key is a syntax error:

```diff
-        "topic": "jason and now his phone number is 123-456-7890"
+        "topic": "jason and now his phone number is 123-456-7890",
         "banned_words": ["jason"],
```

## The rest is the formatter's own output

Once those two parse, black can format them. Everything else here is what pytest-examples' formatter produces — joined lines that now fit the width, and blank lines black adds or removes between definitions. I ran only the `format` step, not `--update-examples`: the latter also calls `run_print_update`, which executes the examples and fails without provider API keys (it takes the suite to 240 failures).

In `jinja-proposal.md` the formatter also restored `#> ` on an output comment that had drifted to `# >`. `#> ` — with the trailing space — is pytest-examples' `comment_prefix`; `# >` isn't matched as an output marker, so this is a repair.

## One file deliberately left alone

`introducing-structured-outputs.md` is excluded. Its block has a bare `#> ` marker standing for an empty printed line, and that form can't survive formatting: black strips the trailing space, the result no longer matches `comment_prefix`, and black then rewrites it to `# >` — which breaks the marker *and* leaves it inconsistent with the `#> ...` lines directly beneath it:

```
            # >          <- rewritten
            #> {"
            #> {"name
```

Writing `# >` into the docs to satisfy the linter seemed worse than leaving the block failing, so it's in the group below.

## The remaining 18 need a decision I didn't want to make alone

These are published posts, and each needs a judgement about intent:

- **empty ` ```python ` blocks** where install instructions appear to have been lost — `multimodal-gemini.md` and `generating-pdf-citations.md` both introduce a snippet with prose ("let's set up our environment with the necessary libraries:") and then show an empty block
- **pseudo-code fenced as python** — e.g. `client.create(..., response_model=type[T]) -> T` in `version-1.md`, a type signature rather than a statement
- **fragments that were never standalone modules** — e.g. a bare `@field_validator` in `chain-of-density.md`

The options are to complete the snippet, re-fence as ` ```text ` (only fences starting with `py`/`{.py` are collected by `find_examples`), or exclude older posts from the lint entirely. Happy to follow up with whichever you'd prefer — I didn't want to rewrite published posts on my own judgement.

## Verification

```console
$ pytest tests/docs/test_posts.py -q
before:  34 failed, 228 passed
after:   18 failed, 244 passed
```

No source code touched — docs only.

### DIFF
--- a/docs/blog/posts/anthropic-prompt-caching.md
+++ b/docs/blog/posts/anthropic-prompt-caching.md
@@ -188,7 +188,6 @@ Let's first initialize our Anthropic client, this will be the same as what we've
 from instructor import Instructor, Mode, patch
 from anthropic import Anthropic
 
-
 client = Instructor(
     client=Anthropic(),
     create=patch(
--- a/docs/blog/posts/anthropic-web-search-structured.md
+++ b/docs/blog/posts/anthropic-web-search-structured.md
@@ -38,7 +38,6 @@ Now, let's define our Pydantic model for the response:
 import instructor
 from pydantic import BaseModel
 
-
 # Noticed thhat we use JSON not TOOLS mode
 client = instructor.from_provider(
     "anthropic/claude-3-7-sonnet-latest",
--- a/docs/blog/posts/bad-schemas-could-break-llms.md
+++ b/docs/blog/posts/bad-schemas-could-break-llms.md
@@ -169,17 +169,20 @@ class AssumptionBasedAnswer(BaseModel):
     logic_flow: str
     answer: int
 
+
 class ErrorAwareCalculation(BaseModel):
     key_steps: list[str]
     potential_pitfalls: list[str]
     intermediate_results: list[str]
     answer: int
 
- lass AnswerWithIntermediateCalculations(BaseModel):
+
+class AnswerWithIntermediateCalculations(BaseModel):
     assumptions: list[str]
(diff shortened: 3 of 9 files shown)


jxnl:
Consolidated into #2597 for Instructor 1.17.0: https://github.com/567-labs/instructor/pull/2597 . The documentation change is included (overlapping typo edits are deduplicated), with contributor attribution retained where applicable. The combined documentation suite passes 610 checks with 89 existing skips. Closing this source PR as superseded; the consolidation PR remains draft pending full release validation and review.

PR #2588, opened 2026-09-02
fix(validation): run declared async validators in the v2 retry path

Ananthr16:
## Problem

`@async_field_validator` and `@async_model_validator` declared on a `response_model` are silently never run. An LLM response that violates an async validator's rule is returned to the caller as a normal, "validated" model — no exception, no retry/reask, nothing. See #2528.

## Root cause

`async_field_validator`/`async_model_validator` (`instructor/v2/validation/async_validators.py`) only `setattr` metadata onto the decorated function. Nothing in the retry path (`instructor/v2/core/retry.py`) ever reads `ASYNC_VALIDATOR_KEY`/`ASYNC_MODEL_VALIDATOR_KEY` or awaits the validators. `AsyncValidationError` already existed and was already registered as a retryable error, so the reask machinery was ready — it just never got triggered.

Reproduced directly against `retry_async_v2` with a fake completion (bypassing the network):

```python
class Answer(BaseModel):
    email: str

    @async_field_validator("email")
    async def must_contain_at(cls, value: str) -> str:
        if "@" not in value:
            raise ValueError(f"{value!r} is not a valid email address")
        return value

# fake completion always returns {"email": "not-an-email"}
result = await retry_async_v2(func=fake_openai_create, response_model=Answer, ...)
# before fix: Answer(email='not-an-email'), validator invoked 0 times
```

## What changed

- **`instructor/v2/validation/async_validators.py`**: added `run_async_validators()`, which recursively awaits declared field/model validators over an already-parsed model tree (depth-first through nested `BaseModel`s inside fields, lists, tuples, and dicts), aggregating every failure into one `AsyncValidationError`. Added `model_declares_async_validators()` for a fast up-front check.
- **`instructor/v2/core/retry.py`**:
  - `retry_async_v2` now runs `run_async_validators` on the parsed result (when it's a `BaseModel` or `list`, and not a streaming response) right after a successful parse, before finalizing. A failure raises `AsyncValidationError`, which is already in `_RETRYABLE_PARSE_ERRORS`, so it flows straight into the existing reask path.
  - `retry_sync_v2` now raises `ConfigurationError` immediately if `response_model` declares async validators, instead of silently returning an unvalidated model — a sync client can never `await` them. (This was an open question in the issue thread; I went with "fail loudly," which is what the two commenters there converged on too.)
  - Both imports are **local to the functions**, not module-level. `instructor.v2.validation` eagerly imports the OpenAI SDK (via `llm_validators.py`'s `from openai import OpenAI`), and `retry.py` is on the import path of `instructor.Instructor`/`instructor.from_provider`. A module-level import would have broken the lazy-import guarantee covered by `tests/test_import_lazy_openai.py` — this is what tripped up the previous attempt at this issue (#2559). Confirmed `tests/test_import_lazy_openai.py` still passes.
- **`instructor/v2/core/errors.py`**: gave `AsyncValidationError` a real `__init__` (it previously had `errors: list[ValueError]` as a bare type annotation with no way to actually set it).
- **`tests/v2/test_async_validators_execution.py`** (new): end-to-end tests against the real retry functions with fake completions — success + value transformation, failure → reask → success, retries exhausted → `InstructorRetryException`, nested-model validation and error aggregation, sync client fails fast, sync client unaffected when no async validators are declared.
- **`CHANGELOG.md`**: entry under `[Unreleased]`.

### Explicitly out of scope

Streaming responses and the `Iterable[Model]`/parallel-tool-calls path (`IterableBase`) are not covered — validators only run against non-streaming `BaseModel`/`list` results. The issue's own repro and discussion were about the single-response-model case; extending to streaming/parallel tool calls raises its own design questions (validating partial state mid-stream) that felt separate from this fix.

## How it was verified

- Reproduced the bug against `main` first (validator invoked 0 times, invalid value returned), then confirmed the fix turns that into a raised `InstructorRetryException` wrapping `AsyncValidationError`.
- `uv run ruff check instructor examples tests` — clean.
- `uv run ruff format instructor examples tests` — clean (already applied).
- `uv run ty check` — 52 diagnostics, identical count and content to `main` before this change (0 new).
- `uv run pytest tests/ -k 'not llm and not openai'` (all extras installed): **3175 passed, 35 failed**. Ran the identical command against unmodified `main` for comparison: also 35 failed with the exact same test IDs (pre-existing doc/blog-example tests that need network/API keys) and 3170 passed. So: 0 regressions, and the 5-test delta is exactly the new regression tests going from failing (against unpatched code) to passing.
- `tests/test_import_lazy_openai.py` — all 11 pass, confirming the lazy-openai-import invariant holds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Async validators**: `@async_field_validator` and `@async_model_validator` are now actually awaited during response parsing (including on nested models) instead of being silently ignored; a sync client given a `response_model` that declares async validators now raises a clear `ConfigurationError` instead of returning an unvalidated model. ([#2528](https://github.com/567-labs/instructor/issues/2528))
+
 ## [1.16.1] - 2026-08-28
 
 ### Changed
--- a/instructor/v2/core/errors.py
+++ b/instructor/v2/core/errors.py
@@ -506,7 +506,15 @@ class AsyncValidationError(ValueError, InstructorError):
         ```
     """
 
-    errors: list[ValueError]
+    def __init__(
+        self,
+        message: str,
+        *args: Any,
+        errors: list[ValueError] | None = None,
+        **kwargs: Any,
+    ):
+        self.errors = errors if errors is not None else []
+        super().__init__(message, *args, **kwargs)
 
 
 class ResponseParsingError(ValueError, InstructorError):
--- a/instructor/v2/core/retry.py
+++ b/instructor/v2/core/retry.py
@@ -25,6 +25,7 @@ from instructor.v2.core.mode import Mode
 from instructor.v2.core.providers import Provider
 from instructor.v2.core.errors import (
     AsyncValidationError,
+    ConfigurationError,
     FailedAttempt,
     IncompleteOutputException,
     InstructorRetryException,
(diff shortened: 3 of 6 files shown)


wWzZb:
Thanks for putting this together. I had been investigating #2528 and paused implementation while waiting for the sync-client contract to be settled. Since you already have a PR open, I'd prefer to coordinate here rather than duplicate the work.

I compared this patch with #2537, #2559, and the original behavior introduced in #856. The retry hook and local imports look like the right direction. I found three cases that may be worth settling before merge:

- `model_declares_async_validators()` only inspects the top-level model MRO, so the sync guard appears to miss validators declared only on nested models, including prepared `list[Model]` and `Maybe(Model)` wrappers.
- Recursive list/dict validation stops at the first nested `AsyncValidationError`, so sibling failures are not aggregated and the errors do not include field or index paths.
- #856 passed the parsed model instance to field validators and ignored successful return values, while this patch passes the model class and applies return values as transformations. Is that compatibility change intentional?

If useful, I can take the nested sync-guard coverage and path-aware aggregation tests, then help tighten the runner while keeping streaming and parallel iterators out of scope for this PR.

jxnl:

PR #2595, opened 2026-09-05
fix(gemini): omit cache-owned config from cached requests

### LINKED ISSUE #2580: v2 GenAI handlers regression: cached_content requests now include system_instruction/tools/tool_config → 400 INVALID_ARGUMENT
## Summary

In 1.15.1, the GenAI handlers omitted `system_instruction`, `tools`, and `tool_config` from the request when the caller passed `cached_content` (the Gemini API rejects requests that set them alongside cached content — they must live in the cache). See `handle_genai_tools` / `handle_genai_structured_outputs` in `instructor/providers/gemini/utils.py` @ 1.15.1, which guard on `user_cached_content is None` with the comment "Adding them causes 400 INVALID_ARGUMENT".

In 1.16.0, `from_genai` routes to the v2 handlers (`instructor/v2/providers/genai/handlers.py`), and both `GenAIToolsHandler.prepare_request` and `GenAIStructuredOutputsHandler.prepare_request` build `base_config` unconditionally — the cached-content guard is gone. Every structured call that passes `config={"cached_content": ...}` now fails with:

```
400 INVALID_ARGUMENT: Tool config, tools and system instruction should not be set in the request when using cached content.
```

## Repro (no network needed — inspect the prepared request)

```python
from instructor.mode import Mode
from instructor.utils.providers import Provider
import instructor.v2.providers.genai.handlers  # noqa: F401

class Out(BaseModel):
    x: int

for mode in (Mode.JSON, Mode.TOOLS):
    prepare = mode_registry.get_handler(Provider.GENAI, mode)
    _, prepared = prepare(Out, {
        "messages": [{"role": "user", "content": "hi"}],
        "config": {"cached_content": "cachedContents/abc"},
    })
    cfg = prepared["config"]
    print(mode, {
        "system_instruction": cfg.system_instruction is not None,
        "tools": cfg.tools is not None,
        "tool_config": cfg.tool_config is not None,
        "cached_content": cfg.cached_content,
    })
```

1.16.0 output: JSON mode sets `system_instruction` (instructor's own JSON instruction); TOOLS mode sets `system_instruction`, `tools`, and `tool_config` — all alongside `cached_content`. Sending any of them with cached content is a guaranteed 400 from the Gemini API.

## Expected

When `cached_content` is present in the user config, `prepare_request` should omit `system_instruction`, `tools`, and `tool_config`, as 1.15.1 did (https://ai.google.dev/gemini-api/docs/caching).

## Environment
- instructor 1.16.0, google-genai 1.75.0, Python 3.12

Kuang-xianxin:
## Describe your changes

Structured GenAI requests with `config.cached_content` currently include system instructions and, in tools mode, tool declarations and tool configuration. Gemini rejects these fields alongside cached content, regressing the pre-v2 request handlers.

Remove the cache-owned fields after merging the shared GenAI configuration, so both JSON and tools mode keep their output schema, sampling settings and cache reference without sending conflicting fields. Requests without cached content retain their current behavior.

- Add eight regression cases using real `google.genai` configuration objects through the registered request handlers: dictionary and SDK-object configs, JSON and tools mode, cached and non-cached requests.
- Check that caller-owned inputs remain unchanged, and preserve JSON schema, temperature, labels and user content.
- Update the changelog and the existing GenAI development note.

Fixes #2580.

## Validation

- Before the fix: 4 new cached-request cases failed; 4 non-cached controls passed.
- Related GenAI tests: 58 passed on Python 3.13.14 with the repository's locked dependencies.
- `ruff check instructor examples tests` passed.
- Ruff formatting and `ty check` passed for the changed Python files.
- The full `ruff check .` also reports 36 pre-existing findings in untouched notebook/script files. Full `ty check` requires additional optional provider SDKs not installed in this environment.

The local commands above were executed; no live Gemini API request was made.

- [ ] I have performed a self-review of my code

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -19,6 +19,7 @@
 - **Factory pattern**: `from_provider()` for automatic provider detection
 - **DSL**: `dsl/` directory with Partial, Iterable, Maybe, Citation extensions
 - **Key modules**: `patch.py` (patching), `process_response.py` (parsing), `function_calls.py` (schemas)
+- **GenAI cached content**: Shared configuration assembly in `v2/providers/gemini/utils.py` must omit cache-owned `system_instruction`, `tools`, and `tool_config` when `cached_content` is set. Regression tests use actual SDK configuration objects in `tests/v2/test_genai_cached_content.py`.
 
 ## Code Style
 - **Typing**: Strict type annotations, use `BaseModel` for structured outputs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Gemini context caching**: Omit cache-owned system instructions, tools, and tool configuration from structured requests using cached content. ([#2580](https://github.com/567-labs/instructor/issues/2580))
+
 ## [1.16.1] - 2026-08-28
 
 ### Changed
--- a/instructor/v2/providers/gemini/utils.py
+++ b/instructor/v2/providers/gemini/utils.py
@@ -315,6 +315,11 @@ def update_genai_kwargs(
             if field_value is not None and field not in base_config:
                 base_config[field] = field_value
 
+    if base_config.get("cached_content") is not None:
+        # These fields belong to the cache; Gemini rejects them on cached requests.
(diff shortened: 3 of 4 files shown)


Kuang-xianxin:
Closing as a duplicate of #2582, which I missed in the initial searches. Sorry for the duplicate review request.

PR #2596, opened 2026-09-05
fix(genai): preserve reusable generation configuration

Kuang-xianxin:
## Describe your changes

Preparing a v2 GenAI request empties the caller's `generation_config` dictionary: the mode handlers shallow-copy kwargs, then the shared configuration mapper consumes its entries with `pop`. A second call using the same dictionary silently loses its token limit and sampling settings. Per-call overrides also write into that caller-owned dictionary before it is consumed.

Copy `generation_config` at the shared GenAI request boundary before either mode handler processes it. Four real-SDK regression cases cover tools/JSON modes, repeated preparation, preservation of all original options, and a one-call temperature override that must not affect a later request.

Related to #2465, whose fix covers `update_gemini_kwargs`. This PR covers the separate v2 GenAI path through `GenAIHandlerBase._clone_kwargs` and `update_genai_kwargs`, which still reproduces on main at `2b64e49`. It does not change the cached-content behavior addressed in #2582.

## Validation

- Before the fix: all 4 new regression cases fail because the original configuration becomes `{}`.
- After the fix: 54 related GenAI tests pass (Python 3.13.14, locked dependencies).
- `ruff check instructor examples tests` passes.
- Ruff formatting and `ty check` pass for the changed Python files.

Tests use the real Google SDK's request/configuration objects; no live provider request was made.

- [ ] I have performed a self-review of my code

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -19,6 +19,7 @@
 - **Factory pattern**: `from_provider()` for automatic provider detection
 - **DSL**: `dsl/` directory with Partial, Iterable, Maybe, Citation extensions
 - **Key modules**: `patch.py` (patching), `process_response.py` (parsing), `function_calls.py` (schemas)
+- **GenAI request ownership**: `_clone_kwargs` copies nested `generation_config` before either mode handler merges and translates sampling options. Keep caller-owned configurations reusable; regression tests are in `tests/v2/test_genai_config_reuse.py`.
 
 ## Code Style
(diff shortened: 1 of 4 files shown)


jxnl:
Consolidated into #2597: https://github.com/567-labs/instructor/pull/2597 . Both contributor commits are retained, with reusable GenAI configuration tests passing locally and Google GenAI CI passing on the consolidation. Closing as superseded; release remains gated on final coverage and independent/security review.

PR #2546, opened 2026-08-17
fix(genai): map OpenAI assistant role to Gemini model

tonycoder-hub:
convert_to_genai_messages rejected OpenAI-style assistant messages with ValueError: Unsupported role: assistant, so multi-turn history and few-shot examples never reached Gemini. The sibling transform_to_gemini_prompt already maps assistant to model; this does the same.

Intentionally updates tests/coverage/test_gemini_coverage.py, which previously asserted the ValueError for assistant (coverage of the buggy branch). ValueError coverage now uses the tool role.

Tests: uv run pytest tests/v2/test_gemini_utils_deterministic.py tests/coverage/test_gemini_coverage.py tests/v2/test_genai_handlers_deterministic.py -q — 48 passed. Distinct from #2542 and #2534.

PR #2214, opened 2026-03-24
fix(genai): raise IncompleteOutputException when finish_reason is MAX_TOKENS

nilerbarcelos:
## Describe your changes

`parse_genai_structured_outputs` was silently trying to parse truncated JSON whenever Gemini stopped due to token limits, resulting in a confusing `ValidationError` instead of an actionable `IncompleteOutputException`.

This adds the same guard that already exists for OpenAI (`finish_reason == "length"`) and Anthropic (`stop_reason == "max_tokens"`):

```python
if (
    hasattr(completion, "candidates")
    and completion.candidates
    and completion.candidates[0].finish_reason == types.FinishReason.MAX_TOKENS
):
    raise IncompleteOutputException(last_completion=completion)
```

Callers can now catch `IncompleteOutputException` and retry with a higher `max_output_tokens` budget, consistent with the behavior on other providers.

## Issue ticket number and link

Closes #2210

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [ ] If it is a core feature, I have added documentation.

PR #2601, opened 2026-09-06
fix(genai): restore MAX_TOKENS truncation check lost in v2 refactor

CTWalk:
## Describe your changes

**What.** On the Google GenAI path, a response truncated at `max_tokens` parses
as if it were complete, so the caller receives a confident object whose missing
fields are schema defaults. This restores the `finish_reason == MAX_TOKENS`
check and extends it to `Mode.TOOLS`.

**Why.** The value is indistinguishable from a real one:

```python
# finish_reason on the wire: FinishReason.MAX_TOKENS
Order(drug='warfarin', dose_mg=10, contraindications=None)
```

Here `contraindications=None` does not mean "none". It means the model was cut
off before it could say. Three things make this a regression rather than a
feature request:

1. **The docs already promise the fixed behaviour.**
   `docs/concepts/error_handling.md` lists `IncompleteOutputException` as
   "Output truncated due to token limit", with no provider carve out.
2. **The same truncation on the OpenAI path raises.** So the inconsistency sits
   inside instructor rather than in Gemini.
3. **The check shipped once and was then lost.** From the v1.15.0 changelog:

   > **Gemini**: Detect truncated responses (`finish_reason=MAX_TOKENS`) in
   > `GENAI_STRUCTURED_OUTPUTS` mode and raise `IncompleteOutputException`
   > immediately instead of retrying with malformed JSON (#2232)

   `60cc815` (*refactor(v2): complete migration cleanup and typing coverage*)
   moved the parser from `instructor/processing/function_calls.py` to the new
   `instructor/v2/core/function_calls.py`, and the guard did not come with it.
   That commit's diff carries the `finish_reason == MAX_TOKENS` comparison as
   removed lines. The method name survives as a deprecated shim, so nothing
   looked missing. At `5ea35cdb` there are zero occurrences of
   `FinishReason.MAX_TOKENS` left in the package. #2232 was closed as already
   applied, which was correct when it was written. The regression came
   afterwards.

**Changes.**

- Restore the truncation guard in `GenAIHandlerBase.parse_response`, at the
  single shared caller of `parse_genai_tools` and
  `parse_genai_structured_outputs`, using the same condition the deleted code
  used.
- Cover `Mode.TOOLS` as well. That mode never had the check and has the same
  defect. Both parse functions share this one caller, so splitting them would
  leave a known hole.
- Place the guard after the streaming early return, since `finish_reason` is not
  known until a stream finishes.
- Add `tests/test_genai_truncation.py` with 10 cases across both modes.
- Add a `CHANGELOG.md` entry under Unreleased.

The deprecated `ResponseSchema.parse_genai_structured_outputs` and
`.parse_genai_tools` shims delegate into this same handler, so they are covered
by the restored guard too. That is the entry point #2232's closing comment was
about.

**Out of scope.** The legacy `Provider.GEMINI` path has no truncation detection
either, but it uses a different SDK and response type and cannot share this fix,
so I left it out of a regression fix. Happy to follow up on it separately if
that would be useful.

### Testing

`tests/test_genai_truncation.py` is keyless and deterministic, built from real
`google.genai` types behind `pytest.importorskip`, in the same style as the
neighbouring `tests/test_genai_reask.py`, so no mocks, per `AGENT.md`:

- `MAX_TOKENS` raises `IncompleteOutputException`, in `Mode.TOOLS` and `Mode.JSON`.
- `STOP`, `SAFETY` and an unset `finish_reason` still parse, so a blocked
  response stays distinguishable from a truncated one.
- An empty `candidates` list behaves exactly as before, with no `IndexError` and
  no false truncation report.

Revert the guard and keep the tests: the two `MAX_TOKENS` cases fail and the
other eight stay green.

Ran locally against `5ea35cdb`. The new file passes 10 of 10, the offline suite
is unchanged, and `ruff check` and `ruff format --check` are clean at the
`v0.9.9` pinned in `.pre-commit-config.yaml`. The added block has full statement
and branch coverage from these tests alone, so the `full-coverage` job's
zero missing statements gate stays green.

One note for reviewers: the `core-tests` job filters `-k 'not test_genai'`, so
this file is skipped there, as `tests/test_genai_reask.py` already is. It does
run in `full-coverage`, which applies no provider filter.

None open. #2213, #2214 and #2232 all proposed this guard and are closed. The
specification is the documented behaviour in `docs/concepts/error_handling.md`
plus the v1.15.0 changelog entry quoted above.

No open PR touches this file. The recent genai work (#2539, #2546, #2582, #2591,
#2595, #2596) is all closed, and none of it touched `parse_response` or
`finish_reason`.

- [ ] If it is a core feature, I have added documentation. (No doc change
      needed. This restores conformance with the behaviour that
      `docs/concepts/error_handling.md` already documents.)

---

Written with assistance from Claude Code. Every claim above
was verified against `5ea35cdb`, and the tests were run locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Gemini truncation detection**: Restore the `finish_reason=MAX_TOKENS` check for the Google GenAI provider and extend it to `Mode.TOOLS`, so a truncated response raises `IncompleteOutputException` instead of parsing into schema defaults. The check shipped in v1.15.0 ([#2232](https://github.com/567-labs/instructor/pull/2232)) and was dropped during the v2 migration.
+
 ### Security
 - Route Anthropic PDF downloads through the bounded public-network fetcher and pin each media connection to a validated IP before sending HTTP.
 - Key cached responses by the complete prepared request, provider, validation context and strictness. Clients have isolated cache namespaces by default; `cache_namespace` explicitly enables sharing and must identify the endpoint and tenant. Revalidate cache hits with the current context and strictness.
--- a/instructor/v2/providers/genai/handlers.py
+++ b/instructor/v2/providers/genai/handlers.py
@@ -7,6 +7,7 @@ from typing import Any, cast
 from pydantic import BaseModel
 
 from instructor.v2.core.decorators import register_mode_handler
+from instructor.v2.core.errors import IncompleteOutputException
 from instructor.v2.core.handler import ModeHandler
 from instructor.v2.core.mode import Mode
 from instructor.v2.core.multimodal import extract_genai_multimodal_content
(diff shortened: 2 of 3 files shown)


jxnl:
Closing as superseded by merged #2628: https://github.com/567-labs/instructor/pull/2628 (main commit d99b6a36c6d7193e505818aa17375f65ed2c9e71). The non-streaming GenAI MAX_TOKENS fix was consolidated with regression tests. Streaming truncation remains outside this change. Original contributions are credited in the consolidation. The code is on main for planned 1.17.1; this does not claim that a release has been published.

PR #2604, opened 2026-09-07
fix: defensively sanitize markdown from raw JSON responses

hariom123-dev:
### TL;DR
Adds defensive sanitization to `_validate_model_from_json` to strip Markdown formatting (e.g., ` ```json `) and handle empty strings before parsing via Pydantic.

### 🔍 Context & Motivation
When using Instructor with various LLM providers (especially smaller open-source models or models that ignore system prompts), they frequently wrap their JSON output in Markdown blocks. Passing this raw string directly to `json.loads()` or `model_validate_json()` causes an immediate parsing crash.

By catching and stripping this formatting upstream, we make the library significantly more resilient to varying LLM output behaviors.

### 🛠️ Changes Implemented
- In `instructor/v2/core/function_calls.py`, added a Regex pre-processing step inside `_validate_model_from_json`.
- It defensively strips surrounding whitespace.
- If the string starts with a markdown block, it cleanly removes the opening (` ```json ` or ` ``` `) and the closing (` ``` `).
- If the LLM returns an empty string, it fails fast with a clear `ValueError` instead of a generic `JSONDecodeError`.

### ✅ Verification
- Added `tests/v2/test_json_sanitization.py`.
- Tested that heavily wrapped/whitespace-padded markdown strings are correctly parsed into the Pydantic model.
- Tested that empty strings raise the correct `ValueError`.


jxnl (changes requested):
Reviewed exact head ff8280b03640361ae81595d53bc725787959c109 for 1.18. Empty input now raises a plain ValueError before the existing parse/validation boundary. The default retry path handles JSON/model parse exceptions, not arbitrary ValueError, so this turns a recoverable malformed completion into a terminal failure. The new unit test asserts the new exception but does not exercise retry behavior. Please preserve the existing retryable error contract and add end-to-end sync/async retry regressions for empty and fenced malformed input. Reuse the existing bounded JSON extraction policy instead of adding a separate regex cleanup path without demonstrating a missing case. Deferred from the release until those contracts are preserved.

hariom123-dev:
@jxnl Thanks for the detailed architectural review. You were completely right about breaking the retry contract with the `ValueError`. I've updated the PR to address your points:

1. **Preserved Retry Contract:** Removed the fast-fail `ValueError`. Empty strings now fall through to the parser, natively raising `JSONDecodeError` / Pydantic `ValidationError` so the retry engine catches and recovers them.
2. **Reused Existing Extractor:** Removed the custom Regex and instead wired the input through the existing JSON extraction utility.
3. **Added End-to-End Tests:** Added `test_retry_contract_on_empty_and_malformed_input` using a mocked client to assert that the retry engine successfully recovers from an empty string, then a malformed block, and finally parses the successful third attempt.

The branch is also synced with `main`. Let me know if the retry regressions look good for 1.18!

Issue #2613, opened 2026-09-09
`is_simple_type` falsely returns True for custom non-Pydantic classes on Python 3.10+, causing schema generation crashes


Gracy769:
## Summary

When passing a list containing a custom class (that does not inherit from `BaseModel`) as a response model, Instructor crashes with a confusing `PydanticSchemaGenerationError` on Python 3.10 and above.

## Reproducer

``python
from instructor.utils.core import prepare_response_model

class MyClass:
    pass

model = prepare_response_model(list[MyClass])
``

**Output:**
``
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class '__main__.MyClass'>. ...
``

## Root Cause

In `instructor/v2/dsl/simple_type.py`, the `is_simple_type` function contains a check intended to identify Python 3.10+ pipe syntax (`|`) for Union types:

``python
# Check for Python 3.10+ pipe syntax
if hasattr(inner_arg, __or__):
    return True
``

However, in Python 3.10, the `__or__` method was added directly to the `type` builtin itself. This means that **every single class in Python** now returns `True` for `hasattr(cls, '__or__').

Because `issubclass(inner_arg, BaseModel)` fails for `MyClass`, it falls through to the `hasattr` check, which returns `True`. Instructor then mistakenly assumes `list[MyClass]` is a 'simple type' and forces it through `ModelAdapter.__class_getitem__`, which ultimately explodes inside Pydantic.

## Suggested Fix

Refine the pipe syntax check. Instead of broadly checking `hasattr(inner_arg, '__or__')`, we should check if the type is actually a Union, or simply let Pydantic handle validation failures more gracefully instead of forcefully wrapping custom classes.

ZIFeIYUuuuuuu:
I opened #2629 with the type-classification fix and regression coverage: https://github.com/567-labs/instructor/pull/2629

PR #2629, opened 2026-09-09
fix(v2): distinguish custom classes from union types

ZIFeIYUuuuuuu:
## Motivation / Context

On Python 3.10 and newer, every class exposes `__or__` through the `type` implementation. `is_simple_type()` was using `hasattr(inner_arg, "__or__")` as a union detector, so `list[CustomClass]` was incorrectly classified as a simple scalar response model. `prepare_response_model(list[CustomClass])` then forced the type through the simple-model adapter and surfaced a confusing `PydanticSchemaGenerationError`.

## Description

- Detect PEP 604 unions using their actual `types.UnionType` origin, alongside `typing.Union`.
- Remove the broad `__or__` attribute checks from list and iterable handling.
- Reject unsupported class elements in iterable response models with a focused `TypeError` explaining that elements must be Pydantic models or typed dictionaries.
- Add regression coverage for custom classes and update the coverage expectation for `list[object]`.

Basic scalar lists, `typing.Union`, PEP 604 unions such as `list[int | str]`, Pydantic model lists, and typed dictionaries retain their existing behavior.

## Related issue number

Related to #2613

## Contribution Checklist

- [x] Tests added or updated for the behavior change.
- [x] Focused tests pass.
- [x] Ruff format and lint pass.
- [x] No documentation changes are needed for this internal type classification fix.

## Reproduction

Before this change:

```python

class CustomClass:
    pass

prepare_response_model(list[CustomClass])
# PydanticSchemaGenerationError: Unable to generate pydantic-core schema ...
```

The minimal type predicate also returned `True` for `is_simple_type(list[CustomClass])`.

After this change, the predicate returns `False` and preparation fails early with an actionable message instead of entering the simple-type adapter path.

## Validation

- `uv run --extra dev pytest tests/dsl/test_simple_types.py tests/coverage/test_dsl_small_coverage.py -q` — **37 passed**.
- `uv run --extra dev pytest tests/dsl -q` — **77 passed, 2 skipped** (54 existing deprecation warnings).
- `uv run ruff format --check instructor/v2/dsl/simple_type.py instructor/v2/core/response_model.py tests/dsl/test_simple_types.py tests/coverage/test_dsl_small_coverage.py` — **all files formatted**.
- `uv run ruff check instructor/v2/dsl/simple_type.py instructor/v2/core/response_model.py tests/dsl/test_simple_types.py tests/coverage/test_dsl_small_coverage.py` — **all checks passed**.
- `git diff --check` — **passed**.


jxnl (changes requested):
Reviewed c842caf6722131b36118dba5061088cea34642b0 against main 1b4664e7. Please preserve existing Pydantic-supported response types before merging. The new response_model.py class gate rejects list[Enum], list[date], list[Decimal], list[UUID], dataclasses and classes implementing __get_pydantic_core_schema__; Iterable[int] also regresses. These prepared and generated schemas on main. Separately, list[Annotated[int, Field(gt=0)]] silently changes from the content adapter to the tasks/IterableBase schema. Removing only the class gate does not preserve that routing. Keep the version-safe UnionType check, but add regressions for supported schema types and existing adapter shapes, and limit the improved diagnostic to genuinely unsupported schemas. Existing focused tests pass but do not cover these regressions. Please add a concise changelog entry and retain Python 3.9 compatibility coverage. Issue #2613 should remain open until a compatible fix is accepted.

ZIFeIYUuuuuuu:
Thanks for the detailed review. Everything you listed reproduced, and I measured the
blast radius against `main` before touching anything.

First, what `main` actually does. `hasattr(inner_arg, "__or__")` was never a union
probe: on Python 3.10+ `type` defines `__or__`, so it is true for *every* class, and on
Python 3.9 it is false for every class. In effect `main` accepted every element type,
and this PR narrowed that to "must be a `BaseModel` subclass" — which is the regression
you saw. Two of the shapes you listed were not even touched by the class gate, which is
why dropping it alone would not have been enough:

| response model | main | this PR before your review |
| --- | --- | --- |
| `list[Enum]`, `list[date]`, `list[Decimal]`, `list[UUID]`, `list[dataclass]`, `list[__get_pydantic_core_schema__]` | content adapter | `TypeError` |
| `Iterable[int]` | `IterableModel` | `TypeError` |
| `list[Annotated[int, Field(gt=0)]]` | content adapter | `IterableModel` (silent) |
| `list[Literal["a", "b"]]` | content adapter | `IterableModel` (silent) |

`list[Literal[...]]` was not in your list but had the same silent switch, so the fix
covers it too.

What changed:

1. `simple_type.py`: the `__or__` probe is gone. A member keeps the content-adapter
   path unless it is (a) a `BaseModel` subclass, (b) a class Pydantic cannot generate a
   schema for, or (c) not a type at all (`None`, `3`, `"ForwardRef"` — exactly the
   members that happened to lack `__or__`, which the old code rejected by accident).
   The version-safe `UnionType` check is kept, so `list[int | str]` and
   `list[Union[int, str]]` are unaffected.
2. `response_model.py`: the iterable gate asks Pydantic instead of pattern-matching on
   `BaseModel`. It raises the explicit `TypeError` only when a schema genuinely cannot
   be generated, so `list[CustomClass]` still fails loudly and #2613 stays fixed.

Verification: I dumped the JSON schema `prepare_response_model` produces on `main` and on
this branch for 12 response models. They are byte-identical, including both silent-switch
shapes above. The only differences from `main` are the diagnostic for unsupported element
types, which is what #2613 is about. `ruff format --check` and `ruff check` pass on the
changed files. With every provider extra installed (`bedrock`, `mistral`, `google-genai`,
`vertexai`), the `tests/dsl tests/v2 tests/core tests/processing` run passes 2137 tests
with a single failure: `test_json_extraction_bounds_and_recovery`, which fails
identically on `main` — it expects `"nesting"` in the error message but the
implementation raises `"JSON extraction malformed-input work limit exceeded"`. That is a
pre-existing mismatch in the test's expected message, unrelated to this change.

Two deviations worth stating rather than hiding:

- `list[T]` with an unsolved `TypeVar` reached the content adapter on `main` (because
  `TypeVar` has `__or__`) and now routes to `IterableModel`. It is invalid input either
  way — say the word if you want that path kept and I will add the exemption.
- For a class whose `__get_pydantic_core_schema__` itself raises, `is_simple_type` now
  propagates that error instead of returning `True` first. `prepare_response_model`
  fails with the identical exception either way; only the helper's return type changes.

Tests:

- Added regressions for each supported element type above. They assert which adapter was
  actually chosen (a `content` field vs a `tasks` field), not merely "does not raise", so
  a silent rerouting fails the test. One test pins that routing no longer depends on
  `hasattr(member, "__or__")`.
- `tests/coverage/test_dsl_small_coverage.py`: the existing assertion was
  `assert is_simple_type(list[object]) is hasattr(object, "__or__")`, which asserts the
  coupling itself rather than the behaviour. It now reads
  `assert is_simple_type(list[object])`; on CPython 3.10+ both spellings evaluate the
  same way. The two `monkeypatch.setattr(simple_type, "hasattr", ...)` calls are removed
  because `is_simple_type` no longer reads `hasattr`, and the `list[Literal["one"]]`
  expectation in the same test flipped with them.
- `CHANGELOG.md` shortened to a single entry under `[Unreleased] → Fixed`.
- The `sys.version_info` coverage for the legacy `Iterable` origin is untouched.

`list[User]` and `Iterable[User]` still route to `IterableModel`, unchanged. #2613 stays
open until you are happy with this.

ZIFeIYUuuuuuu:
@jxnl Conflict resolved: I merged `main` (22 commits since I branched) and fixed the only
conflicting file, `CHANGELOG.md` — upstream had just cut 1.17.1, so I kept those release
notes and left this branch's entry under `[Unreleased]`. The PR is mergeable again.

Re-checked against the new `main`:

- `prepare_response_model` still produces an identical JSON schema shape for all 12
  response models, including the two silent-switch shapes.
- `tests/dsl tests/v2 tests/core tests/processing`: **2343 passed, 187 skipped, 0 failed**.

One correction to my earlier report: `test_json_extraction_bounds_and_recovery`, the single
failure I listed, now passes — a merged commit fixed the message mismatch, so that caveat no
longer applies.

ZIFeIYUuuuuuu:
@jxnl one more thing while this is in your queue — the Actions runs on the current head (`438debe2`) are still at `action_required`, so `Test`, `ty`, `Ruff`, `Package compatibility`, and `Release Readiness` have not executed on the merged-with-main revision.

If you could approve the workflow runs when you re-review, CI would confirm the conflict resolution didn't move anything.

The head SHA itself changed while this sat in your queue: same tree, only the commit author metadata was corrected (the first commit was authored with a local machine address, so it showed up as an unlinked commit). The diff is byte-identical, so your review applies as written.

PR #2642, opened 2026-09-12
fix(v2): stop misclassifying custom classes as simple types on Python 3.10+

### LINKED ISSUE #2613: `is_simple_type` falsely returns True for custom non-Pydantic classes on Python 3.10+, causing schema generation crashes

chenlichao:
## Describe your changes

`is_simple_type()` misclassifies response models like `list[MyClass]` (where `MyClass` is a custom, non-`BaseModel` class) as simple types on Python 3.10+, so `prepare_response_model` crashes with an opaque `PydanticSchemaGenerationError` instead of handling the unsupported input gracefully.

Root cause: the function probed `hasattr(inner_arg, "__or__")` to detect PEP 604 pipe unions. On Python 3.10+ `__or__` lives on `type` itself, so **every** class satisfies that check and any custom class inside `list[...]` is misrouted into `ModelAdapter` / `IterableModel`, which then explodes inside Pydantic.

Changes:

- `instructor/v2/dsl/simple_type.py`: replace both `hasattr(..., "__or__")` probes with a real union-origin check (`get_origin(X | Y) is types.UnionType`), using the same version-guarded `_UNION_ORIGINS` tuple already used in `iterable.py` and `response_model.py`. The `typing.Union[...]` checks directly above already cover `List[Union[int, str]]`, so behavior for genuine unions is unchanged.
- `instructor/v2/core/response_model.py`: hoist the list-element `_is_model_type` helper to module scope (unchanged logic) and raise a clear `TypeError` when a `list[...]`/`Iterable[...]` element is not a `BaseModel` subclass, `TypedDict`, or a union of those — instead of the opaque Pydantic error.
- `tests/dsl/test_simple_types.py`: regression tests for the issue repro (`is_simple_type(list[MyClass]) is False` and `prepare_response_model(list[MyClass])` raises `TypeError`).
- `tests/coverage/test_dsl_small_coverage.py`: update one assertion that hard-coded the old behavior as a tautology (`is_simple_type(list[object]) is hasattr(object, "__or__")` → `not is_simple_type(list[object])`).
- `CHANGELOG.md`: entry under `[Unreleased]` per the repo's changelog policy.

Fixes #2613

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versioning: [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Fixed
+- **Response model validation**: Custom (non-BaseModel) classes used as `list[...]` element types are no longer misclassified as simple types on Python 3.10+ (`__or__` now lives on `type` itself, so probing for it matched every class), and `prepare_response_model` raises a clear `TypeError` for unsupported element types instead of an opaque Pydantic schema error. ([#2613](https://github.com/567-labs/instructor/issues/2613))
+
 ## [1.17.1] - 2026-09-09
 
 ### Upgrade Notes
--- a/instructor/v2/core/response_model.py
+++ b/instructor/v2/core/response_model.py
@@ -65,6 +65,18 @@ def is_simple_type(typehint: type[T]) -> bool:
     return _is_simple_type(typehint)
 
 
+def _is_model_type(candidate: Any) -> bool:
+    """True for BaseModel subclasses, TypedDicts, and unions of those."""
+    if is_typed_dict(candidate):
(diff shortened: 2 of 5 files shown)


chenlichao:
Friendly ping for a first look — on 3.10+ PEP 604 puts `__or__` on `type` itself, so `hasattr(cls, "__or__")` matched every class and `list[MyClass]` was misrouted into the model adapter, dying with an opaque `PydanticSchemaGenerationError`. The fix uses a real union-origin check and raises a clear `TypeError` for unsupported element types, with tests. Closes #2613.

PR #2644, opened 2026-09-13
fix(dsl): refine UnionType check in is_simple_type for Python 3.10+ (#2613)

### LINKED ISSUE #2613: `is_simple_type` falsely returns True for custom non-Pydantic classes on Python 3.10+, causing schema generation crashes

Tyagiquamar:
Fixes #2613


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #2664, opened 2026-09-19
fix(streaming): don't crash on an empty choices/candidates list mid-stream

pcbeingused333:
## The bug

Mistral, Gemini, GenAI, and VertexAI's streaming JSON extractors (`instructor/v2/providers/{mistral,gemini,genai,vertexai}/handlers.py`) index `choices[0]`/`candidates[0]` unconditionally, but that list can legitimately be empty mid-stream:

- **Mistral**: a trailing usage-only chunk (`choices: [], usage: {...}`) — the same shape OpenAI-compatible APIs use for `stream_options={"include_usage": True}`.
- **Gemini / GenAI / VertexAI**: a safety-blocked chunk (empty `candidates`).

Each of these four handlers wraps the extraction in `except AttributeError: continue`, specifically to skip malformed/incomplete chunks gracefully — but `choices[0]`/`candidates[0]` on an empty list raises `IndexError`, not `AttributeError`, so it wasn't caught. The exception propagates out of the generator into the streaming consumer and kills the whole structured-output stream, instead of just skipping that one chunk.

This repo already has the correct pattern in two places that these four handlers don't follow:
- `openai/handlers.py` guards with `if not getattr(chunk, "choices", None): continue` before indexing.
- `gemini/handlers.py`'s own `parse_gemini_tools` already catches `(AttributeError, IndexError, KeyError, TypeError)`.

## The fix

Widened each `except AttributeError:` to `except (AttributeError, IndexError):` at all 9 affected sites (2 each in mistral/gemini/genai/vertexai handlers, plus the shared `extract_gemini_chunk_text` helper's caller-facing contract — that one already relies on its callers to catch the propagated error, so its own docstring/comment was updated rather than its except clause, since its fallback line raises the IndexError itself, past its own try/except).

## Testing

Added an empty-choices/empty-candidates case to each provider's existing "skip empty/bad chunks" coverage test in `tests/coverage/`, built from the real SDK types where the existing tests already do (e.g. Mistral's `CompletionChunk`/`CompletionEvent`).

Verified red against the pre-fix code — all 8 new cases raised `IndexError`:
```
instructor/v2/providers/vertexai/handlers.py:270: IndexError
instructor/v2/providers/mistral/handlers.py:... IndexError
instructor/v2/providers/gemini/handlers.py:... IndexError
instructor/v2/providers/genai/handlers.py:... IndexError
```

Green after the fix: `ruff check`/`ruff format --check` clean, `ty check` clean (0 errors), and the full non-network `tests/coverage` + `tests/core` + `tests/dsl` suites pass (668 passed; the only failures are pre-existing gaps from provider SDKs I didn't install locally — cohere, groq, writer, bedrock/botocore, cerebras, fireworks, xai, diskcache — unrelated to this change).


pcbeingused333:
No CI checks have started yet on this PR — if that's a fork-approval gate, local verification while it's waiting:

\`\`\`
$ .venv/bin/python -m ruff check <changed files>
All checks passed!
$ .venv/bin/ty check <changed source files>
All checks passed!
$ .venv/bin/python -m pytest tests/coverage/test_mistral_coverage.py tests/coverage/test_gemini_coverage.py tests/coverage/test_genai_coverage.py tests/coverage/test_vertexai_coverage.py -q
81 passed
$ .venv/bin/python -m pytest tests/coverage tests/core tests/dsl -q -k "not llm and not openai" --ignore=tests/coverage/test_bedrock_coverage.py --ignore=tests/coverage/test_cohere_coverage.py
668 passed (remaining failures are pre-existing gaps from provider SDKs not installed locally — cohere, groq, writer, bedrock/botocore, cerebras, fireworks, xai, diskcache — unrelated to this change)
\`\`\`

